PR #11, opened 2026-05-26
Adds support for MoE mxfp8 models (tested with Qwen3.6-35B-A3B-mxfp8)

FlashShift:
This code was generated through a Claude session, and gets Qwen3.6-35B-A3B-mxfp8 working.


ddalcu:
thanks for your PR, but why MXFP8 ? It seems like a dead format. 
Most people have moved on to Int4 variants & NVFP4 better/faster/smaller.

Are you trying to run any particular model ?

PR #17, opened 2026-06-04
feat: support fully-dense bf16 models (Qwen3.5/3.6 MoE and similar)

pablomoralesm:
## What this adds

Support for fully-dense bf16 models such as `mlx-community/Qwen3.6-27B-bf16` and `mlx-community/Qwen3.6-35B-A3B-bf16` (`qwen3_5` / `Qwen3_5ForConditionalGeneration`, `language_model.model.*` weight prefix).

Previously these crashed immediately after loading weights:
```
MISSING WEIGHT: language_model.model.embed_tokens.scales
```

**Memory note:** Qwen3.6-27B-bf16 is ~51 GB — comparable to the DeepSeek V4 Flash GGUF (~80 GB) already supported via ds4.

## Detection via config.json

`quant_bits` now defaults to `0` instead of `4`. Models with a `"quantization"` key in config.json set `quant_bits` from that (unchanged). Models without it — dense bf16 — correctly report `quant_bits=0`. This is the primary signal used throughout loading and inference.

## Root cause

mlx-serve already handled mixed-precision Unsloth Dynamic checkpoints where individual layers are bf16 within an otherwise-quantized model. Fully-dense bf16 models extend that case: *all* layers are bf16, so no `.scales` or `.biases` tensors exist anywhere. Every fetch site used the mandatory `getWeightFmt`/`getLayerWeight` helpers, which panic on a missing tensor.

## Changes

**`src/model.zig`**
- `quant_bits` default: `4` → `0` (0 = dense bf16, no `"quantization"` key)

**`src/transformer.zig`**

| Change | Description |
|---|---|
| New `getLayerScaleOrEmpty(quant_bits)` | `quant_bits==0` → empty array (ctx==null); `>0` → mandatory fetch so corrupted quantized models still fail fast. Replaces 46 mandatory scale/bias fetches across all layer initializers. |
| `embed_tokens` / `lm_head` loading | `config.quant_bits==0` guard |
| `embedding()` | Skip `mlx_take_axis`+`mlx_dequantize` when `emb_s.ctx==null`; cast directly to bf16 |
| `dequantTake()` | Same early-return dense path |
| `appendFullAttnWeights()` | Guard appends with `ctx!=null` (matches `appendLinearAttnWeights` / `appendHybridMlpWeights`) |
| `qmatmulBits()` bf16 path | MLP weights are pre-transposed by `maybeTransposeForBf16`; attention projections are not. Shape-based detection: `w.last==x.last` → `[out,in]`, transpose; otherwise → `[in,out]`, existing direct path unchanged |

**`src/main.zig`**
- Remove redundant `errdefer` on `config_storage` — `defer` already covers all exit paths, causing a double-free with Zig's debug allocator

**`README.md`**
- "Any quantized MLX model" → "Any MLX model (quantized or dense bf16)"

## Tests

`src/model.zig`:
- `parseConfigFromJson quant_bits is 0 for dense bf16 (no quantization key)`
- `parseConfigFromJson quant_bits is set from quantization key`

`src/transformer.zig`:
- `qmatmulBits transposes [out, in] bf16 weight when scales has null ctx`
- `appendFullAttnWeights skips fields with null ctx (plain bf16 layers)`

**`zig build -Doptimize=ReleaseFast test`: 389 passed, 3 skipped, 0 failed**

Tested end-to-end: `Qwen3.6-27B-bf16` loads, warms up, and generates correctly.

Note: `swift test` has a pre-existing build failure in the pinned `swift-sdk` 0.10.x dependency unrelated to this change. It passes with `swift test -Xswiftc -swift-version -Xswiftc 5` as documented in CLAUDE.md.

🤖 Generated with [Claude Code](https://claude.ai/claude-code)

### DIFF
--- a/README.md
+++ b/README.md
@@ -151,7 +151,7 @@ Outputs go to `~/.mlx-serve/generations/images/YYYY-MM-DD/` and `.../videos/YYYY
 | **DeepSeek V4 Flash** | `deepseek_v4` (GGUF) | DeepSeek-V4-Flash | DSV4 | -- |
 | **Anything else as GGUF** | via embedded llama.cpp | any `.gguf` on HuggingFace | per-template | -- |
 
-Any quantized MLX model using one of the above architectures works natively. Anything else can be served as GGUF through the embedded llama.cpp engine — just pick the `.gguf` file in the Model Browser and the server auto-routes by format. Models with unsupported architectures are flagged in the Model Browser but can still be downloaded.
+Any MLX model (quantized or dense bf16) using one of the above architectures works natively. Anything else can be served as GGUF through the embedded llama.cpp engine — just pick the `.gguf` file in the Model Browser and the server auto-routes by format. Models with unsupported architectures are flagged in the Model Browser but can still be downloaded.
 
 ## Prerequisites
 
--- a/src/main.zig
+++ b/src/main.zig
@@ -505,7 +505,6 @@ pub fn main(init: std.process.Init) !void {
     // explicit defer on `config_storage`.
     const config_storage = try allocator.create(model_mod.ModelConfig);
     var config_owned_by_registry = false;
-    errdefer if (!config_owned_by_registry) allocator.destroy(config_storage);
     defer if (!config_owned_by_registry) allocator.destroy(config_storage);
     config_storage.* = try model_mod.parseConfig(io, allocator, model_dir);
     const config = config_storage;
--- a/src/model.zig
+++ b/src/model.zig
@@ -34,8 +34,8 @@ pub const ModelConfig = struct {
     sliding_window: u32 = 1024,
     sliding_window_pattern: u32 = 6,
 
-    // Quantization
-    quant_bits: u32 = 4,
+    // Quantization. 0 = dense bf16 (no "quantization" key in config.json).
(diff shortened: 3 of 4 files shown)


pablomoralesm:
Closing to continue work — there's an additional case to handle for stacked MoE expert weights (mlx_gather_qmm needs a dense bf16 path). Will reopen once fixed.

pablomoralesm:
Reopening with two additional fixes for stacked MoE expert weights:

1. `transposeBf16Weight`: generalized from 2D-only (`perm=[1,0]`) to swap the last two axes for any rank — fixing stacked expert tensors like `[experts, out, in]` → `[experts, in, out]`.

2. New `gatherExpertMm` helper + `mlx_gather_mm` binding: the 6 `mlx_gather_qmm` call sites in the Qwen3.5 MoE forward pass now route to `mlx_gather_mm` (dense) when `sc.ctx==null`, instead of passing empty scale arrays to `mlx_gather_qmm`.

Both Qwen3.6-27B-bf16 and Qwen3.6-35B-A3B-bf16 now load and generate correctly. Tests: 389 passed, 3 skipped, 0 failed.

PR #18, opened 2026-06-04
feat: support fully-dense bf16 models (Qwen3.5/3.6 MoE and similar)

pablomoralesm:
## What this adds

Support for fully-dense bf16 models such as `mlx-community/Qwen3.6-27B-bf16` and `mlx-community/Qwen3.6-35B-A3B-bf16` (`qwen3_5` / `qwen3_5_moe`, `language_model.model.*` weight prefix).

Previously these crashed immediately after loading weights:
```
MISSING WEIGHT: language_model.model.embed_tokens.scales
```

**Memory note:** Qwen3.6-27B-bf16 is ~51 GB — comparable to the DeepSeek V4 Flash GGUF (~80 GB) already supported via ds4.

## Detection via config.json

`quant_bits` now defaults to `0` instead of `4`. Models with a `"quantization"` key in config.json set `quant_bits` from that (unchanged). Models without it — dense bf16 — correctly report `quant_bits=0`. This is the primary signal used throughout loading and inference.

## Changes

**`src/model.zig`**
- `quant_bits` default: `4` → `0` (0 = dense bf16, no `"quantization"` key)

**`src/mlx.zig`**
- Added `mlx_gather_mm` binding (dense gathered matmul, used for bf16 MoE expert dispatch)

**`src/transformer.zig`**

| Change | Description |
|---|---|
| New `getLayerScaleOrEmpty(quant_bits)` | `quant_bits==0` → empty array; `>0` → mandatory fetch so corrupted quantized models still fail fast. Replaces 46 mandatory scale/bias fetches. |
| `embed_tokens` / `lm_head` loading | `config.quant_bits==0` guard |
| `embedding()` | Skip `mlx_take_axis`+`mlx_dequantize` when `emb_s.ctx==null` |
| `dequantTake()` | Same early-return dense path |
| `appendFullAttnWeights()` | Guard appends with `ctx!=null` |
| `transposeBf16Weight()` | Generalized from 2D-only (`perm=[1,0]`) to swap last two axes for any rank — fixes stacked MoE expert tensors `[experts, out, in]` → `[experts, in, out]` |
| New `gatherExpertMm()` helper | Routes to `mlx_gather_mm` (dense) when `sc.ctx==null`, replacing the 6 `mlx_gather_qmm` call sites in the Qwen3.5 MoE forward pass |
| `qmatmulBits()` bf16 path | Shape-based detection: `w.last==x.last` → `[out,in]`, use `transposeBf16Weight`; otherwise → `[in,out]`, existing direct path |

**`src/main.zig`**
- Remove redundant `errdefer` on `config_storage` — double-free with Zig's debug allocator

**`README.md`**
- "Any quantized MLX model" → "Any MLX model (quantized or dense bf16)"

## Tests

`src/model.zig`: 2 new tests for `quant_bits` config parsing  
`src/transformer.zig`: 2 new tests for bf16 weight paths  
**`zig build -Doptimize=ReleaseFast test`: 389 passed, 3 skipped, 0 failed**

Tested: `Qwen3.6-27B-bf16` and `Qwen3.6-35B-A3B-bf16` load, warm up, and generate correctly.

Note: `swift test` has a pre-existing build failure in the pinned `swift-sdk` 0.10.x dependency. Passes with `swift test -Xswiftc -swift-version -Xswiftc 5` as documented in CLAUDE.md.

🤖 Generated with [Claude Code](https://claude.ai/claude-code)

### DIFF
--- a/README.md
+++ b/README.md
@@ -151,7 +151,7 @@ Outputs go to `~/.mlx-serve/generations/images/YYYY-MM-DD/` and `.../videos/YYYY
 | **DeepSeek V4 Flash** | `deepseek_v4` (GGUF) | DeepSeek-V4-Flash | DSV4 | -- |
 | **Anything else as GGUF** | via embedded llama.cpp | any `.gguf` on HuggingFace | per-template | -- |
 
-Any quantized MLX model using one of the above architectures works natively. Anything else can be served as GGUF through the embedded llama.cpp engine — just pick the `.gguf` file in the Model Browser and the server auto-routes by format. Models with unsupported architectures are flagged in the Model Browser but can still be downloaded.
+Any MLX model (quantized or dense bf16) using one of the above architectures works natively. Anything else can be served as GGUF through the embedded llama.cpp engine — just pick the `.gguf` file in the Model Browser and the server auto-routes by format. Models with unsupported architectures are flagged in the Model Browser but can still be downloaded.
 
 ## Prerequisites
 
--- a/src/main.zig
+++ b/src/main.zig
@@ -505,7 +505,6 @@ pub fn main(init: std.process.Init) !void {
     // explicit defer on `config_storage`.
     const config_storage = try allocator.create(model_mod.ModelConfig);
     var config_owned_by_registry = false;
-    errdefer if (!config_owned_by_registry) allocator.destroy(config_storage);
     defer if (!config_owned_by_registry) allocator.destroy(config_storage);
     config_storage.* = try model_mod.parseConfig(io, allocator, model_dir);
     const config = config_storage;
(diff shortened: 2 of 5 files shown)


ddalcu:
thanks for this, but curious, why BF16 ? these are not models you should be running directly since they are big & slow. 8bit is near lossless and twice the speed.

pablomoralesm:
Fair point. For pure inference, 8-bit is the right call — the main bf16 use cases (fine-tuning, creating quants, benchmarking) may be out of the scope you intended for this project.

Full disclosure: this is actually a reopened version of a previous PR I closed while chasing a generation quality bug in the Qwen3.6-35B-A3B-bf16 MoE model that I haven't fully resolved. Closing for now — happy to reopen once the MoE generation issue is fixed if bf16 support is something you'd eventually want.

PR #32, opened 2026-06-13
Fix macOS dependency syntax for MLX Core

pnavais:
Fixes the syntax issue 

```
Warning: Calling string comparison format for `depends_on macos:` is deprecated! Use `depends_on macos: :sonoma` instead.
Please report this issue to the Homebrew/homebrew-mlx-serve tap, or even better, submit a PR to fix it:
  /opt/homebrew/Library/Taps/homebrew/homebrew-mlx-serve/Casks/mlx-core.rb:10
```


ddalcu:
thanks, but mlx-core depends on MacOS 26 and above, Sonoma is no longer supported. Can you please update to 26 and above ?

pnavais:
Done, the new syntax will check macOS Tahoe (26) or newer

PR #36, opened 2026-06-14
Add Qwen2 (Qwen2.5) architecture support

### LINKED ISSUE #35: Add Qwen2 (Qwen2.5) architecture support — currently skipped by discovery
Qwen2.5 checkpoints (`model_type: "qwen2"`, e.g. `Qwen2.5-Coder-32B-Instruct`) are silently dropped by `--model-dir` discovery and never appear in `/v1/models`, because the server has no `qwen2` forward path (`qwen2` isn't in `model_discovery.supported_model_types`). The Swift client, however, lists `qwen2` in `supportedModelTypes` — so the app happily downloads a Qwen2.5 model and shows it as usable, but the server then can't serve it. Client/server mismatch.

Qwen2 is close to Qwen3 but differs in three ways that all matter:

1. **No QK-norm** (Qwen3 has q_norm/k_norm; Qwen2 doesn't).
2. **Additive qkv-projection biases** — Qwen2 carries `q/k/v_proj.bias` (additive `+b`, distinct from the MLX quant `.biases`), which the standard attention path doesn't apply.
3. **head_dim** is omitted from config.json (implicit = `hidden_size / num_attention_heads`); the parser left it at the stale 256 sentinel, which would corrupt attention. (Also affects llama/mistral exports that omit head_dim.)
4. **Tokenizer merges** are stored as space-joined strings (`"a b"`) rather than pair-arrays (`["a","b"]`); the loader assumes arrays, and in ReleaseFast reading the wrong union field spins the merges loop on garbage → the server **hangs during tokenizer load**.

Fix in PR. Verified end-to-end: `Qwen2.5-Coder-32B-Instruct-8bit` loads (head_dim 128, 8-bit affine) and generates correct output.

ujwal-setlur:
Fixes #35.

Adds full server-side support for `model_type: "qwen2"` (Qwen2.5 family). Qwen2 is Llama-family but differs from Qwen3 in ways that each required handling — discovered while bringing up `Qwen2.5-Coder-32B-Instruct`.

## Changes
- **Dispatch + config** (`model.zig`): classify `qwen2` (no QK-norm, silu). Default `head_dim` to `hidden_size / num_attention_heads` when config.json omits it (Qwen2.5 does) — previously the stale 256 sentinel corrupted attention. Also fixes llama/mistral exports that omit `head_dim`.
- **Additive qkv bias** (`transformer.zig`): Qwen2 carries `q/k/v_proj.bias` (additive, distinct from the quant `.biases`), which the standard attention path didn't apply. Load them as optional `LayerWeights` fields and apply via a new `qmatmulMaybeBias` helper — weight-presence-driven (no-op for qwen3/llama/mistral), wired through both the regular and batched eval paths.
- **Tokenizer merges** (`tokenizer.zig`): accept the older space-joined-string merge format (`"a b"`) alongside the pair-array format (`["a","b"]`). Qwen2.5 / GPT-2-style exports use the string form; reading it as an array in ReleaseFast spun the load loop on garbage and **hung the server during tokenizer init**. Factored into a testable `parseMergePair`.
- **Discovery** (`model_discovery.zig`): add `"qwen2"` to the supported model_types allow-list.

The Swift client already lists `qwen2` as supported, so no app change is needed — this closes the client/server mismatch.

## Testing
- New Zig unit tests: `parseConfigFromJson qwen2` (no QK-norm, head_dim 128) and `parseMergePair` (array + space-joined-string + malformed). Full `zig build test` green.
- **Live end-to-end**: `Qwen2.5-Coder-32B-Instruct-8bit` loads (head_dim 128, 8-bit affine) and generates correct code; reports cleanly in `/v1/models`.

### DIFF
--- a/src/main.zig
+++ b/src/main.zig
@@ -587,19 +587,25 @@ pub fn main(init: std.process.Init) !void {
     };
     chat_config.* = try chat_mod.loadChatConfig(io, allocator, model_dir);
 
-    // Resolve EOS tokens from tokenizer if config.json didn't specify any
-    if (config.num_eos_tokens == 0) {
-        if (chat_config.eos_token) |eos_str| {
-            if (tok.special_tokens.get(eos_str)) |eos_id| {
+    // Merge the tokenizer's chat-terminator EOS into the stop set — ALWAYS,
+    // even when config.json already specified an eos_token_id. Some checkpoints
+    // (e.g. Qwen2.5-Coder-7B) set config.json eos_token_id to <|endoftext|>
+    // (151643) but their chat template ends turns with <|im_end|> (151645);
+    // stopping only on config's id leaks <|im_end|> into the output (breaks
+    // structured-JSON / tool-calling). Additive + dedup-guarded: this can only
+    // ADD a model-declared stop token, never remove one.
+    if (chat_config.eos_token) |eos_str| {
+        if (tok.special_tokens.get(eos_str)) |eos_id| {
+            if (!config.isEosToken(eos_id)) {
                 config.addEosToken(eos_id);
                 log.info("EOS token from tokenizer: {s} (id={d})\n", .{ eos_str, eos_id });
             }
         }
-        // Also add <|endoftext|> if it exists and wasn't already added
-        if (tok.special_tokens.get("<|endoftext|>")) |eot_id| {
-            if (!config.isEosToken(eot_id)) {
-                config.addEosToken(eot_id);
-            }
+    }
+    // Also add <|endoftext|> if it exists and wasn't already added.
+    if (tok.special_tokens.get("<|endoftext|>")) |eot_id| {
+        if (!config.isEosToken(eot_id)) {
+            config.addEosToken(eot_id);
         }
     }
 
--- a/src/model.zig
+++ b/src/model.zig
@@ -1246,6 +1246,31 @@ test "ModelConfig addEosToken max capacity" {
     try testing.expect(!config.isEosToken(999));
 }
 
+test "EOS merge: chat-terminator added even when config already provided an eos" {
+    // Regression for the Qwen2.5-Coder-7B leak: its config.json sets
+    // eos_token_id=<|endoftext|> (151643), but its chat template ends turns
+    // with <|im_end|> (151645). The load path (main.zig / scheduler doLoad)
+    // must ALWAYS merge the tokenizer's chat-terminator EOS — additively and
+    // dedup-guarded — not only when config provided none; otherwise <|im_end|>
+    // is never a stop token and leaks into the output (broke structured JSON /
+    // tool calling). This pins the merge invariant those call sites implement.
+    var config = ModelConfig{};
+    config.addEosToken(151643); // from config.json eos_token_id
+    try testing.expectEqual(@as(u32, 1), config.num_eos_tokens);
+
+    // Merge step the fix performs: add the chat terminator if absent.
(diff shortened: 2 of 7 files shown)


ddalcu:
thanks for this @ujwal-setlur, have you also tested with the tests integration folder (you will need to download a matrix of models, and depends how much ram+hdd space you have on your machine)

Main thing we need to test on this (and other + future pr's on zig code):

1. [llmprobe](https://github.com/ddalcu/responses-chat-messages-validator) test, to make sure we remain compatible with all api types.
2. model architecture test across all families.
3. quant test
4. and most important is the bench.sh test to make sure we don't lose any performance.

ujwal-setlur:
@ddalcu I will do that later today. Thanks!

ujwal-setlur:
Ran the integration suite. The short version: llmprobe surfaced a real EOS bug, which I fixed in this PR, and a cross-family pass shows no regression.

### llmprobe (responses-chat-messages-validator), per arch

| Model | Arch | llmprobe |
|---|---|---|
| Qwen2.5-Coder-7B-Instruct-4bit | qwen2 | 108/112 |
| Qwen3-Coder-30B-A3B-8bit | qwen3_moe | 112/112 |
| Qwen3-4B-4bit | qwen3 (dense) | 112/112 |
| gemma-3-12b-it-qat-4bit | gemma3 | 112/112 |
| Llama-3.2-3B-Instruct-4bit | llama | 91/112 |

### EOS bug found + fixed (now in this PR)
llmprobe showed the tokenizer's chat-terminator EOS was only merged into the stop set when config.json provided **no** `eos_token_id`. Qwen2.5-Coder-7B sets config.json `eos_token_id=<|endoftext|>` but ends turns with `<|im_end|>`, so `<|im_end|>` leaked into output → invalid structured JSON + tool failures. Fix: always merge the tokenizer's chat EOS (additive, dedup-guarded) in both load paths (`main.zig` + scheduler `doLoad`), with a `ModelConfig` regression test. That took the 7B from 102 → 108/112.

### No regression (the important part)
- **Llama-3.2-3B is byte-identical on `main` and this branch** — 91/112 with the same 7 failures. Those are a weak 3B model (tool-calling / unicode round-trip / long-context needle), present on `main` too — not arch-related.
- gemma3, qwen3 dense, and qwen3_moe are all 112/112.
- **Perf A/B** on Qwen3-4B (dense — the only inference path this PR touches, via the `qmatmulMaybeBias` null-check): `main` 93.7 → branch 100.6 tok/s. No regression. MoE / hybrid paths are untouched by construction.

### The remaining qwen2 misses are model behavior, not the arch
- Qwen2.5-Coder-7B's 4 failures are all **parallel/streaming tool-call emission**; single tool calls and structured JSON pass.
- Confirmed on the 32B as well: single + streaming tool calls work, but a "call the tool for each" prompt emits **one** call. Qwen2.5-Coder simply doesn't emit parallel tool calls (Qwen3 strengthened this, which is why qwen3-coder passes `tool-parallel`). The server parser handles whatever the model emits.
- The 32B's earlier wholesale llmprobe failures were purely decode speed (32B dense ~6 tok/s) against the 150s per-check timeout — not correctness.

Also added a `qwen2` row to `tests/test_validator_matrix.sh`.

### Gaps / notes
- Didn't run the llama-engine GGUF, ds4, or nvfp4 cells (no local weights).
- One gemma checkpoint (`mlx-community/gemma-3-4b-it-4bit`) fails to load, but that reproduces on `main` and is a separate pre-existing bug — filed as #43, unrelated to this PR.

PR #38, opened 2026-06-14
Report full metadata for unloaded models in /v1/models

### LINKED ISSUE #37: /v1/models reports no context/capabilities for unloaded models
For **unloaded** (stub) entries, `/v1/models` returns only `id`, `state`, `bytes_on_disk`, and `architecture` — no `context_length`, `model_max_tokens`, `is_moe`, dimensions, or `capabilities`. Those fields only populate once a model is loaded, because `server.renderModelEntry` reads them off the *resident* config/chat_config.

So a client can't tell a model's context window or whether it supports tools/vision until it's faulted into memory. Example (`--model-dir` with one loaded, the rest stubs):

```
Qwen3-Coder-30B-A3B-Instruct-8bit (loaded): ctx=154506 caps=[chat,tool_use,...] is_moe=true ...
Qwen3-Coder-Next-4bit (unloaded):           ctx=null    caps=null              is_moe=null ...
```

All of it is in `config.json` (+ chat-template presence) and readable without loading weights — discovery already opens `config.json` to read `model_type`. This also bites the Zed model-list sync helper, which falls back to a guessed context window for any model that isn't currently loaded.

Fix in PR: read the stub metadata from disk and emit dims / context / MoE / quant / capabilities (chat, tool_use, streaming, json_schema, vision) for unloaded entries too. Reasoning/audio stay load-gated (need the live template/encoder).

ujwal-setlur:
Fixes #37.

`/v1/models` returned only `id`/`state`/`bytes_on_disk`/`architecture` for unloaded (stub) entries — context window, dimensions, MoE flag, and capabilities only appeared once a model was loaded (the renderer read them off the resident config). Clients (and the Zed model-sync helper) therefore couldn't see a model's real context window or tool support until it was faulted into memory.

All of that lives in `config.json` (+ chat-template presence) and is cheap to read without loading weights — discovery already opens `config.json` for `model_type`.

## Changes
- `model_discovery.parseStubMeta` (pure, unit-tested) + `readStubMeta` (thin I/O wrapper): extract vocab/hidden/layers, `max_position_embeddings`, quant bits, MoE-ness (`num_experts`/`num_local_experts`/`n_routed_experts`), vision (`vision_config` on a non-`_text` arch), and chat-template presence (`chat_template.jinja` or a `chat_template` key in `tokenizer_config.json`).
- `server.renderModelEntry` stub branch now emits dims, `context_length` + `model_max_tokens`, `is_moe`, `quantization`, and capabilities — `chat`/`tool_use`/`streaming`/`json_schema` (gated on a chat template, the same rule the loaded path uses) plus `vision` + image modality.

`/v1/models` isn't a hot path, so reading a small JSON per unloaded entry is negligible. Reasoning/audio remain load-gated (they need the live template/encoder), so a stub may under-report those two; everything else now matches the loaded entry. Unloaded `context_length` reports the model's max (`max_position_embeddings`) since the memory-bounded effective window isn't known until load.

## Testing
- New `parseStubMeta` unit tests (dims/ctx/quant/MoE extraction; chat gated on template + suppressed for `bert`; vision gated on non-`_text`; malformed → `found=false`). Full `zig build test` green.
- **Live**: `--model-dir` with one model loaded — the unloaded sibling now reports `caps=[chat,tool_use,streaming,json_schema]`, `ctx`, `is_moe`, `quant`, and dims (previously all null).

### DIFF
--- a/src/model_discovery.zig
+++ b/src/model_discovery.zig
@@ -326,6 +326,105 @@ pub fn probeModelDir(io: std.Io, allocator: std.mem.Allocator, abs_path: []const
     return .{ .model_type = model_type, .bytes_on_disk = if (bytes_ok) bytes else null };
 }
 
+/// Metadata an `unloaded` stub can advertise via `/v1/models` without faulting
+/// in weights — all sourced from `config.json` (+ chat-template presence). Lets
+/// clients see context window, dims, MoE-ness, and capabilities (tools/vision)
+/// before a cold load. `found=false` means config.json couldn't be read/parsed.
+pub const StubMeta = struct {
+    found: bool = false,
+    vocab_size: u32 = 0,
+    hidden_size: u32 = 0,
+    num_hidden_layers: u32 = 0,
+    max_position_embeddings: u32 = 0,
+    quant_bits: u32 = 0,
+    is_moe: bool = false,
+    has_vision: bool = false,
+    has_chat: bool = false,
+};
+
+fn jsonU32(obj: std.json.ObjectMap, key: []const u8) u32 {
+    if (obj.get(key)) |v| {
+        if (v == .integer and v.integer > 0) return @intCast(v.integer);
+    }
+    return 0;
+}
+
+/// Pure: extract `StubMeta` from raw config.json bytes. `has_chat_template` is
+/// supplied by the caller (the template lives outside config.json), and
+/// determines chat/tool capabilities — gated off for encoder-only (`bert`)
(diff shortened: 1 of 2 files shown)


ddalcu (approved):

ddalcu:
thanks @ujwal-setlur , we should add a test for this in my https://github.com/ddalcu/responses-chat-messages-validator project, I use that to make sure mlx-serve api remains true and valid with frontier api compatibility

PR #48, opened 2026-06-18
Fix double-free SIGSEGV on the GPU-OOM pre-flight refusal path (#45/#47)

### LINKED ISSUE #47: SIGSEGV during shutdown on the GPU-OOM pre-flight refusal path (#45)
# SIGSEGV on the GPU-OOM pre-flight refusal path (#45)

**Repo:** ddalcu/mlx-serve · **Affected commit:** `2bb9dbf` ("Handle GPU OOM at model load gracefully (pre-flight + legible error) (#45)"), present through current HEAD `540fd5a`.
**Severity:** Low-impact but ugly — the process was already exiting (model can't load), but it dies with `Segmentation fault: 11` (exit signal 11) *after* printing the intended legible error, instead of exiting cleanly with an error code. A process supervisor will report a crash rather than a graceful refusal.

## Summary

When the #45 memory pre-flight refuses a load (`error.InsufficientMemory`), the legible error is printed correctly — then the process **segfaults during shutdown** in `chat.ChatConfig.deinit`, on `main.main`'s normal defer-unwind. The success path never runs this deinit (it exits via the signal handler), so #45's new error-return path is the first code to exercise it, exposing a latent bad free.

## Backtrace (EXC_BAD_ACCESS / SIGSEGV, main thread)

```
mem.Allocator.rawFree
mem.Allocator.free__anon_2651
chat.ChatConfig.deinit          # frees self.chat_template (always heap-allocated)
main.main                       # defer-unwind after `try serve(...)` returns error
start.callMain
dyld  start
```

Reproduced **8/8** on a clean build of `540fd5a` with no local modifications.

## Deterministic reproducer (no memory pressure required)

The pre-flight sums `*.safetensors` sizes via `statFile` (it never opens them — `modelDiskBytes` in `scheduler.zig`), so a **sparse** dummy file inflates the apparent weight size past free RAM and forces the refusal with zero real memory/disk cost:

```bash
REAL=~/.mlx-serve/models/mlx-community/gemma-4-e2b-it-8bit   # any real model dir
FAKE=/tmp/fake-oom-model
rm -rf "$FAKE"; mkdir -p "$FAKE"
for f in "$REAL"/*; do ln -sf "$f" "$FAKE/$(basename "$f")"; done
truncate -s 60G "$FAKE/zzz_dummy.safetensors"   # sparse: ~0 real bytes, statFile reports 60 GB

./zig-out/bin/mlx-serve --model "$FAKE" --serve --port 11296 --no-pld
echo "exit: $?"   # => 139 (128 + SIGSEGV), after "Insufficient memory to load model…"
```

Equivalent real-world trigger: launch a model that genuinely fails the pre-flight (e.g. restart before a prior instance frees its weights, leaving free RAM below `weights + weights/7 + 1 GB`).

## Mechanism

1. `doLoadOnInferenceThread` (`scheduler.zig` ~L1791) returns `error.InsufficientMemory` from the pre-flight. This is **before** the entry-install (`entry.config = …; entry.chat_config = …` at ~L2020), and all the per-pointer `errdefer`s are registered after the return point, so none run.
2. `Scheduler.init` joins the inference thread and returns `error.LoadFailed`; `serve()` propagates it; `main.main`'s `try server_mod.serve(...)` returns the error.
3. `main.main` unwinds its defers. The ownership-transfer guard at `main.zig` ~L693 only flips `chat_config_owned_by_registry = true` **if `entry.config != null`** — which the pre-flight abort never sets. So the guard stays false and the defer at `main.zig` ~L584 runs `chat_config.deinit(); allocator.destroy(chat_config);`.
4. `ChatConfig.deinit` (`chat.zig`) frees `self.chat_template`, and **that free faults**.

### Open sub-question for the maintainer

`chat_template` is always heap-allocated (`allocator.dupe`, including the empty-string fallback), the inference thread does not free it on this path, and the registry entry does not own it on this path — so a *single* free of it should be valid. Yet it faults deterministically (8/8), and under Guard Malloc (`libgmalloc`) the fault does not reproduce — both signatures of a layout-sensitive heap-corruption / double-free. Pinning the exact first-free site needs a dynamic trace (lldb breakpoint on `free` for that pointer, or `MallocStackLogging` + `malloc_history`). The static suspects are the ownership accounting at `main.zig` L584/L693 vs. the discovery-entry path at L684 (`registry.peek`).

## Suggested directions

- Make the ownership signal robust to the pre-flight abort (don't key the transfer solely on `entry.config != null`, or free `chat_config` exactly once on every exit path), **and/or**
- Run the pre-flight in `main.zig` (or earlier in `serve()`) **before** `chat_config`/`tok`/`config` ownership is in flight, so the refusal returns through a single, well-defined cleanup path.

## Not related to observability

Discovered while baselining a separate observability feature branch onto current upstream. Confirmed the crash reproduces identically on **plain `540fd5a`** with none of that code present (the gauge-sampler thread never spawns on this path — `serve()` errors at `Scheduler.init` before the spawn; no `"Prometheus metrics: ENABLED"` log line is emitted). Filing purely as an upstream finding.

bitworks-io:
Fixes #47.

## Problem

When the #45 GPU-memory pre-flight refuses a load (`error.InsufficientMemory`), `serve()` returns the error and `main()` unwinds its defers — and the process **segfaults during shutdown**, *after* printing the intended legible error:

```
mem.Allocator.rawFree
mem.Allocator.free__anon_2651
chat.ChatConfig.deinit
main.main
```

It's `EXC_BAD_ACCESS / SIGSEGV` (exit 139) — the process dies with a crash signal instead of exiting cleanly with an error code, so a supervisor reports a crash rather than a graceful refusal.

## Root cause

The serve-path cleanup for `config`, `tok`, and `chat_config` in `main.zig` was registered as **both an `errdefer` and a `defer` with identical bodies**, e.g.:

```zig
errdefer if (!chat_config_owned_by_registry) { chat_config.deinit(); allocator.destroy(chat_config); };
defer    if (!chat_config_owned_by_registry) { chat_config.deinit(); allocator.destroy(chat_config); };
```

On an **error** return Zig runs *both* the `errdefer` and the `defer`, so each resource is freed twice → double-free. `chat_config` crashes first because it's registered last (LIFO unwind), matching the backtrace.

This was latent before #45: the success path exits via the signal handler and never unwinds these defers, so the redundant pair was never exercised until the pre-flight gave `serve()` an error to return.

## Fix

Drop the redundant `errdefer` for all three resources. A plain `defer` already runs on the error-return path, and the existing `*_owned_by_registry` runtime guard keeps the single cleanup correct on every exit:

- serve-mode → registry takes ownership, guard skips the free;
- offline mode (`--prompt`) → normal return, `defer` frees;
- error return (incl. the pre-flight refusal) → `defer` frees exactly once.

The other serve helpers (ds4 / llama stubs) already use `errdefer`-only and are unaffected.

## Reproducer (no memory pressure needed)

The pre-flight sums `*.safetensors` sizes via `statFile` (it never opens them), so a **sparse** dummy weight forces a deterministic refusal:

```bash
REAL=~/.mlx-serve/models/<any-model>
FAKE=/tmp/fake-oom; rm -rf $FAKE; mkdir -p $FAKE
for f in "$REAL"/*; do ln -sf "$f" "$FAKE/$(basename "$f")"; done
truncate -s 80G "$FAKE/zzz_dummy.safetensors"

./zig-out/bin/mlx-serve --model "$FAKE" --serve --port 11296
echo "exit: $?"   # before: 139 (SIGSEGV) ;  after: 1 (clean error exit)
```

## Verification

- Added `tests/test_oom_preflight.sh` (model-gated like the other integration tests, so it skips in CI without a model). Verified **red-on-revert**: reverting only `main.zig` brings back exit 139; with the fix, exit 1.
- `zig build test` green; `zig build -Doptimize=ReleaseFast` clean.

## Note (separate, pre-existing — not addressed here)

Those same `defer`s call `.deinit()` on memory that is only initialized on the *next* line (`chat_config.* = try loadChatConfig(...)`). If `loadChatConfig` / `loadTokenizer` / `parseConfig` themselves fail, the `defer` runs `deinit()` on an uninitialized struct. This is pre-existing and a different trigger than #47 (this change actually halves the exposure, 2 deinits → 1); happy to address it in a follow-up if you'd like.

### DIFF
--- a/src/main.zig
+++ b/src/main.zig
@@ -545,7 +545,10 @@ pub fn main(init: std.process.Init) !void {
     // explicit defer on `config_storage`.
     const config_storage = try allocator.create(model_mod.ModelConfig);
     var config_owned_by_registry = false;
-    errdefer if (!config_owned_by_registry) allocator.destroy(config_storage);
+    // defer-only, NOT errdefer + defer: a plain `defer` already runs on the
+    // error-return path, so pairing it with an errdefer that has the same body
+    // frees the resource twice on error (double-free / SIGSEGV). The runtime
+    // `owned_by_registry` guard makes the single defer correct on every exit.
     defer if (!config_owned_by_registry) allocator.destroy(config_storage);
     config_storage.* = try model_mod.parseConfig(io, allocator, model_dir);
     const config = config_storage;
@@ -564,10 +567,8 @@ pub fn main(init: std.process.Init) !void {
     log.info("Loading tokenizer...\n", .{});
     const tok = try allocator.create(tokenizer_mod.Tokenizer);
     var tok_owned_by_registry = false;
-    errdefer if (!tok_owned_by_registry) {
-        tok.deinit();
-        allocator.destroy(tok);
-    };
+    // defer-only (see config note above): errdefer + defer with the same body
+    // double-frees on the error-return path.
     defer if (!tok_owned_by_registry) {
         tok.deinit();
         allocator.destroy(tok);
@@ -577,10 +578,9 @@ pub fn main(init: std.process.Init) !void {
     // Load chat config — heap-allocated, ownership transfers to registry on serve_mode.
     const chat_config = try allocator.create(chat_mod.ChatConfig);
     var chat_config_owned_by_registry = false;
-    errdefer if (!chat_config_owned_by_registry) {
-        chat_config.deinit();
-        allocator.destroy(chat_config);
-    };
+    // defer-only (see config note above): errdefer + defer with the same body
+    // double-frees on the error-return path — this is the one that crashed in
+    // the #45 GPU-OOM pre-flight refusal (ChatConfig.deinit ran twice).
     defer if (!chat_config_owned_by_registry) {
         chat_config.deinit();
         allocator.destroy(chat_config);
--- a/tests/test_oom_preflight.sh
+++ b/tests/test_oom_preflight.sh
@@ -1,0 +1,84 @@
+#!/bin/bash
+# Regression test for the GPU-OOM pre-flight refusal path (issue #47).
+#
+# When the #45 memory pre-flight refuses a load (error.InsufficientMemory),
+# main() returns the error and unwinds its defers. A redundant errdefer+defer
+# pair (identical bodies) for config/tok/chat_config used to free each twice on
+# that error path → double-free → SIGSEGV during shutdown, AFTER the legible
+# error was printed. This pins the fix: the refusal must exit cleanly (no
+# signal-11), not crash.
+#
+# The pre-flight sums *.safetensors sizes via statFile (it never opens them), so
+# a SPARSE dummy .safetensors inflates the apparent weight size past free RAM
+# and forces a deterministic refusal — no real memory pressure required.
+#
+# Usage: ./tests/test_oom_preflight.sh [model_dir] [port]
+#   Needs any loadable model dir for CPU-side setup (config.json + tokenizer).
+
+set -u
+
+MODEL="${1:-$HOME/.mlx-serve/models/mlx-community/gemma-4-e2b-it-8bit}"
+PORT="${2:-11293}"
+BINARY="${BINARY:-./zig-out/bin/mlx-serve}"
+FAKE="/tmp/test_oom_preflight_model"
+LOG=/tmp/test_oom_preflight.log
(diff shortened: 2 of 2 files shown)


ddalcu:
thank you

PR #53, opened 2026-06-20
feat: opt-in Prometheus metrics + zero-dependency admin dashboard (--metrics)

bitworks-io:
## Opt-in observability: Prometheus metrics + zero-dependency admin dashboard

Adds an **opt-in** observability layer behind a new `--metrics` flag. Off by default, and engineered so it never touches mlx-serve's serving performance.

### Design contract
- **Off (default):** a single `?*Metrics` null-check per *request* — never per token. No allocation, no atomics, no work.
- **On:** a handful of relaxed (`.monotonic`) atomic adds per request, recorded at the existing `finishSlot` funnel, which is already off the per-token decode path. The inference thread never locks, waits, or blocks on metrics. Gauges are sampled on a dedicated thread; `/metrics` renders on the scrape connection thread.

### What's added
- **`GET /metrics`** — Prometheus text exposition. Counters (prompt/generation tokens, success/cancelled requests, prefix-cache queries/hits), latency histograms (TTFT, e2e, prefill, decode), token-count histograms, and gauges (running/waiting requests, GPU %, memory, live generation tokens).
- **`GET /admin`** — a self-contained, zero-dependency web dashboard (single embedded HTML file): throughput, latency, GPU %, memory, real-time tok/s with sparklines, plus a ready-to-paste Prometheus scrape config.
- **`GET /admin/metrics.json`** — the same metrics as JSON for the dashboard / custom tooling.
- **`POST /admin/evict-prefix-cache`** — async prefix-cache eviction, optionally protected by `--admin-key` (bearer token, constant-time compare).

### Notable details
- **Real-time tok/s:** a `generation_tokens_live` gauge (completed tokens + tokens generated so far by in-flight slots) sampled every 2 s, so the dashboard shows a live rate even mid-request instead of reading 0 between long-request completions.
- **`memory_mb` uses phys_footprint** (TASK_VM_INFO flavor 22), the only figure that reflects a loaded model's true Metal/IOKit footprint on Apple Silicon — `resident_size` badly under-reports it.
- **Metric namespace:** emitted under `vllm:` so existing vLLM Grafana dashboards work out of the box. Happy to switch everything to `mlx_serve:` if you'd prefer project-native names — let me know.

### Tests
- `tests/test_metrics.sh` — 25 checks: 503-when-off, Prometheus shape, counters/histograms/gauges increment after a request, and the live-gauge invariants (`live > 0` after a request, `live == total` at rest).
- `tests/test_admin_api.sh` — 19 checks: JSON shape, dashboard serving, and the full `--admin-key` auth matrix (open mode / missing header → 401 / wrong key → 401 / correct key → 200).
- Unit tests in `src/metrics.zig` (render/atomics) and `src/status.zig` (phys_footprint).

All green; `zig build -Doptimize=ReleaseFast` clean. Rebased on current `main`.

### DIFF
--- a/app/Sources/MLXServe/Models/ServerOptions.swift
+++ b/app/Sources/MLXServe/Models/ServerOptions.swift
@@ -21,6 +21,15 @@ struct ServerOptions: Codable, Equatable {
     var logLevel: LogLevel = .info
     var requestTimeout: Int = 300       // seconds; 0 = unlimited
 
+    // Observability (server-launch flags)
+    /// Opt-in Prometheus metrics + zero-dependency admin dashboard. When true,
+    /// launches with `--metrics`, exposing `/metrics` and `/admin`.
+    var enableMetrics: Bool = false
+    /// Optional bearer token protecting admin POST actions (cache eviction,
+    /// etc.). Only emitted (`--admin-key`) when `enableMetrics` is also true —
+    /// never sent standalone.
+    var adminKey: String = ""
+
     // Speculative decoding (server-launch flags)
     var enablePLD: Bool = true          // --pld is default-on now (CLI flips with --no-pld)
     var pldDraftLen: Int = 5
@@ -289,6 +298,8 @@ struct ServerOptions: Codable, Equatable {
         noVision == other.noVision &&
         logLevel == other.logLevel &&
         requestTimeout == other.requestTimeout &&
+        enableMetrics == other.enableMetrics &&
+        adminKey == other.adminKey &&
         enablePLD == other.enablePLD &&
         pldDraftLen == other.pldDraftLen &&
         pldKeyLen == other.pldKeyLen &&
@@ -361,6 +372,12 @@ struct ServerOptions: Codable, Equatable {
         }
         if requestTimeout != 300 {
             args += ["--timeout", "\(requestTimeout)"]
+        }
+        if enableMetrics {
+            args += ["--metrics"]
+            if !adminKey.trimmingCharacters(in: .whitespaces).isEmpty {
+                args += ["--admin-key", adminKey.trimmingCharacters(in: .whitespaces)]
+            }
         }
         // Spec-decode: explicit flags either way so the server's CLI defaults
         // can't drift out from under the UI.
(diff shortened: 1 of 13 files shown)


bitworks-io:
Just a quick note on the intent for this feature addition. I am looking to use mlx-serve on some headless OSX systems and this provides a lightweight observability setup for those cases where not local with the admin UI available.

ddalcu:
cool idea, but it has some issues, like using vllm name in some keys, not including a way to enable this via the swift app. It requires a bit of cleanup.

Also would take this chance to extract current html index, into a html folder. So it can potentially grow with other html based server features.

bitworks-io:
The vllm keys were on purpose, it's what a lot of monitoring solutions use
that are not vllm so they can monitor runtimes. Outside of that you are
right, I focused on the CLI admin

bitworks-io:
Thanks for the review! Addressed both points and rebased onto the latest `main` (now sits on top of the image/video/TTS/Ollama work as one clean commit):

- **Enable it from the Swift app** — added a **Metrics & Dashboard** toggle in Settings that adds `--metrics` to the launch, an optional **Admin Key** field for `--admin-key`, and an **Open Metrics Dashboard** item in the menu-bar menu that opens `/admin`. No longer CLI-only.
- **`vllm:` keys** — kept intentionally: lots of monitoring stacks (Grafana dashboards, alert rules, autoscalers) key off the `vllm:` names to watch *any* OpenAI-compatible runtime, so mlx-serve drops into an existing vLLM dashboard with zero config. Apple-specific extras (GPU util, unified-memory footprint, live tok/s) use `mlx_serve:`.

Verification on the rebased branch: `zig build test`, `test_metrics.sh` 25/25, `test_admin_api.sh` 19/19, Swift `ServerOptions` tests green, app builds. Happy to adjust naming or split further if you'd prefer.

ddalcu:
thanks for updating, I need some time to look into this since its a huge PR and it touches a bit on performance which is very important to me. But we will make it happen !

bitworks-io:
> thanks for updating, I need some time to look into this since its a huge PR and it touches a bit on performance which is very important to me. But we will make it happen !

Built so it would not have any impact on actual LLM performance on purpose, definitely worth more eyes on though.

ddalcu:
merged in main via a different commit, I adjusted the code a bit, removed the admin key, added api-key. and some tweaks here and there, thanks !

PR #62, opened 2026-07-07
fix(loader): clear error on incomplete/empty weights dir, not misleading MISSING WEIGHT

bitworks-io:
## Summary

Pointing mlx-serve at a model directory whose weight shards have not finished downloading produces a misleading error. When a download is interrupted, the small files (`config.json`, `tokenizer.json`, `model.safetensors.index.json`) are typically present while the larger `*.safetensors` shards are not. `loadWeightsOpt` returns an empty weights map, and the first weight lookup in `Transformer.init` — `getWeightFmt("{prefix}.embed_tokens.weight")` — fails:

```
Loaded 0 weights from 0 file(s)
MISSING WEIGHT: model.embed_tokens.weight
```

followed by `unreachable` (undefined behavior under ReleaseFast). Because the message reports whichever weight the architecture loads first, the failure resembles an architecture-specific loader bug rather than an incomplete checkpoint. A complete checkpoint loads and runs normally.

## Change

`loadWeightsOpt` now delegates to `loadWeightsFromOpenDir`, which returns `error.NoWeightFiles` when no usable weights are found:

```
no usable weights loaded from <dir> (0 *.safetensors file(s) found) — the checkpoint looks like
an incomplete download (config/tokenizer present, weight shards missing). Re-download the model
(e.g. `mlx-serve pull <model>`) or delete the dir and re-fetch.
```

This is consistent with the existing "incomplete download" message on the tokenizer load path in `main.zig`, and replaces the previous `unreachable`. Extracting the open-directory variant (parallel to `model_discovery.discoverModelsInDir`) allows the guard to be unit-tested with a temporary directory; the success path is otherwise unchanged.

## Tests

- `loadWeights on a weightless dir …` reproduces the failure (confirmed failing before the guard, passing after).
- `parseConfigFromJson mistral honors explicit head_dim …` records that `head_dim = 128` is taken from the config rather than recomputed as `hidden_size / num_attention_heads` (`5120 / 32 = 160`), along with the flat `"model"` prefix, `quant_bits`, and layer count.

## Verification

- `zig build test` passes.
- `zig build -Doptimize=ReleaseFast` builds cleanly.
- The built server loads a complete model and reports `Model ready`; against a directory with no weight shards it returns `error.NoWeightFiles` and exits without a crash.

## Scope

This guard covers the case where no weight files are present. A partially downloaded checkpoint (some shards present, one missing) still fails with a `MISSING WEIGHT` error for a tensor in the absent shard; validating on-disk shards against `model.safetensors.index.json` would address that and is a reasonable follow-up.

### DIFF
--- a/src/model.zig
+++ b/src/model.zig
@@ -1203,14 +1203,22 @@ pub fn loadWeightsWithVision(io: std.Io, allocator: std.mem.Allocator, model_dir
 }
 
 fn loadWeightsOpt(io: std.Io, allocator: std.mem.Allocator, model_dir: []const u8, load_vision: bool) !Weights {
+    var dir = try std.Io.Dir.openDirAbsolute(io, model_dir, .{ .iterate = true });
+    defer dir.close(io);
+    return loadWeightsFromOpenDir(io, allocator, dir, model_dir, load_vision);
+}
+
+/// Load every `*.safetensors` in an already-open `dir` into a Weights map.
+/// `model_dir` is the on-disk path string, used both to build the per-file
+/// absolute path for `mlx_load_safetensors` and to phrase the error message.
+/// Split out of `loadWeightsOpt` so the incomplete-checkpoint guard below is
+/// unit-testable against a `tmpDir` (mirrors `model_discovery.discoverModelsInDir`).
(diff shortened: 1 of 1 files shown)


ddalcu:
thanks !

PR #69, opened 2026-07-11
fix: bound quantized-KV prefill memory so long context fits

JaySNL:
## Summary

Long prompts **OOM-crashed the server far below the model's real capacity** on a `head_dim=256` MoE model. Two compounding causes:

- `head_dim=256` falls off MLX's fused SDPA kernel onto the score-materializing attention path, whose `[heads, chunk, ctx]` scratch scales with `prefill_chunk × context`.
- With `--kv-quant`, attention dequantizes the KV cache back to dense fp16 per layer, and MLX lazy-eval holds every layer's transient at once.

A single prefill step thus allocated tens of GB. This bounds both, and fixes the now-inaccurate preflight estimate. All new behavior is gated on `head_dim > 128`, so standard-head-dim models keep their full prefill chunk and coarse eval cadence — no speed regression.

## Expected vs actual

- **Expected:** a `head_dim=256` MoE model prefills a long prompt within its advertised context.
- **Actual:** the server OOM-crashed at ~100k tokens, far below the model's real capacity. `head_dim=128` models were unaffected.

## Root cause

- **Prefill chunk (`src/generate.zig`):** `PREFILL_CHUNK` (default 8K) sizes the score-materializing scratch `[heads, chunk, ctx]`. At `head_dim > 128` (no fused SDPA) and long context this scratch alone OOMs the command buffer.
- **Per-layer KV dequant (`src/transformer.zig`, `forwardMoeWith`):** during prefill the MoE forward pass only `eval`s every `MOE_EVAL_EVERY_N_LAYERS` (4) layers. With `--kv-quant`, each layer dequantizes its KV cache to a dense fp16 transient; under MLX lazy-eval four layers' transients accumulate before the barrier.
- **Preflight estimate (`src/server.zig`, `checkAttentionMemory`):** the working-set term scaled with the full `seq` and used a dense/Gemma FFN default, over-counting the real peak ~5×, while the KV term used a fixed dense-fp16 width even under `--kv-quant`.

## Fix

- **generate.zig:** cap the prefill chunk to 512 when `head_dim > 128`; fused head dims (≤128) keep the full chunk for prefill throughput.
- **transformer.zig (`forwardMoeWith`):** force an `mlx_array_eval` barrier every layer (cadence 1) during prefill when `head_dim > 128` (was every 4), so each layer's dense-fp16 KV-dequant transient frees before the next builds — peak bounded to ~1 layer. Decode path unchanged.
- **server.zig (`checkAttentionMemory`):** accurate per-request estimate — MoE-aware FFN, KV scaled by the active kv-quant bits, working-set seq bounded by the prefill chunk (previously full seq + a dense/Gemma default → ~5× over-count).

## Verification

- A **255k-token prompt now serves** on a `head_dim=256` MoE model (was OOM at ~100k).
- A `head_dim=128` model is **unaffected** — same prefill chunk and eval cadence, no measurable change.

## Environment

- macOS 26.2, Apple M4 Max, Apple clang 16
- Zig 0.16.0
- mlx-c 0.6.0, mlx 0.31.2 (Homebrew)

## Scope

Three files (`generate.zig`, `transformer.zig`, `server.zig`), +29/−5. All new behavior is gated on `head_dim > 128`, so standard-head-dim models are untouched — no speed regression. Companion to #68 (both fix headless / long-context serving); independent diffs.

### DIFF
--- a/src/generate.zig
+++ b/src/generate.zig
@@ -742,7 +742,13 @@ pub const Generator = struct {
         // PREFILL_CHUNK overridable via env MLX_SERVE_PREFILL_CHUNK for tuning,
         // or via the module-level `prefill_chunk_override` (set by --prefill-chunk
         // CLI flag in main.zig). Env var wins if both are set.
-        const PREFILL_CHUNK: usize = readEnvUsize("MLX_SERVE_PREFILL_CHUNK", prefill_chunk_override);
+        var PREFILL_CHUNK: usize = readEnvUsize("MLX_SERVE_PREFILL_CHUNK", prefill_chunk_override);
+        // Large head_dim (>128, e.g. 256) misses MLX's fused SDPA kernel and
+        // falls onto the score-materializing path, whose [heads, chunk, ctx]
+        // scratch scales with the chunk - an 8K chunk OOMs the command buffer at
+        // long contexts. Cap the chunk so that scratch stays bounded; fused head
+        // dims (<=128) keep the full chunk for prefill throughput.
+        if (xfm.config.head_dim > 128 and PREFILL_CHUNK > 512) PREFILL_CHUNK = 512;
         // Phase-level prefill instrumentation. Enabled at debug level OR via
         // MLX_SERVE_PREFILL_TRACE=1 (which forces the trace line at info).
         // Phase 0 of plan 04 — gives us a decomposed view of where cold prefill
--- a/src/server.zig
+++ b/src/server.zig
@@ -2122,16 +2122,25 @@ fn checkAttentionMemory(allocator: std.mem.Allocator, stream: *Conn, prompt_len:
     const kv_heads: u64 = config.num_key_value_heads;
     const hdim: u64 = config.head_dim;
     const hidden: u64 = config.hidden_size;
-    const ffn: u64 = @max(config.intermediate_size, config.moe_intermediate_size + config.shared_expert_intermediate_size);
+    const ffn: u64 = if (config.isMoe())
+        config.moe_intermediate_size + config.shared_expert_intermediate_size
+    else
+        config.intermediate_size;
+    // Active KV-quant width in bits: dense fp16 = 16, affine q8 = 8, q4 = 4.
+    const kv_bits: u64 = if (global_scheduler) |sch|
+        (if (sch.kv_quant_config.scheme == .off) 16 else sch.kv_quant_config.bits)
+    else
+        16;
 
     // KV cache (all layers, fp16): layers × 2(K+V) × seq × kv_heads × head_dim × 2 bytes.
     // This is persistent for the rest of the request, so it's a real hard cost.
-    const kv_bytes: u64 = layers * 2 * seq * kv_heads * hdim * 2;
+    const kv_bytes: u64 = layers * 2 * seq * kv_heads * hdim * kv_bits / 8;
     // Per-layer working memory during prefill (fp16). The transformer eval()s every N layers,
     // so transient tensors from earlier layers are released. Peak is bounded by a single layer:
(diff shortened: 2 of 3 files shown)


ddalcu:
Thanks for discovering this! I will have a different PR to fix this, it needs some Fable time.

Here's the whole thing in plain English, if you send a really long prompt (like a 250,000-token document) to your server, the whole server process crashes instead of answering. Not a graceful error, the process just dies, because it runs out of GPU memory in a way that can't be caught.

Why does it run out? Two reasons stack up:

1. A math shortcut models can't use. Apple's MLX library has a fast, memory-efficient attention kernel, but it only works for models with a "head size" of 128 or less. Almost every model we run (all the Gemma-4s, all the Qwen 3.5/3.6s) uses head size 256. So they fall back to a slow path that builds a giant temporary "scores" table in memory. The size of that table grows with how much text you're processing at once × how long the total prompt is. At 250K tokens it's ~67 GB *for one temporary table* and boom, dead server. I verified this with a live experiment: head size 128 uses ~0 extra memory, head size 256 allocates the whole giant table.
2. Cleanup happens too rarely. During prompt processing, the engine only "flushes" its temporary memory every 4 layers of the model. So roughly 3 of those giant tables can be sitting in memory at the same time, tripling the damage.

There was also a third bug in the opposite direction: the server's bouncer (the check that estimates "will this prompt fit in memory?") used a bad formula — it wildly over-estimated for some models (rejecting prompts that would have been fine) while completely missing the giant-table cost (admitting prompts that then crashed the server).


Your diagnosis was correct, but your fix is a sledgehammer: for any model with head size over 128, always process the prompt in tiny 512-token bites and flush memory after every layer, on every request, even short ones. Which greatly affects performance.

Fix:

- Adaptive bite size: Instead of always taking tiny bites, the engine now calculates how big that temporary table would be for this specific prompt and only shrinks the bite when it would exceed a 4 GB budget. Normal prompts (chat, agent work, ~anything under ~16K tokens) are processed exactly as before — full speed. A 100K+ prompt automatically gets smaller bites so the table stays small.

Test results are looking good:
+14% throughput while cutting peak 51.6→27 GB 

 1.5K    1538 tok/s, chunk 8192    -> 1533 tok/s, chunk 8192 — untouched 
 5.1K    1294 tok/s, peak 18.1 GB  -> 1305 tok/s — untouched             
 102K  │ 262.7 tok/s, peak 51.6 GB -> 298.8 tok/s (+14%), peak 27.0 GB   

Fixed via:
https://github.com/ddalcu/mlx-serve/pull/72

Added you as a co-author, thank you !

ddalcu:
Also reported this bug here, so hopefully they will fix it directly in the mlx library https://github.com/ml-explore/mlx/issues/3658

JaySNL:
Nice — the #72 rewrite is cleaner than our chunk cap (keeping chunk 8192 while halving peak memory is the right call), and filing the root cause at ml-explore/mlx#3658 is the fix that actually matters. If the fused-SDPA `head_dim > 128` path gets handled upstream, long context gets faster for everyone for free.

On why we lean on mlx-serve: it's the local-inference backend for a multi-CLI agent setup — a frontier model orchestrates and offloads the grunt work (code lookups, scaffolding, summaries, bulk edits) to local Qwen3-30B-A3B models served here, with tasks brokered over a shared bus between the CLIs. On-demand load-by-name + prefix cache + kv-quant are exactly what make that handoff cheap, so **headless serve is our hot path** — which is why we kept hitting edges in precisely those corners (kv-quant, cold-load flags, long-context). Genuinely great tool to build that on.

Looking forward to more finds and optimization wins — we'll keep sending fixes upstream as we hit them.

Issue #89, opened 2026-07-18
MCP servers need the vsock guest transport (kernel kernels-v3 + vz-agent); the guest booted the legacy console shell


OliverArmstrongDev:
Problem:
_⚠️ MCP startup issues:
• filesystem: MCP servers need the vsock guest transport (kernel kernels-v3 + vz-agent); the guest booted the legacy console shell
No MCP servers are connected — the model has no MCP tools available for this turn. Open the gear icon on the MCP pill to fix or disable broken servers._

MCP works fine in non-sandbox mode and not in sandboxed mode (I understand why with playwright etc.). 

I like keeping agents isolated from my host (e.g. I run opencode in docker to keep it isolated from my Mac), so I was wondering if there is a way to get MCP tools to work in sandboxed mode? - even if it's just the 'filesystem' MCP tool, or even being able to add a custom external MCP server?

TIA

ddalcu:
thanks for reporting this, yep... should be a easy fix. 

As far as running opencode/pi/claude/hermes/openclaw INSIDE the sandbox, totally doable, would be a cool feature !  Will add it to my todo.

OliverArmstrongDev:
No worries, thanks for creating an awesome app!

Opencode etc in the sandbox will be great, one less docker container to run 😊

I'll keep a lookout for the changes in future releases.

Thanks again!

OliverArmstrongDev:
@ddalcu  I was thinking about this a little more and perhaps the MCP Marketplace should be disabled when agent is in sandbox mode as playwright, docker etc. cannot be used due to being isolated from the host?

Also as per my original request, being able to add a custom MCP server (Fields: URL, API, custom headers) that IS accessible to sandboxed agent, would be great and would open up the functionaility even more. :)

ddalcu:
Hi @OliverArmstrongDev , Check out the new PR, https://github.com/ddalcu/mlx-serve/pull/91  Its pretty simple to build if you have Xcode installed.  Added your suggestions above. 

As for playwright, I havent tested it yet, but you should be able to go inside the sandbox via terminal, install it, and it should work.

I could use some help on testing it if you are willing. And just let me know if any adjustments need to be made. Or open a PR towards that branch.

Issue #84, opened 2026-07-17
Speculative Decoding Failure with Gemma-4-31B and Assistant Drafter


nunez:
## Summary
Speculative decoding fails with an `MLX error: expected a non-empty mlx_array` when using a Gemma-4-31B-Instruct model with a corresponding assistant drafter model. Disabling the drafter allows the server to start successfully.

## Environment
- **OS**: macOS
- **MLX Serve Version**: (Check with `mlx-serve --version`)
- **Main Model**: `mlx-community/gemma-4-31b-it-bf16`
- **Drafter Model**: `mlx-community/gemma-4-31B-it-assistant-bf16`
- **MLX Core App**: `/Applications/MLX Core.app`

## Command That Fails
```bash
"/Applications/MLX Core.app/Contents/MacOS/mlx-serve" \
  --model ~/.omlx/models/mlx-community/gemma-4-31b-it-bf16 \
  --serve --port 8080 --host 0.0.0.0 --log-level info \
  --model-dir ~/.omlx/models/mlx-community \
  --metrics \
  --no-vision \
  --drafter ~/.mlx-serve/models/mlx-community/gemma-4-31B-it-assistant-bf16 \
  --draft-block-size 4 \
  --max-concurrent 4 \
  --prefix-cache-entries 8 \
  --prefix-cache-disk 10GB \
  --temp 0.8 \
  --top-p 0.95
```

## Error Output
```
Model: gemma4 (60 layers, 5376-dim, head_dim=256, 32h/16kv, 0-bit affine quant)        
Loading tokenizer...  
  parsed tokenizer (30 MB) in 105ms
  loaded 262144 vocab entries in 6ms
  loaded 514906 merges in 15ms
Tokenizer loaded: 262144 vocab, 514906 merges, 24 special tokens (SentencePiece BPE)
User turn marker: "<|turn>user\n" -> 3 tokens
[registry] max_resident_models=3, max_resident_mem=375.0 GB
[preflight] weights ~58.25 GB, available 487.12 GB
Loading model-00012-of-00012.safetensors...
... (all 12 shards loaded)
Loaded 832 weights from 12 file(s)
Precomputing layer weights...
Batch eval all weights: 2069ms
GELU compiled (kernel fusion enabled)
GeGLU compiled (kernel fusion enabled)
Softcap compiled (kernel fusion enabled)
Loading model.safetensors...
Loaded 48 weights from 1 file(s)
[...]

Issue #81, opened 2026-07-16
MTP head crashes with "Biases must be provided for affine quantization" when the target model's lm_head is mxfp8-quantized


usenetj23:
### Summary
`headQuantParams()` in `src/mtp.zig` calls `transformer_mod.affineParamsFromGeometry()` before checking the target model's actual `quant_mode`. That function infers `bits`/`group_size` purely from tensor shape and always returns `mode = .affine` when the shape math resolves to a plausible (bits, group_size) pair — it never inspects the scales tensor's dtype or the model config's declared quantization mode.

For a target model whose `lm_head` is genuinely **mxfp8**-quantized (scales dtype `uint8`, no `.biases` tensor), the shape math can coincidentally match a valid affine configuration. Concretely, for `lm_head.weight` shape `[vocab, hidden*8/32]` (U32-packed) and `lm_head.scales` shape `[vocab, hidden/32]` (U8), with `hidden=5120`:
- `bits = (w_cols * 32) / in_dim = 8`
- `gs = in_dim / s_cols = 32`

Both land in `affineParamsFromGeometry`'s accepted `bits`/`gs` switch cases, so it returns `{bits: 8, group_size: 32, mode: .affine}` — a false positive. `headQuantParams` then uses this result directly (`src/mtp.zig` around line 1050-1061) without ever falling through to the config-driven branch that would have correctly returned `mode: .mxfp8`.

`targetLmHead()` then calls `mlx_quantized_matmul(..., qp.mode.cstr(), ...)` with `mode="affine"` against a tensor that has no `.biases` (mxfp8 has none by design — see the comment at the top of `model.zig`: "non-affine modes store NO `.biases` tensors"). This throws:
```
MLX error: [dequantize] Biases must be provided for affine quantization. at .../mlx-c/mlx/c/ops.cpp:1106
```

### Reproduction
- Model: `dealignai/Qwen3.6-27B-MXFP8-CRACK-MTP` (base: `Qwen/Qwen3.6-27B`), MXFP8 8-bit/gs32 quantization, MTP weights embedded in the main safetensors shards (not shipped as a sidecar).
- Built a native sidecar with `tests/build_mtp_sidecar.py Qwen/Qwen3.6-27B <out>` (unmodified, against the original upstream bf16 repo) — this step works fine, sidecar loads cleanly (`[mtp] loaded native MTP head (dense-mlp; ...)` logs with no error).
[...]

Issue #79, opened 2026-07-15
Is it possible to serve EmbeddingGemma in /embeddings  using mlx-serve?


ajram23:
Looks like it still needs 

Parse and honor:

- use_bidirectional_attention: true
- tokenizer add_bos_token / add_eos_token or tokenizer post-processor

Load/apply EmbeddingGemma projection:

- dense.0
- dense.1
- then existing mean-pool + L2-normalize
-

ddalcu:
yep, should be pretty easy, do you want to attempt it and open a PR :) ?

PR #91, opened 2026-07-19
26.7.10 - Insanely fast, LAN Sharing, Sandbox Pi & Hermes with 1 click, Laguna support

ddalcu:
- **LAN model sharing.** Turn it on and every Mac on your network can use the models this Mac hosts, chat and image / speech / music / video / 3D generation alike. Macs find each other over Bonjour, shared models show up in every picker as "model · peer", requests stream to the Mac that has the model. Off by default, you pick what to share, and only inference is exposed (management & metrics stay private).
- **Run coding agents inside the sandbox.** The Sandbox window now has a real terminal, so you can run pi, hermes or a plain shell fully inside the isolated Linux VM talking to your local model, nothing runs on your Mac. First run installs the CLI in the guest, and there's a copyable ssh command if you'd rather use your own terminal app. The sandbox image is now pinned (the custom base-image setting is gone), which also fixes upgrades being stuck on the old pre-ssh image with an "image out of date" popup that re-pull and reset couldn't clear. You can also switch models mid-session with `/model`, and sandboxed agents can now use a LAN-shared model too. MCP servers work inside the sandbox now as well; the guest agent they connect through was missing from released builds, so they could never start before. (#89)
- **Laguna support.** poolside's Laguna S 2.1, a 117.6B-A8.5B MoE coding model, runs end to end: chat, thinking, tool calls, and long context. (faster than other engines, but working on making it faster next release)
- **oQ checkpoints get native MTP too.** mlx-serve now reads MTP heads straight out of oMLX's own oQ-format checkpoints, no sidecar conversion needed. Also fixed a head-norm bug that was quietly tanking the acceptance rate on these. Net result: mlx-serve now beats oMLX's own MTP on oMLX's own checkpoint.
- **M5 neural accelerator support.** We now build MLX ourselves with Apple's NAX kernels enabled (the stock package ships them disabled) and added our own NAX kernels for speculative verification, 1.2-2.2x faster than the stock path, drafting up to depth 8 on M5. M1-M4 Macs keep their existing tuned path. Prefill also got noticeably faster across the board from new fused attention and quantized matmul kernels, independent of the M5 work. Heads up: all builds now require macOS 26.2+.
- **EmbeddingGemma support.** Google's EmbeddingGemma models now work on `/v1/embeddings`, validated at 0.98+ parity vs the reference implementation. (#79)
- **Speculative decoding fixes.** Fixed a Gemma BF16 failure with the assistant drafter (#84), a crash loading mxfp8-quantized MTP heads (#81), and cross-request acceptance seeding being stuck off.
- **Ready for macOS 27.** Builds and runs on the macOS 27 beta (moved to a pinned Zig 0.17 nightly, the build stages it for you), released binaries still run on macOS 26.2+.
- **Closer to the OpenAI spec.** `reasoning_effort` now turns thinking on for chat completions (any effort except "none", budget scales with the level), so standard clients don't need the `enable_thinking` extension anymore. `parallel_tool_calls: false` is now honored with at most one tool call per response (the OpenAI SDK sends it in strict structured-output mode), same for Anthropic's `disable_parallel_tool_use`. Asking for `n` > 1 choices now gets a clear 400 instead of silently returning one choice. The Anthropic API now reports which stop sequence fired (`stop_reason: "stop_sequence"` plus the matched string, non-streaming & streaming). Chat completions usage now reports prompt cache hits in `prompt_tokens_details.cached_tokens` (the cache was always there, it just wasn't visible in the standard field, so cost dashboards and conformance tools read it as "no caching"). Also fixed a small per-request memory leak on responses with no visible content.
- **A few small additions.** Dictation in the Voice Clone tab, and a community model tier list on the website.
- **Bug fixes.** A KV cache crash, a clearer error when a model's quantization isn't supported, a skill-trigger matching bug (#92), and ds4 bumped to its latest upstream for another speed bump.


ddalcu:
image

Notes:
* We are now on Zig 0.17..  @ivanfioravanti has upgraded to Mac 27 Beta, and zig 0.16 won't work there. It was good to upgrade anyway since both Mac 27 and Zig 0.17 are weeks away from GA. Still works on Mac 26 also.
* Zig 0.17 is now "bundled" in the install script. So simply running `app/build.sh` should do everything.
* MLX-C and MLX C++ projects have been added a git submodules, and we compile them, no longer use the brew version.. [this is why we did not see some NAX benefits](https://github.com/ml-explore/mlx/commit/4367c73b60541ddd5a266ce4644fd93d20223b6e) (like prefil speed) 
 
Requirements: Xcode + Metal Toolchain. (Settings->Components->Metal Toolchain), NodeJS (for benchmark)

Same testing as was done in https://github.com/ddalcu/mlx-serve/pull/83

For the old binary of mlx-serve you can use: `/Applications/MLX\ Core.app/Contents/MacOS/mlx-serve`
And for the new, just `app/build.sh` and use the one in `./zig-out/bin/mlx-serve`

LLM Probe has been updated a bit, so its not the same, but should output the same numbers.
`npx llmprobe@latest http://localhost:11234 --full --bench`


Anyone with a M5/M5Pro/M5Max would be awesome to help out testing this, because we want to make sure our profile is tuned correctly.... I tried looking to see if I can rent one online, but nobody has them.
(Only the performance stuff, I know this PR is huge, ignore the rest of it)

Thanks @CerebralCoding  @AvidDabbler @ivanfioravanti

CerebralCoding:
## M5 Max testing update

I tested the current branch (`804f9b1`) on an **M5 Max with 64 GB RAM**. The bundled Zig 0.17/source-built MLX integration is working correctly, NAX activates as expected, and the new build improves the targeted sustained decode and speculative-decoding workloads without a measurable prefill or TTFT regression once the runs are thermally controlled.

### Build and runtime validation

- Zig `0.17.0-dev.1422+e863bf3be`
- MLX `0.32.0`
- MLX-C `fba4470b8907`
- Runtime reports `nax on (M5 neural accelerators)`
- Zig tests: **982 passed, 78 skipped**
- Staged MLX/NAX checks: **9/9 passed**
- CLI, app bundle, signing and DMG packaging completed successfully

The only local setup requirement was installing and selecting the Metal Toolchain component.

### Test setup and comparison baseline

Throughout the results below, **Old** means the locally installed stable release build:

- Binary: `/Applications/MLX Core.app/Contents/MacOS/mlx-serve`
- mlx-serve version: `26.7.9` (`v26.7.9` release baseline)
- MLX version: `0.31.2`

**New** means the candidate built directly from the current PR branch:

- Binary: `./zig-out/bin/mlx-serve`
- Git commit: `804f9b1`
- Bundled Zig: `0.17.0-dev.1422+e863bf3be`
- Source-built MLX: `0.32.0`
- Source-built MLX-C: `fba4470b8907`

This is therefore a comparison of the installed `v26.7.9` release baseline against the exact PR candidate, not two arbitrary local builds.

Model:

`Youssofal/Qwen3.6-27B-MTPLX-Optimized-Speed`

The initial suite used:

```console
npx --yes llmprobe@latest http://127.0.0.1:11234 --full --bench
```

I then ran a controlled comparison with a fresh server process for every cell and `MLX_SERVE_MTP_EV_SEED=0` for both binaries. The sustained-load sequence was balanced as old -> new -> new -> old.

### Sustained generation and MTP performance

| Metric | Old average | New average | Change |
|---|---:|---:|---:|
| Decode throughput | 45.15 tok/s | 47.80 tok/s | **+5.9%** |
| Predictable MTP | 80.3 tok/s | 90.1 tok/s | **+12.1%** |
| Novel MTP | 43.0 tok/s | 44.2 tok/s | **+2.8%** |
| Speculative ratio | 1.87x | 2.04x | **+9.4%** |

The new binary was also more consistent under sustained load:

- New decode medians: **48.5 and 47.1 tok/s**
- Old decode medians: **45.0 and 45.3 tok/s**, with individual samples ranging from **40.0 to 56.6 tok/s**
- New predictable MTP: **89.4 and 90.7 tok/s**
- Old predictable MTP: **85.8 and 74.8 tok/s**

All four controlled runs passed **100% engine conformance**.

### Prefill and TTFT

The first sequential comparison appeared to show 3.2% lower prefill and 8.4% higher TTFT. That turned out to be a thermal/run-order artifact rather than an inference regression.

The old binary itself varied from **714.6 to 628.3 tok/s prefill** and **191 to 250 ms TTFT** depending on where it ran in the sequence. Those swings were much larger than the apparent old/new difference.

I therefore repeated the comparison from equivalent cooled baselines:

| Metric | Old cooled average | New cooled average | Change |
|---|---:|---:|---:|
| Prefill throughput | 721.9 tok/s | 726.5 tok/s | **+0.6%** |
| Time to first token | 195.0 ms | 192.5 ms | **1.3% faster** |

The raw cooled measurements were:

- Old: **714.6 / 729.1 tok/s prefill**, **191 / 199 ms TTFT**
- New: **719.5 / 733.4 tok/s prefill**, **195 / 190 ms TTFT**

So prefill and TTFT are effectively at parity; there is no measurable regression to fix in the inference path. Under sustained hot conditions, the new binary was also slightly faster than the old binary rather than slower.

### Shared MTP controller state

One separate benchmarking issue did surface. Healthy requests write their acceptance/depth EMAs into model-wide MTP controller state, and later requests inherit that state. Since `llmprobe --full --bench` runs conformance requests before the performance phase, those earlier prompts influence the benchmark's starting MTP depth.

The new M5 profile consistently selected a maximum depth of **8**, but its realized draft depth and predictable throughput varied substantially with inherited state. For example, state-dependent predictable results ranged from roughly **65 to 151 tok/s**. Those values should not be used as the headline old/new comparison.

With cross-request seeding disabled, the repeatable result was **90.1 tok/s for the new build versus 80.3 tok/s for the old build**, or **+12.1%**.

### Controlled ddalcu 4-bit model comparison

I repeated the same controlled comparison on `ddalcu/Qwen3.6-27B-4bit-MTP-MLX-Serve`: two runs per binary, fresh server processes, `MLX_SERVE_MTP_EV_SEED=0`, balanced old -> new -> new -> old order, and a measured GPU starting temperature of approximately 46-47 C before every cell.

The old binary selected MTP depth 6 and the new binary selected the calibrated M5 depth 8 profile.

| Metric | Old A / B | New A / B | Old average | New average | Change |
|---|---:|---:|---:|---:|---:|
| Decode throughput | 52.5 / 53.7 | 58.6 / 55.0 | 53.1 tok/s | 56.8 tok/s | **+7.0%** |
| Time to first token | 204 / 193 | 192 / 196 | 198.5 ms | 194.0 ms | **2.3% faster** |
| Prefill throughput | 721.9 / 719.9 | 755.2 / 731.2 | 720.9 tok/s | 743.2 tok/s | **+3.1%** |
| Predictable MTP | 84.6 / 83.9 | 95.9 / 96.0 | 84.3 tok/s | 96.0 tok/s | **+13.9%** |
| Novel MTP | 42.2 / 48.4 | 51.2 / 46.6 | 45.3 tok/s | 48.9 tok/s | **+7.9%** |
| Speculative ratio | 2.01x / 1.73x | 1.87x / 2.06x | 1.87x | 1.97x | **+5.1%** |

Longer-context decode also generally improved:

| Prompt size | Old decode | New decode | Change | TTFT change |
|---|---:|---:|---:|---:|
| ~464 tokens | 62.3 tok/s | 62.1 tok/s | -0.4% | 2.8% slower |
| ~3.7K tokens | 57.4 tok/s | 70.2 tok/s | **+22.4%** | 2.3% faster |
| ~7.4K tokens | 59.1 tok/s | 64.9 tok/s | **+9.8%** | 1.5% faster |
| ~15K tokens | 57.8 tok/s | 59.8 tok/s | **+3.5%** | 1.5% faster |

All four controlled ddalcu-model runs passed **100% engine conformance** at llmprobe's quick depth. In the earlier full-depth run, one `responses.max_tokens` request returned 15 tokens when 8 were requested; the same new binary passed that full-depth test with the MTPLX model, so that still looks request/model-specific rather than a general API regression.

### Conclusion

The Zig 0.17 build, source-built MLX integration, NAX activation and depth-8 M5 profile all appear healthy on the M5 Max.

The defensible performance conclusion is:

- MTPLX Speed: **+5.9% decode**, **+12.1% predictable MTP**, **+2.8% novel MTP**, with cooled prefill/TTFT effectively unchanged
- ddalcu 4-bit: **+7.0% decode**, **+13.9% predictable MTP**, **+7.9% novel MTP**, **+3.1% prefill**, and **2.3% faster TTFT**

For future comparisons, I would use a fresh server process, disable cross-request MTP seeding, and start each binary from the same GPU-temperature range. Otherwise both controller history and thermal state can overwhelm the actual old/new difference.

ddalcu:
Thanks @CerebralCoding ! Looks good 👍 . 
I'm going to let this one sit here for a day or two while I'm testing it before release.

mifopen:
Hey! Is it possible to make a .dmg a CI artifact? Would make testing so much easier :)

ddalcu:
Good suggestion @mifopen , I adjusted the build to produce a downloadable artifact on PR push, here you go https://github.com/ddalcu/mlx-serve/actions/runs/29791039110 -> click download on the artifact, and DMG will be inside of it.

mifopen:
Error if I try to turn LAN sharing on while having 0 local models downloaded.

ddalcu:
A bit faster now, could probably use some M5 Nax code added for oQe to draft to 8. If anyone wants to attempt that.

M4 numbers, prefil went from 239-> 272 above 2k context windows.  Decode in 2.7.6 235-> 264 this branch  :
image
image


Currently working on speeding up Laguna S and we should be good to merge.

CerebralCoding:
I took a pass at the requested M5 NAX work for oQe:

https://github.com/ddalcu/mlx-serve/pull/99

The underlying problem was that the oQ4e trunk mixes affine q4, q5 and q6 weights. The verification NAX kernel only handled q4, so the q5/q6 projections fell back to stock MLX and made depth 8 too expensive.

The implementation adds q5/q6 unpacking to the M5 verification kernel, but only routes the shapes that benchmark faster. The narrow 1024-output K/V projections remain on stock MLX. It also adds an exact layer/role fingerprint for `Jundot/Qwen3.6-27B-oQ4e` and a separately calibrated MTP cost profile, so unrelated mixed-precision models do not accidentally receive the depth-8 profile.

M5 Max, saturated deterministic echo workload, one discarded warmup and three measured runs:

| Configuration | Decode |
|---|---:|
| Depth 1 | 57.2 tok/s |
| Depth 3 | 95.9 tok/s |
| Depth 6 | 108.9 tok/s |
| Depth 8, stock verification | 96.3 tok/s |
| Depth 8, q4 NAX only | 125.6 tok/s |
| Depth 8, mixed q4/q5/q6 NAX | **136.5 tok/s** |

That makes the completed depth-8 path:

- **25.3% faster** than depth 6
- **8.7% faster** than q4-only depth-8 NAX
- **41.7% faster** than stock depth-8 verification

The 80-iteration interleaved kernel measurements explain the shape routing:

| Projection | Stock | NAX | Ratio |
|---|---:|---:|---:|
| q5 input/Z | 0.446 ms | 0.355 ms | 0.80x |
| q5 Q | 0.516 ms | 0.452 ms | 0.88x |
| q5 K, N=1024 | 0.253 ms | 0.289 ms | 1.14x — stock retained |
| q5 output | 0.604 ms | 0.527 ms | 0.87x |
| q5 down | 0.659 ms | 0.499 ms | 0.76x |
| q6 V, N=1024 | 0.292 ms | 0.304 ms | 1.04x — stock retained |
| q6 down | 0.706 ms | 0.550 ms | 0.78x |

The normal runtime now identifies the checkpoint and starts with `MTP head ready (depth=8)`. A separate real generation produced 75.7 tok/s with depth 8, 108 accepted draft tokens over 52 attempts and 2.08 accepted tokens/attempt. Setting `MLX_SERVE_VERIFY_QMM_NAX_MIXED=0` disables the mixed path and correctly falls back to depth 6.

Validation:

- Zig 0.17 formatting check passed
- **996 tests passed, 81 skipped**
- ReleaseFast build passed
- No calibration-only start-depth or dispatch tracing hooks remain in the submitted code

PR #93, opened 2026-07-22
fix: parse MiniCPM5 V3 XML tool calls, fix streaming leak

uncle9x9:
## Summary

Adds native parsing support for MiniCPM5's V3 XML tool-call format —
`<function name="X"><param name="K">V</param></function>` — across both the
OpenAI-compatible (`/v1/chat/completions`) and Anthropic-compatible
(`/v1/messages`) APIs, non-streaming and streaming.

This format matched none of the existing dialects in `parseToolCalls`: its
marker is `<function` followed by a quoted attribute (never `<tool`), which
also differs from the equals-sign Hermes function-tag form (`<function=NAME>`)
already handled. A dedicated scan, `parseMiniCpm5ToolCalls`, is wired into
`parseToolCalls` the same way as every other dialect (Gemma 4, bare Hermes,
etc.) — no new abstraction, no changes to schema validation or any other
dialect's behavior.

Two streaming-layer gaps are also fixed, since they were the actual cause of
raw tool-call XML reaching clients rather than the parsing gap alone:

- `streamShouldBufferForTools` (used by the chat-completions stream) had no
  awareness of any `<function` marker, so the model's tool-call XML streamed
  to the client token-by-token before the final-text parser ever ran.
  Extending it also incidentally fixes streaming for the pre-existing bare
  Hermes `<function=NAME>` format, which shared the same blind spot.
- The Anthropic `/v1/messages` streaming handler carried its own separate,
  narrower tool-detection check (only `<tool_call`/`<|tool_call`/raw-JSON)
  instead of the shared gate above. Any tag-format dialect it didn't
  recognize — including MiniCPM5 — leaked as a duplicate visible `text` block
  alongside the correctly parsed `tool_use` block. Now routed through the
  same shared gate as the OpenAI path.

### Supported MiniCPM5 tool-call shapes

- Single and multiple string parameters
- CDATA-wrapped values (`<param name="x"><![CDATA[...]]></param>`), including
  values whose CDATA payload itself contains a literal `</param>` or
  `</function>` — the close-tag search is CDATA-boundary-aware
  (`miniCpm5FindCloseTag`), so such content is not truncated
- Multiple sequential `<function>` calls with no outer wrapper
- Duplicate `<param name="K">` — first occurrence wins, consistent with the
  dedup rule already used by the Hermes and Hy3 parsers
- Truncated calls (EOS/`max_tokens` mid-call) — only fully-closed
  `<param>` pairs are recovered; a partial trailing value is never salvaged
- An unrecognized function name or an undeclared parameter is passed through
  rather than silently dropped or guessed at — schema validation stays
  centralized downstream (`toolCallConformsToSchema` et al.), unchanged for
  every dialect, matching the existing behavior for tag-format calls from
  every other model family

Not supported (not part of the required contract, and not part of
MiniCPM5's own documented format): a self-closing `<function ... />` form,
or an outer `<tool_calls>`-style wrapper around MiniCPM5 calls.

## Test plan

- `zig build test` — full suite: baseline `935 pass, 78 skip, 1 fail` →
  `957 pass, 78 skip, 1 fail` (+22 tests; the one failure is a pre-existing,
  unrelated GPU-kernel-precision test that reproduces identically on `main`)
- Red→green TDD throughout: new tests were added first against the
  unimplemented behavior (confirmed failing for the right reason), then the
  implementation landed, then all green
- Hermetic regression corpus entries in `src/format_corpus_test.zig` covering
  single/multiple args, CDATA (including the boundary case above), sequential
  calls, malformed XML, unknown function/parameter, missing required
  parameter, duplicate parameter, prose before/after a call, and no-call
  prose — plus confirmation the pre-existing bare-Hermes `<function=...>`
  format is unaffected
- Streaming-gate unit tests for partial-prefix growth (`<fun` → `<function`),
  an incomplete `<param>`, and an incomplete closing tag
- A deterministic test against the real `parseToolCallsForRequest` chokepoint
  every HTTP dispatch site calls, and a matching test for the Anthropic
  streaming handler, both exercised independently of any live model
- Real end-to-end verification against `mlx-community/MiniCPM5-1B-OptiQ-4bit`
  (non-streaming and streaming, both APIs): clean structured tool calls, zero
  raw XML in visible content, plain chat unaffected

## Scope note

This PR is a focused parser/protocol compatibility fix: it makes mlx-serve
correctly recognize and execute MiniCPM5's tool-call format instead of
leaking it as text or dropping it. It does not address, and does not claim to
address, MiniCPM5's behavior as a coding-agent model once a tool result is
fed back into a multi-turn conversation — that is a model-capability
question, orthogonal to whether the server parses its output correctly, and
is being tracked separately. Parser correctness and model-level agentic
reliability are independent concerns; this PR only speaks to the former.

### DIFF
--- a/src/chat.zig
+++ b/src/chat.zig
@@ -1306,6 +1306,9 @@ pub fn endsWithPartialThinkOpen(buf: []const u8) bool {
 ///     accepted families: `<tool>`, `<tool …>`, `<tool_call…>`, `<tool_calls…>`,
 ///     `<tool_request…>`, `<tool_requests…>` (mirrors `parseToolCalls`).
 ///   * `buf` contains the Gemma 4 `<|tool_call` substring.
+///   * `buf` contains `<function` followed by whitespace (MiniCPM5's
+///     `<function name="…">`, and incidentally the equals-sign Hermes
+///     function-tag form too).
 ///   * `buf[0] == '{'` and `buf` contains `"name"` (raw JSON tool-call shape).
 ///   * `buf` ends with a partial prefix that could grow into one of the above
 ///     in the next token (`<`, `<t`, `<to`, `<too`, `<tool`, or any prefix
@@ -1334,6 +1337,25 @@ pub fn streamShouldBufferForTools(buf: []const u8) bool {
         scan = after;
     }
 
+    // `<function…` family: MiniCPM5's `<function name="…">` (space then a
+    // quoted attribute) — walk every `<function` occurrence, accept the
+    // first one immediately followed by whitespace. This also naturally
+    // covers the pre-existing equals-sign Hermes function-tag format
+    // (`<function=NAME>`, see parseHermesToolCall) which shares the same
+    // `<function` marker but was never covered by this gate before — a
+    // strict improvement, not a behavior change, since that format was
+    // simply never held during streaming until now. A bare `<function>`
+    // with no attribute at all (not part of either format) is left alone,
+    // matching how unrelated `<toolkit>`/`<toolbar>` prose flows through
+    // the `<tool` family above.
+    scan = 0;
+    while (std.mem.indexOf(u8, buf[scan..], "<function")) |rel| {
+        const after = scan + rel + "<function".len;
+        if (after >= buf.len) return true; // truncated mid-prefix → in progress
+        if (std.ascii.isWhitespace(buf[after])) return true;
+        scan = after;
+    }
+
     // Trailing partial prefixes — the next streamed token could complete any
     // of these into a real tool open. Order doesn't matter; first endsWith
     // hit wins. Listed shortest-first for legibility.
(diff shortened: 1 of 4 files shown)


ddalcu:
Thanks for this, some conflicts to solve, but besides that there are a few fixes needed in addition:

* The tail-prefix list is "<f","<fu","<fun","<func","<functi","<functio","<function", <funct is missing. A token boundary landing exactly there returns false, flushes <funct to the client, and the rest of the tag leaks. That is precisely the class the PR fixes. The unit test enumerates the same list, so it's blind to the hole.
* Gate is looser than its neighbour: the <tool branch validates the family, the new <function branch accepts any <function + whitespace. <function foo> in prose buffers the whole rest of a tool-enabled response. Requiring name= would match the existing rigor.
* Missing the docs, no docs/gotchas/tool-calling.md story

PR #102, opened 2026-07-24
Fix Qwen vision preprocessing, prefill, and native MTP positioning

CerebralCoding:
## Summary

This fixes incorrect and repetitive Qwen vision responses, restores reference-compatible image preprocessing, and safely enables native MTP for multimodal requests.

The affected paths had three independent issues:

- Image preprocessing used fixed pixel limits and nearest-neighbor resizing instead of the checkpoint's processor settings and Pillow-compatible bicubic resampling.
- SSM checkpoint alignment split vision prefill into multiple forward passes, causing each pass to reuse the beginning of the vision embedding tensor.
- Native MTP was disabled whenever vision embeddings were present because the sidecar KV cache did not carry Qwen's three-axis M-RoPE positions.

### Image preprocessing

The preprocessing path ignored `min_pixels` and `max_pixels` from the checkpoint processor configuration, used hardcoded Qwen defaults, and resized decoded images with nearest-neighbor sampling. This changed both the number and contents of the image tokens compared with the official processor.

### Vision prefill

Vision requests already selected the full prefix length as their default prefill chunk. However, SSM checkpointing was enabled independently whenever the model exposed SSM state and a checkpoint stride was configured. `nextChunkEnd` therefore continued to force checkpoint-aligned boundaries, typically every 256 tokens.

Vision embeddings are scattered from the beginning of their source tensor within each forward pass. Splitting a vision prefix restarted that indexing for every chunk, so later chunks repeatedly consumed the first image features. This produced duplicated visual concepts and repeated text in the generated response.

### Native MTP and M-RoPE

The scheduler avoided incorrect speculative decoding by disabling native MTP for every request containing vision embeddings. Simply removing that guard would not be correct: the MTP head maintains its own prompt-history KV cache and previously applied ordinary scalar, cache-relative RoPE to it. Qwen image tokens instead require the same explicit temporal, height, and width positions used by the trunk.

The MTP cache can also contain only a suffix of a restored or long prompt, so its relative cache positions cannot be treated as absolute prompt positions.

This change maps MTP cache offsets back to absolute prompt positions. Positions inside the prompt use the explicit three-axis M-RoPE table; generated text beyond the table collapses to Qwen's scalar `absolute_position + mrope_delta` convention. Multi-token verification continues to use the existing generated-text path.

## Changes

- Parse Qwen image-area bounds from `processor_config.json`, falling back to `preprocessor_config.json`.
- Support both current and legacy processor configuration layouts.
- Thread the resolved pixel bounds through request image preprocessing.
- Replace nearest-neighbor resizing with Pillow-compatible bicubic resampling, including scale-aware anti-aliasing and fixed-point rounding for downscaling.
- Disable SSM checkpoint-aligned prefill splitting when vision embeddings are present.
- Preserve the existing SSM checkpoint behavior for text-only requests.
- Add an M-RoPE position context that maps suffix-only cache offsets to absolute prompt positions.
- Apply explicit three-axis M-RoPE to the native MTP head's prompt-history KV cache.
- Carry the position mapping through MTP prefill, draft steps, and merged history forwards.
- Enable native MTP for multimodal requests while leaving the unrelated external-drafter path guarded.
- Add unit and reference-parity coverage for processor bounds, bicubic resizing, substantial downscaling, vision-prefill chunking, suffix position mapping, and the prompt-to-generated M-RoPE boundary.

Image-bearing prompts remain excluded from prefix reuse because identical placeholder token IDs do not imply identical image content. Skipping intermediate SSM checkpoints for vision prefill therefore does not remove a valid cache-hit path.

## Validation

Full test suite:

```text
Build Summary: 6/6 steps succeeded
965 tests passed
79 skipped
0 failures
```

The `ReleaseFast` build also succeeds.

Reference preprocessing parity:

```text
730x487 -> 736x480: max_abs=0.000000, mean_abs=0.000000
730x487 -> 288x192: max_abs=0.000000, mean_abs=0.000000
22/22 tests passed for each case
```

The second case intentionally forces a substantial downscale to exercise Pillow's scale-aware anti-aliasing path.

End-to-end testing changed the affected image response from repeated fragments to a complete and correct reading of the visible content.

### Vision preprocessing and prefill

On an M5 Max with the same model, image, and prompt:

| Version | Prefill | Decode |
| --- | ---: | ---: |
| Before the preprocessing/prefill fix | ~500–506 tok/s | ~31.9 tok/s |
| Fixed preprocessing and single-pass prefill | 562.5 tok/s | 31.3 tok/s |

The single-pass vision prefill improves prefill throughput by approximately 11–12%, while ordinary autoregressive decode performance remains effectively unchanged.

### Native MTP and GPU NAX

A separate controlled run used the Qwen3.6 27B 4-bit MTP model, the same image and prompt, temperature 0, MTP depth 8, KV quantization off, and prefix caching disabled:

| Configuration | Prefill | Decode | Request time |
| --- | ---: | ---: | ---: |
| Autoregressive, MTP off | 202.23 tok/s | 33.20 tok/s | 11.66 s |
| MTP depth 8, GPU NAX disabled | 202.47–206.78 tok/s | 87.88–88.08 tok/s | 5.60–5.65 s |
| MTP depth 8, GPU NAX enabled | 200.94–208.74 tok/s | 95.53–96.34 tok/s | 5.28–5.34 s |

This is approximately:

- **2.65x** decode throughput from native MTP without NAX;
- a further **9.5%** from the GPU NAX verification path; and
- **2.90x** decode throughput overall versus autoregressive generation.

The NAX-enabled and NAX-disabled MTP runs produced the same visible response, completion-token count, and output length. Repeated runs were stable.

The speculative statistics confirm that the multimodal request actually reaches the configured depth:

```text
mode=mtp
attempts=64
accepts=263
avg_per_round=4.11
per_draft_pct=91.3%
depth=8
runtime_disabled=false
```

The NAX comparison was made by restarting the same release binary with the verification-QMM NAX kill switch disabled. This isolates the GPU Neural Accelerator contribution from the larger speculative-decoding gain.


ddalcu:
man, your timing is impeccable  !  We need this for adding support to MageFlow :)   
Please fix conflicts and we can merge 👍

PR #105, opened 2026-07-24
Feature/mageflow

### LINKED ISSUE #103: Support Images edit api
To support image editing in tools like OpenWebui there is a separate open ai api compatible endpoint required (images/edit), [openai documentation here](https://developers.openai.com/api/reference/resources/images/methods/edit)

There already is the image generation api endpoint available, which seems to support editing by specifying a specific edit mode.

Do you think you could add support for this new endpoint and route it to the already existing image edit capabilities?

ddalcu:
Looking for testers, please let me know if mageflow works for you. 

Get DMG from the artifact on this page: https://github.com/ddalcu/mlx-serve/actions/runs/30134990740

Also Fixes #103


ivanfioravanti:
Great job! Gonna test this!

bravo-delta:
I have tested this with Mage-Flow Edit Turbo and I was able to generate and edit an image within the Image Generation modal of the application, which is great!

I also tested the new images/edits endpoint via Open WebUI, but here the calls fail with 400 Bad Request. Is there any way to log the incoming payload of the call via mlx-serve, to figure out whats going wrong?

Open WebUI fails at this line of code: [app/backend/open_webui/routers/images.py#L1024](https://github.com/open-webui/open-webui/blob/main/backend/open_webui/routers/images.py#L1024)

ddalcu:
thanks guys. Fixed.

image
image

PR #118, opened 2026-08-02
Multi-Lora Stacking

justinluque:
Feature update that allows the user to attach multiple loras. Also includes a fix that makes the lora loader architecture-aware which allows it to more reliably load lora files. As far as I can tell it works but I can't figure out why it fails tests. Co-authored by Claude.


ddalcu:
awesome work here, please let me know when you are ready to merge this... currently adding support for Minimax H3, which for sure will cause conflicts.

justinluque:
If you want to take a look I think this is good

ddalcu:
merged this into my upcoming PR, and adding you as a contributor to it, thanks !  Could really use your help testing once it's up, if you don't mind.

PR #122, opened 2026-08-03
Feature/minmax h3

ddalcu:
Creating PR, so it kicks off a build and makes it downloadable via Actions build.  DMG Should be signed, if you dont want to deal with compiling it.

Not ready for merging to main yet, needs a lot of testing, please report bugs in this PR.

P.S. Im looking for help on this project !   Both coding, and community building. (message & follow me on X  https://x.com/ddalcu)

## MiniMax-H3 on Apple Silicon, native and fast

This lands the full MiniMax-H3 (Hailuo 3.0) pipeline in mlx-serve: a 34B DiT that generates video and its stereo soundtrack together, in one pass, from one prompt. No Python, no ComfyUI, just POST /v1/video/generations. And it's fast: the same 1344x768, 124-frame, fully on-spec generation that took 2h19m now renders in 49 minutes. Same seed, same machine, 2.83x.

How it got fast

We profiled before touching anything. An in-run ablation ladder priced every component of a denoising step without breaking GPU pipelining, and the verdict was humbling: the DiT already runs at the compute roofline (~13.5 TFLOPS effective on both GEMMs and attention). No custom kernel was going to save us. So every win below comes from doing less work, not from doing work faster:

- Wide-M dequant GEMM. The quantized 8-bit path was paying the quantized_matmul tile tax at 9K-row widths. Dequantizing to bf16 and riding the steel GEMM cuts 13% off every step, with byte-identical frames. Default on.
- Velocity caching. Adjacent denoising steps at high sigma are nearly identical, so when the accumulated input change is small we reuse the previous velocity instead of running 50 transformer blocks. Roughly half the steps become free.
- Attention broadcast. Attention outputs drift slowly mid-schedule, and attention is ~70% of a 768p step. We cache each block's attention output and recompute it every other step, re-applying only the timestep gating in between.
- AdaLN precompute. 39% of the model (13B parameters) exists only to turn a timestep into modulation vectors. We now evaluate the entire schedule's modulations up front and free those weights before sampling starts. DiT residency drops from 33 GB to 20 GB.

The fast recipe is default on. If you want every step fully computed for a final render, send "fast": false or flip "Max quality (slower)" in the app. Both paths are pinned by tests that count engagement, not vibes.

### Runs on small Macs now

A new 4-bit pack (MiniMax-H3-FL2VA-MLX-Serve-4bit) ships as a second preset: 40 GB download instead of 69, an 11 GB resident DiT, and the same generation speed since the workload is compute-bound. Combined with the staged-residency preflight (the text encoder is loaded, run, and freed before the DiT ever loads, and the server now bills exactly that), H3 fits comfortably on a 32 GB machine.

Also in here

- First/last-frame conditioning (fl2va): attach an image and frame 0 reproduces it. The new VAE encoder matches the reference implementation at cosine 0.999998, verified against ComfyUI's own code actually executing.
- Spatial tiling and temporal chunking parity for the VAE, seam-checked numerically.
- A full integration suite (tests/test_minimax_h3.sh): capabilities, named 400s, SSE progress, frame-ladder snapping, fl2va adherence, fast-recipe on/off engagement.
- The profiling harness stays in the tree (MINIMAX_H3_ABLATE=ladder), plus a training-free sparse attention implementation behind an env flag for future tuning.

Every number above was measured with paired same-seed runs, and everything quality-touching shipped only after side-by-side clips passed review.

HF Model Links:
https://huggingface.co/ddalcu/MiniMax-H3-FL2VA-MLX-Serve-4bit
https://huggingface.co/ddalcu/MiniMax-H3-FL2VA-MLX-Serve-8bit


Ricardo9527:
Hi David — thanks for the excellent H3 work. Your M4 Max profiling results, wide-M q8 → BF16 Steel path, and PAB implementation have been very helpful for our own native MLX Ref2VA experiments.

On our M5 Max, we have been testing a true W8A8 projection path:

- grouped-256 Hadamard rotation folded into per-output-channel int8 weights;
- runtime Hadamard rotation with per-token symmetric activation quantization;
- Metal 4 `mpp::tensor_ops::matmul2d` for int8×int8→int32, followed by a two-scale epilogue.

Our current `mx.fast.metal_kernel` implementation measures about 1.344x over `mx.matmul` across QKV, attention-out, FC1 and FC2 at roughly 17K rows. This translates to around 1.22x per complete DiT step.

A two-step full-stack comparison against BF16 measured video latent cosine 0.999882 and relative L2 0.0156. We did not see a visible quality loss in the initial decoded A/B, although a broader multi-prompt quality suite is still pending.

A few questions:

1. Since your published H3 benchmarks were collected on an M4 Max, have you had access to M5 hardware to test the Metal 4 integer tensor path?
2. Would an optional M5-specific W8A8 backend be useful for mlx-serve, while retaining your current q8 → BF16 Steel implementation as the M4/general fallback?
3. At 768p, the full per-block PAB cache becomes quite large. Have you considered partial-layer PAB or compressed attention caches for 32 GB machines?

Happy to exchange implementation details and benchmark results if this direction is useful.

skymars13:
Since the PR notes say this needs testing: I've been running essentially the same staged-residency architecture for a few days on a 64 GB M5 Max (built on the pipenetwork Python port), so here are reference numbers from that setup. All same-seed, back-to-back:

- Everything loaded up front: ~73 GB peak. Staged (encode → release the encoder → load the DiT): 52 GB during encode, 30 GB during denoise.
- With the Qwen3-VL-32B encoder requantized to 8-bit (group 64, layers with in-dim % 64 ≠ 0 kept bf16): 66.7 → 27.5 GB on disk, encode+release in ~22 s.
- 10 s @ 544×960: 31.3 min at stock 15 steps → 16.6 min with sigma-progress step skipping → 12.7 min with a turbo LoRA merged into the quant at 8 steps.

Two things that might save you testing time:

1. A small on-disk conditioning cache stacks really well with staged residency. The embeddings only depend on (prompt, image bytes, encoder dir), not on seed/steps/resolution, so seed rerolls skip the encoder stage completely.
2. Skip caps matter more than the threshold. At 0.15, three consecutive skips gave visible ghosting; ≤2 at 15 steps / ≤1 at 8 steps stayed clean in frame checks. An output-change-rate signal correctly refuses to skip at turbo 8 steps — no margin left there.

Notes and scripts: https://github.com/skymars13/minimax-h3-mlx-64gb

Staged residency is the right call — on 64 GB machines it's the difference between "doesn't fit" and comfortable.

Issue #126, opened 2026-08-05
[Bug] Eviction gate overestimates staged-residency peak for MiniMax-H3, causing false OOM


funk80rus:
## Summary

`ddalcu/MiniMax-H3-FL2VA-MLX-Serve-4bit` is unloadable on a 48 GB machine. Every
attempt returns:

```
HTTP 503
{"error":{"message":"Not enough memory to load model; retry after current requests complete","type":"out_of_memory","param":null,"code":503}}
```

This happens on a **completely idle server with zero models loaded and RSS of 21 MB**,
so the "retry after current requests complete" advice is unactionable — there is
nothing to wait for.

The refusal is a false positive. The model's real staged peak is **22.83 GiB**, and
`gen.zig` already computes that number correctly. But the eviction gate in
`scheduler.zig` runs *first* and bills the **sum of every safetensors in the
directory + 10%** = **41.30 GiB**, which exceeds the default
`max_resident_mem` of 30.0 GiB. The load is refused before
`doLoadGenOnInferenceThread` — and therefore before the correct estimator — ever runs.

Raising `--max-resident-mem` makes the model load and generate video normally.

## Environment

- mlx-serve 26.8.2 (MLX 0.32.0), also reproduced on 26.8.1
- MLX Core.app 26.8.2
- MacBook Pro, Apple M5 Pro, 48 GB unified memory (`hw.model = Mac17,8`)
- macOS 26.x (Darwin 25.6.0)
- Model: `ddalcu/MiniMax-H3-FL2VA-MLX-Serve-4bit` (4-bit, `model_type: minimax_h3`)
- `iogpu.wired_limit_mb = 0` (stock); server reports `[wired] mode=max limit=38338 MB`

## Reproduction

```bash
mlx-serve --serve --port 11234 --model-dir ~/.mlx-serve/models

curl -X POST http://127.0.0.1:11234/v1/load-model \
  -H 'Content-Type: application/json' \
  -d '{"model":"ddalcu/MiniMax-H3-FL2VA-MLX-Serve-4bit"}'
```

Server log at the point of refusal:

```
[discovery] skip nsfw_image_detection: unsupported model_type 'vit'
Discovered 1 model(s) under /Users/<user>/.mlx-serve/models:
  - ddalcu/MiniMax-H3-FL2VA-MLX-Serve-4bit (37.5 GB)
mlx-serve 26.8.2 (MLX 0.32.0)
[...]

PR #130, opened 2026-08-07
v26.8.3 — MiniMax-H3 references and Turbo, stacked LoRAs, model folders  -- Need help testing if anyone is available.

ddalcu:
### MiniMax-H3: references, Turbo, longer clips

- Attach pictures, clips or audio and the video is built around them. Refer to them in the prompt by number: `<Picture 1>`, `<Video 1>`, `<Audio 1>`. This needs the REF2VA build, so the app only offers it when that build is loaded and the server refuses references on a build that would silently ignore them.
- **Turbo** renders in 4 steps instead of 30, about twice as fast end to end, with slightly softer detail. It ships with the model now, and packs downloaded before it existed fetch it once when you tick the box. Thanks to [larryvrh](https://huggingface.co/larryvrh/MiniMax-H3-Turbo-Lora) for the adapter.
- Clips reach 15 seconds. Under the Generate button there is a live time estimate that updates as you change size, length and steps, and the memory warning now uses H3's own numbers instead of LTX's.
- Video models no longer refuse to load on a 48 GB Mac (#126). The memory gate billed the sum of every file in the folder, 37.55 GB, where the real peak is 22.83 GB. Thanks to funk80rus for the report and the diagnosis.
- Over the API, long clips can be generated as chained windows (`chain_windows`), each continuing from the last frame of the one before.

### LoRAs

- Attach several style adapters at once, on any image or video model. Their effects add up, so the order does not matter. Thanks to Justin ([@justinluque](https://github.com/justinluque)) for the multi-adapter work (#118).
- Adapters now run at the strength their own file declares. Anything exported by PEFT or diffusers was running up to 8x too strong, which renders as static rather than a stronger style. If you dialled one down to compensate, put it back up.
- MiniMax-H3 takes them too, stacked on top of Turbo.

### Model folders: pick where downloads go

- **Settings ▸ Model Folders ▸ Default folder** chooses where new downloads land. Everything already downloaded keeps working, because the old folder stays in the scan list.
- Custom folders are actually served now. `--model-dir` is repeatable, so every folder you list shows up in `/v1/models` instead of only in the app's own picker. If the same model sits in two folders the download folder wins, and a folder that is not reachable is skipped with a warning instead of stopping the server.

### Logprobs. Fixed.

- Three things were wrong at once, on every model: the numbers moved when you changed the temperature instead of being the model's own, ties handed back an arbitrary token, and the whole list was off by one so each token came back with the *next* token's alternatives.
- None of this changed the text any model produced. It matters if you score outputs, route on confidence, or build evals.
- `/v1/completions` ignored its `logprobs` field entirely. It works now, in the shape the OpenAI legacy API defines.

### Also new

- **Pick your quant from a multi-variant repo.** Publishers who ship every quantization in one repo as subfolders used to read as a single model; each variant now shows its own size and downloads, loads and deletes independently, like GGUF repos already did.
- **The seed is a text field you can paste into** in the image and video panes, with a dice button beside it that gives you a concrete number to read off and paste back.
- **The app says why a reply stopped.** A cut caused by the model going in circles is labeled as one instead of claiming you hit the output limit. On the API that is `finish_details: {"type": "repetition_loop"}` beside `finish_reason`.
- **A quant playground on the website**: pick a Mac, a model, a quantization and a context length, and see the speed you can expect and whether it fits in RAM.
- LICENSE, NOTICE and the Apache-2.0 text now ship with the CLI tarball, the app bundle and the Homebrew install, crediting the kernels and libraries mlx-serve builds on.

### Fixes

- Reserved tokens such as `<|fim_hole|>` can no longer be sampled into a reply. The list is built per model from its own tokenizer and chat template, so thinking tags and tool markers are untouched.
- Tool calls no longer vanish when an argument mentions `</think>`. Coding agents write files about prompts, so this was reachable on every thinking model.
- The transcript follows a streaming reply and stops the moment you scroll up, on every input rather than only the mouse wheel.
- The MiniMax-H3 fast recipe silently did nothing on runs of about 8 steps or fewer while still reporting that it was on.
- Added a repetition guard for loops that reword themselves; the existing guards only caught exact repeats.
- Streaming a long thinking block no longer costs CPU that grows with the length of the thought.
- `/detokenize` returned bodies that no JSON parser would accept if any token contained a control byte.


AideYu:
Test report: 24GB Mac mini M4 Pro (from a community tester)

Environment: Mac mini M4 Pro 24GB, macOS 26.6.1, MiniMax-H3-FL2VA-MLX-Serve-4bit, launch with --skip-mem-preflight. Build tested: 26.8.0-pr130.da9d23a (Actions artifact).

Test 1 — Smoke (512x288 / 56 frames / 4 steps / turbo:true / first+last keyframes):
- HTTP 200 in 127s (sampling ~85s + decode ~27s)
- Log confirms Turbo engaged: [lora] turbo_lora.safetensors: scale 1.0000, lora 1/1: 259/259 modules, scale 1.00 (turbo), speed recipe: ... turbo=true
- Valid frame content (silver car, racetrack, golden hour) — no artifacts

Test 2 — Full spec (1344x768 / 124 frames / 4 steps / turbo:true / original 1280x720 keyframes):
- HTTP 200 in 2812s (~47 min), 512.9 MB response
- 124 frames @ 24fps, native audio (32kHz stereo) included
- Valid content verified on first/last frames (car, mountains, sunset) — no visible artifacts
- Memory stable: DiT resident ~11.4 GB, swap well below the non-turbo stress peak (22.5 GB)
- Observation: at 1344x768 the per-step time (~10.4 min) is not faster than the non-turbo 4-step path (~10.1 min) — the LoRA compute overhead eats the speedup at this resolution. But quality is clearly better (non-turbo 4-step output is noticeably soft/blurry; turbo output looks like a much higher step count). So on 24GB hardware turbo is a free quality upgrade at same wall time for large canvases, and a speed+quality win on small canvases.

Notes:
- Client-side timeout must be generous (>45 min) for 1344x768; a disconnected client leaves an orphan job that holds the server queue until finished.
- --skip-mem-preflight still required on 24GB (preflight estimates ~22.8 GB peak which is above free memory with OS overhead).

Thanks for the great work — turbo is working as described on Apple Silicon.

funk80rus:
Thank you for the incredible work! The references and Turbo are what I was waiting.

Since my last report I've switched to the 8-bit checkpoints. Good news: they also fit on a 48 GB Mac and the results are excellent — 8-bit sampling came out no slower than 4-bit, it just costs more RAM. So the memory ceiling on this hardware is higher than the current estimator assumes. Here are my test results:

Tested this build on a 48 GB M5 Pro Mac (macOS 26.x, `26.8.0-pr130.bcb94d8`, MLX 0.32.0). I'm the reporter of the sum-of-safetensors eviction-gate issue — **that part is fixed, confirmed.** Two things still block the 8-bit H3 checkpoints, and one of them is a GUI-only blocker.

## Confirmed fixed

The gate now bills the staged-residency peak instead of the sum of every safetensors in the dir, and it names its numbers. On `ddalcu/MiniMax-H3-FL2VA-MLX-Serve-8bit`:

```
Refusing to load ddalcu/MiniMax-H3-FL2VA-MLX-Serve-8bit: needs ~42.87 GB but
--max-resident-mem is 29.95 GB (0.00 GB already resident across 0 model(s),
none evictable). Raise or disable the cap with --max-resident-mem <size>|0.
```

42.87 = `(max(te, dit) + video_vae + audio_vae + turbo_lora) × 1.1` = the `gen.zig` estimator. The old path would have billed `sum × 1.1` = **71.8 GB** on the same files. `bytes_resident` dropped from 65.25 GB to 38.97 GB too. The new refusal line is exactly what was missing — thank you.

## Issue 1: MLX Core.app can't reach the workaround

`--max-resident-mem` is still not in the app's server argument list, and there's no extra-args passthrough:

```
$ strings "MLX Core.app/Contents/MacOS/MLXCore" | grep -E "^--(max-resident-mem|skip-mem-preflight)$"
--skip-mem-preflight
```

So from the GUI the 8-bit checkpoints are unloadable at any setting. `--skip-mem-preflight` (the one memory flag the app does pass) doesn't help, because the registry gate runs before the preflight. Everything below is only reachable from a hand-launched server.

## Issue 2: the staged-peak formula still overcounts, by a lot

The estimate is `max(te, dit) + video_vae + audio_vae + lora`, but the backend never holds those concurrently — its own log narrates the unload chain. Full sequence on this checkpoint:

```
[minimax-h3] text encoded, encoder released
[minimax-h3] adaln precomputed for N timesteps; 13B modulation weights released
[minimax-h3] dit resident: 19.99 GB mlx-active
[minimax-h3] sampling done, DiT released
[minimax-h3] video decoded (load+decode)
```

Two separate overcounts:

1. **The VAEs are billed against the DiT.** They only load after the DiT is released, so the real shape is `max(te, dit + lora, video_vae + audio_vae)`, not `max(te, dit) + vaes + lora`. On this checkpoint that's `max(26.28, 33.55, 5.41)` = 33.55 GiB, down from 38.97 GiB.
2. **`transformer.safetensors` is a bad proxy for DiT residency.** The 13B modulation weights are released right after AdaLN precompute, so a 32.83 GiB file settles at **~20 GiB** resident. The ratio is stable across quants (0.62 on 4-bit, 0.61 on 8-bit) and it's the dominant error term.

Measured against `footprint --sample 2` (`phys_footprint`, not `ps -o rss`, which under-reports Metal unified-memory residency here), 768×448 / 124f / 15 steps:

| | 4-bit | 8-bit |
|---|---|---|
| `transformer.safetensors` | 17.41 GiB | 32.83 GiB |
| DiT resident (self-reported `mlx-active`) | 10.84 GiB | 20.19 GiB |
| **process peak (`phys_footprint`)** | **17 GB** | **26 GB** |
| gate estimate (staged peak × 1.1) | 25.11 GiB | 42.87 GiB |

So the estimator lands at roughly 1.7× the measured peak on the 8-bit checkpoint. That's what makes it unloadable: 42.87 vs the auto cap of 29.95 GiB (80% of this machine's 38338 MB wired limit), when the process actually peaks at 26 GB — comfortably inside the cap it's being refused by.

## The 8-bit checkpoints do run on 48 GB

With `--max-resident-mem 0 --skip-mem-preflight` (both flags needed — gate first, then preflight refuses with `media peak ~38.97 GB, available 24.62 GB`):

- 256×256 / 22f / 4 steps / fast → HTTP 200 in **10.9 s**, DiT resident 19.99 GB
- 768×448 / 124f / 15 steps / fast=false → HTTP 200 in **564.7 s**, peak 26 GB, no swap growth
- Ref2VA 8-bit, 448×768 / 124f / 15 steps → HTTP 200 in **583.4 s**, peak 27 GB

Side note from the same runs: 8-bit sampling was 1.4% *faster* than 4-bit (532.0 s vs 539.7 s — inside noise) at +53% peak memory, so the dq-gemm path looks compute-bound rather than bandwidth-bound at this resolution.

ddalcu:
[funk80rus](https://github.com/funk80rus) ... just when I was about to hit the release button...  haha, thanks, I will fix it here shortly and push it.

NavSha:
**MiniMax-H3 never loads: `modelKindFromType()` is missing a `minimax_h3` branch, so the pack falls through to the generic LLM loader.**

Repro (64 GB M-series, macOS 26.x, pack `ddalcu/MiniMax-H3-FL2VA-MLX-Serve-4bit` @ `9c92735`, builds `v26.8.2` / `v26.8.3` / `26.8.0-pr130.da9d23a`, and the MLX Core app):

```
POST /v1/video/generations  →  MISSING WEIGHT: model.norm.weight   (exit 133)
```

Note it's the bare `MISSING WEIGHT:`, not your `[minimax-h3] MISSING WEIGHT:` — a different loader is running.

Cause, in `src/model_discovery.zig`: `minimax_h3` is in `isSupportedModelType` (line 63) but has no case in `modelKindFromType` (line 396), so it reaches `if (isSupportedModelType(model_type)) return .chat;` and loads as a chat model. That loader wants `model.norm.weight`, then `lm_head.weight` — LLM tensors the pack correctly omits (`text_encoder.layers = 50`, no LM head, per `MODIFICATIONS.md`). `VideoEngine.load` has the right `minimax_h3` branch; it's just never reached.

```diff
      if (std.mem.eql(u8, model_type, "AudioVideo")) return .video;
+     if (std.mem.eql(u8, model_type, "minimax_h3")) return .video;
      if (std.mem.startsWith(u8, model_type, "hunyuan3d")) return .mesh;
```

Worth checking whether `ltx` / `mageflow` types have the same gap.

Ruled out: all three builds; `--model` vs `serve --model-dir`; text-only vs first/last keyframe requests; canonical vs non-canonical dir name; incomplete download (all 15 files verified); memory gates (`--skip-mem-preflight --max-resident-mem 0` in use — without them the 41.47 GB auto cap refuses a 42.10 GB estimate, consistent with @funk80rus's ~1.7× overcount). Via the app it fails as `400 "Target model does not support this media modality"` — same defect at dispatch instead of load.

Two notes:

- The startup banner is a red herring: it lists only text endpoints, but `/v1/video/generations` is live (answers 400/503, not 404). I initially misread its absence as the route being unregistered.
- @AideYu's 24 GB success on `pr130.da9d23a` is hard to square with this — if the branch is genuinely absent, that path shouldn't work anywhere. Could you post `shasum -a 256` of the binary you ran, and confirm whether your pack predates the `b454555` Turbo-LoRA swap?

I'm building the patched binary against this pack now and will report back with the result (happy to open a PR if it works).

NavSha:
**Confirmed: the one-line patch fixes it.** Built from source with the branch added and H3 generates video+audio on my 64 GB M-series machine.

```diff
      if (std.mem.eql(u8, model_type, "AudioVideo")) return .video;
+     if (std.mem.eql(u8, model_type, "minimax_h3")) return .video;
      if (std.mem.startsWith(u8, model_type, "hunyuan3d")) return .mesh;
```

Before: registers as chat, dies on `MISSING WEIGHT: model.norm.weight`.
After: `/v1/models` reports `MiniMax-H3-FL2VA-MLX-Serve-4bit | capabilities: ['video']` and the right loader runs —

```
[minimax-h3] speed recipe: step-cache 0.000, attn-broadcast k=0 (fast=null, turbo=true)
[minimax-h3] step 1/4 sigma 1.0000 (5035 ms)
[minimax-h3] step 4/4 sigma 0.8000 (3862 ms)
[minimax-h3] sampling done in 16242 ms, DiT released
[minimax-h3] video decoded in 5546 ms (load+decode)
[minimax-h3] audio decoded: 74400 samples/ch at 32000 Hz (1370 ms)
[video] -> 56f 288x512 (24772608 rgb bytes)
```

512×288 / 56f / 4-step turbo / first+last keyframes: **~23 s end to end**, HTTP 200, valid frames and non-silent 32 kHz stereo (mean −28.5 dBFS). Still needs `--skip-mem-preflight --max-resident-mem 0` here (auto cap 41.47 GB vs a 42.10 GB estimate — the overcount @funk80rus described).

Happy to open a PR with just that line if useful. Two side notes from building `v26.8.3` on a machine whose Homebrew lives at `/usr/local` (Intel prefix, arm64 hardware), in case you want the build to be more forgiving:

- `build.zig` hardcodes `/opt/homebrew/{include,lib}` for webp, and `linkSystemLibrary("webp")` additionally picks up `/usr/local/opt/webp` — on this setup that's an x86_64 dylib, so the link fails with `invalid cpu architecture: x86_64`. A `-Dwebp-prefix=<dir>` option (like the existing `-Dios-include`) would have made this a one-flag fix instead of patching four call sites.
- `zig-out/bin/mlx-serve` has relative rpaths (`lib/mlx/lib`, `lib/llama/lib`), so it only runs from the repo root. Worth a line in CONTRIBUTING, or `@executable_path`-relative rpaths.

AideYu:
Thanks for the deep dive, NavSha. Here's my data for comparison:

- Binary used for my 24GB test report (Actions artifact `26.8.0-pr130.da9d23a`, downloaded 2026-08-07): `sha256 183f000b2526c05d90da4cd2c7b0d2d5c424fd31117751cb304eb269402d5787`
- Pack: `ddalcu/MiniMax-H3-FL2VA-MLX-Serve-4bit` — config has `model_type: minimax_h3`, turbo LoRA = ema_ckpt850, i.e. **not** predating `b454555`
- Current binary (v26.8.3, from release assets, updated this morning): `sha256 e81285311de68b308fafefe4ed6f7a35639ba9662ee03194a6f1b30cf3903316` — generated a full 1344×768 / 192f / 4-step turbo clip successfully today (~96 min wall time, HTTP 200, valid frames + audio)
- Launch path on both: `--model <pack dir>` + `--skip-mem-preflight` (24GB Mac mini M4 Pro, macOS 26.6.1)

Both binaries load and run H3 video fine here, so I can't reproduce the `MISSING WEIGHT: model.norm.weight` failure on this path — even though my pack is the current one with the `minimax_h3` model_type. If your repro is deterministic, maybe the difference is in exact launch args or how the pack directory is resolved; happy to run anything specific and report back.

PR #139, opened 2026-08-08
feat: add drag and drop image input to media gen

justinluque:


app/Sources/MLXServe/Views/VideoGenView.swift
```diff
@@ -1049,6 +1068,36 @@ struct VideoGenView: View {
...
+        return true
+    }
+
+    /// Same extension allow-list as ImageGenView's drop handling.
+    private static func isImageFile(_ url: URL) -> Bool {
+        let ext = url.pathExtension.lowercased()
+        return ["png", "jpg", "jpeg", "heic", "heif", "webp", "tiff", "tif", "gif", "bmp"].contains(ext)
+    }
+
+    private func numberField(_ label: String, value: Binding<Int>, step: Int) -> some View {
```
ddalcu: what does this function do ? 

https://www.youtube.com/shorts/GPIwgAUwXYw

justinluque: Looks like a leftover piece of code that went unused that was accidentally left in I went ahead and removed it

ddalcu:
Hi @justinluque  just wanted to let you know, not forgetting about you and your PR's
But... we have a BIG PR about to get merged, from Aaron C, and after talking with him a bit, we might be getting rid of the media gen screens in the future and integrating it more inline with chat, but we still need to think about it a bit.

I want to keep your PR's open because we might merge them one way or another !  Or at least take some ideas and merge them into the new UI.

justinluque:
Ready for merge, your choice. I've fixed any conflicts with Aaron's commit.

ddalcu:
- A comment states the opposite of what the code does. ImageGenView.swift:271  "so a full pane bounces the file instead of swallowing it". With source set and refs at 3, limit is 1 + 0 = 1, so MediaDropLoader.load returns true (drop animates in) and ImageDropPlacement.place no-ops. It swallows. Confirmed by running the pure function. This repo files a wrong contract comment as a bug, so either make the limit 0 there , which does bounce  or fix the sentence.

- Variation mode keeps the LAST file of a multi-file drop. limit: 1 + max(0, maxRefImages - refImageURLs.count) is 4 on a model with no reference edit (refs are always empty there), so place replaces the source four times: dropping 4 files keeps c.png and silently discards the rest. Video and 3D take urls.first. Limit should be 1 when !effectiveEditMode.
- No tests, on three pure enums extracted with the comment "so the routing can be read (and tested) on its own". H3RefDrop.route spends per-type + combined caps correctly and MediaDrop.accepted does preserve drop order across holes , but that's exactly what H3PlanningTests/TruncationNoticeTests pin for their neighbors. ~30 lines.

Minor:

- Any file URL is accepted: type filtering happens after the async resolve, so a dropped .txt lights the dashed border and animates in, then nothing. ChatView's existing drop uses onDrop(of: [.image, .pdf, .audio]) and bounces properly, the shared modifier could put the kind's UTTypes in of: and get that for free.
- supportsReferences: in place is dead, effectiveEditMode already includes it.
- Video's first-frame slot kept the bordered button + hint text instead of MediaDropWell, so it looks different from Image/3D.

ddalcu:
looks good, thank you, merging !

PR #140, opened 2026-08-08
feat: add user provided benchmark/example prompts to media gen panes

justinluque:
image


ddalcu:
hmm, not to sure about this one. How useful is it for someone to generate the same video/image over and over  ?
For benchmarking, people can just use the API and create a script.

justinluque:
Fair enough

PR #146, opened 2026-08-08
Chatview QOL

justinluque:
- cmdN for new chat, cmdR to regenerate
- adjust typing cursor inset
- edit messages by double clicking
- bulk select and delete conversations


justinluque:
Merged the latest from Aaron here

ddalcu:
1. swift test fails: ChatWorkspaceTests.testEverySidebarRowSharesOneGutterAndOneInset. ChatView.swift:710 puts .listStyle(.sidebar) back on the conversations ScrollView. That guard exists for exactly this and the comment at :678 says why it came off. It does nothing on a ScrollView anyway, just drop the line.

2. proceedSend() lost applyScroll(.userSentMessage). When you pulled the config out into buildTurnConfig the call ended up inside editAndResend instead. So sending from halfway up the history no longer scrolls to your message. The comment explaining it moved with it, so the code reads like it's still there.

3. isNearBottom (ChatView.swift:1130) is never read. You set it in regenerate and editAndResend and nothing looks at it. Both of those wanted applyScroll().

4. The double click gesture at :2972 sits on every bubble with the onEdit != nil check inside the closure instead of on the modifier. Your own comment says highPriorityGesture is there to win over textSelection, so on assistant replies it wins and then does nothing. That kills double click to select a word on the replies, which is most of what anyone selects. Move the check onto the modifier.


Minor/Medium
- No tests. The comment at :580 says SidebarMultiSelect is pure so it can be driven from SidebarMultiSelectTests, and that file isn't in the PR. SidebarSelectionTests and SidebarSessionGroupsTests are right there as the template. Shift range across the Agents/Chats boundary, cmd clicking the last selected row, the nearest survivor pick, that's the stuff that breaks silently six months from now.
- deleteSessions is deleteSession's body in a loop and they already drifted: one falls back to chatSessions.first, the other to visibleChatSessions.first. Make deleteSession(id) call deleteSessions([id]).
- 3 new deprecation warnings from onChange(of:perform:) at 725/732/739. Everywhere else uses the two arg version.
- ⌫ deletes the whole selection with no confirmation, and with nothing selected it deletes the chat you're in. N conversations gone on one keypress with no undo makes me nervous.
- Small stuff: // MARK: - Chat Detail got deleted, there's a blank line between the emptyState doc comment and the declaration which detaches it, and the "Only the USER gets a bubble" comment landed mid branch where it doesn't describe the code under it.

justinluque:
Appreciate the review, I'd like to add a few more features I've thought of if that's alright before merge:

- enter to submit edited chat
- enter to work to confirm deletion menus for conversations
- escape to cancel LLM generation
- Holding ⌘ shows numbers for selecting a chat

justinluque:
https://youtu.be/g-bVEc8oZvk

justinluque:
Ready for merge but probably needs a thorough review which I'll make when I get a minute

PR #147, opened 2026-08-08
fix: loop-stop banner shown twice in plain chat

justinluque:
The server states an ending twice on any stream that asked for usage: the final chunk carries finish_reason + finish_details, and the stream_options.include_usage chunk repeats both beside the usage object (server.zig, the two sendSSEChunk calls
closinghandleStreamingGeneration). The app always requests include_usage, APIClient yielded .truncated per chunk, and updateLastMessage(content:) appends -- so plain chat put two identical "the model started repeating itself" banners under one reply. The agent loop was unaffected: it records the cause and shows it once at the turn's exit.

APIClient.TruncationGate latches the cut per stream, so a restated ending is dropped no matter which backend restates it; streamPlainResponse now records the cause and appends after the stream, mirroring the agent path, so the append site is structurally single rather than merely fed one event. Wire format unchanged -- repeating finish_reason deviates from OpenAI (choices: [] there) but is harmless to correct clients, and changing it would move behaviour other consumers read.

Before:
image
After:
image

### DIFF
--- a/app/Sources/MLXServe/Services/APIClient.swift
+++ b/app/Sources/MLXServe/Services/APIClient.swift
@@ -333,6 +333,34 @@ class APIClient {
     /// any other OpenAI-compatible backend) sends no such field and reads as
     /// `.maxTokens`, which is exactly the behaviour this replaced.
     ///
+    /// One cut, one event — no matter how many times the stream states it.
+    ///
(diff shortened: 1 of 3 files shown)


ddalcu:
i have this merged into my branch with a few tweaks, thank you !

Issue #135, opened 2026-08-08
LM Studio Style .json settings per model


IIIIIIIllllllllIIIII:
It would be really cool to be able to set system prompts and i dividual model settings such as temperature, context window, etc etc PER MODEL and not only have a one-fits-all solution... let the user define a folder with presets and select json with settings for esch model the user has... best wouöd be cross compatibiöity between lm studio settings and mlx serve settings so those can be in a shared folder i.e. with the downloaded models! 

Thanks so much!

ddalcu:
yeah... I was thinking about this also.. some things are server vars (like context), and most are chat completions api that can be changed at the chat time.

Have you tried the agents feature ? You can customize some things there. But yes... a model specific config is missing, I will think about it.

<img width="1149" height="1373" alt="Image" src="https://github.com/user-attachments/assets/645a7ebd-2a94-431d-9058-d43e62e44f1b" />

IIIIIIIllllllllIIIII:
I will have to try that.... chat_template.jinja is also something that should be customized/imported since when 3.6 needs a fix here for some models....  also find attached some examples of "essential" qwen json configs :)

this would be something really nice to have and would allow me to finally switch from lm studio for good :) 

[qwen3.6-40b-davidau-deckard-heretic-instruct.preset.json](https://github.com/user-attachments/files/30851588/qwen3.6-40b-davidau-deckard-heretic-instruct.preset.json)
[qwen3.6-40b-davidau-deckard-heretic-thinking-coding.preset.json](https://github.com/user-attachments/files/30851594/qwen3.6-40b-davidau-deckard-heretic-thinking-coding.preset.json)
[qwen3.6-40b-davidau-deckard-heretic-thinking-general.preset.json](https://github.com/user-attachments/files/30851593/qwen3.6-40b-davidau-deckard-heretic-thinking-general.preset.json)
[qwen3.6-27b-hauhaucs-instruct.preset.json](https://github.com/user-attachments/files/30851592/qwen3.6-27b-hauhaucs-instruct.preset.json)
[qwen3.6-27b-hauhaucs-thinking-coding.preset.json](https://github.com/user-attachments/files/30851595/qwen3.6-27b-hauhaucs-thinking-coding.preset.json)
[qwen3.6-27b-hauhaucs-thinking-general.preset.json](https://github.com/user-attachments/files/30851596/qwen3.6-27b-hauhaucs-thinking-general.preset.json)
[...]

PR #149, opened 2026-08-09
One-window Mac app: Agents editor, welcome sheet, and the panes that moved into it

Cougler:
Collapses the Mac app from seven windows into one, then rebuilds the surfaces that move into it. Two sessions of work, merged up to `main` (0 behind).

## One window

Models, Tasks, Settings and the four media generators are **modes of the chat window** (`ChatWorkspace`), not `Window` scenes. Each has exactly one door on `AppState` that sets the mode *and* raises the window — setting the mode alone switches a window nobody is looking at; opening the window alone lands on a transcript. Retired scene ids are gone from every surface, because `openWindow(id:)` on an unknown id is a silent no-op, so a leftover id is a control that opens nothing.

The sidebar is one list: destinations and conversations share a gutter, a row inset and a hover fill, with exactly one row lit at a time. It is a `ScrollView`, not a `.sidebar` List — that style adds a horizontal margin `listRowInsets` does not control, so the same "8" drew two different widths and zeroing the insets changed nothing on screen.

## Agents

- The editor is hand-built cards (`AgentEditorChrome`) rather than a grouped `Form`. The Form owned the radii, insets, label typography and section spacing, and right-aligns a TextField's text — which pinned the agent's own name to the trailing edge of a field labelled at the other end. It was also silently promoting `Toggle` to a switch: outside one, a macOS Toggle is a checkbox, so leaving the Form quietly demoted every capability control.
- A thread is **named for its agent**, read live, so renaming an agent renames its rows. The thread's own title becomes the caption — the only thing telling two threads of the same agent apart.
- An empty agent thread greets you as the agent and drops the app's discovery chips. Three of the four navigate out of the conversation and the fourth rewires the composer into a generator the agent has no part in. Nothing is lost: the Tools menu is the always-available twin, and a test pins that it still reaches all four.

## Welcome

Now a **sheet on the chat window** instead of a floating `NSWindow`. That kills a dead end structurally: every exit used to have to remember to open Chat, and a sheet always has a composer behind it. Second card is App / Menu Bar / Terminal — the first two ship in the bundle so they state *Installed* rather than offering a control that could only be dead; the third is the CLI install. Third card plays a silent, looping, chrome-free demo.

## Notable fixes

- **A sheet does not inherit the environment of the view it hangs on.** Hanging the welcome off the chat scene after its whole `.environmentObject` chain still trapped in `WelcomeView.server` on every launch. `SheetEnvironmentAuditTests` closes the class: for every sheet it reads the presented view's own `@EnvironmentObject` declarations, one hop into the children it constructs, and demands each at the presentation site. The existing window audit is blind to it by construction — it asks what the *window* injects, and the bug is where the reading happens.
- **Hiding a window toolbar hides its residents** — the traffic lights and the sidebar-collapse button live in the bar. Retiring an empty band means hiding its *material*.
- **Composer placeholder alignment**: a literal `9` against a real `5 + 7 + 2`, so the caret overlapped its own placeholder. All three insets are one constant now, read by the AppKit editor and the SwiftUI overlay both.
- **Extended attributes break signing**: an asset dragged off the Desktop carries `com.apple.FinderInfo`, and codesign refuses the whole bundle three minutes into a build, naming the binary rather than the file. `build.sh` strips them; `tests/test_resource_xattrs.sh` catches it in a second, checking only the attributes that actually block signing.

## Merge notes

Three conflicts needed judgement rather than a side:

- **Agent sampling.** `#135` added five controls to the editor this branch had just rebuilt. Kept main's seven rows — they are the capability — and moved them onto the card row grammar.
- **Create panes.** main replaced the Model control with a Picker carrying `CustomMediaModels`; this branch replaced it with `MediaModelChooser`. Kept the chooser and fed it the discovery as its own "On This Mac" group.
- **`loadErrorFromName`.** main's refactor kept only the `InsufficientMemory` spelling and dropped `OutOfMemory`, which would have sent an allocator failure during load back to a generic 500. Added the arm, ported the test, deleted the branch's superseded twin.

## Testing

`swift test` **2620 passed, 0 failures**. `zig build test` **9/9 steps, 1406 passed**. Ad-hoc-signed bundle builds and launches clean.

**Not verified visually.** The machine locked partway through, so the merged sampling card, the welcome sheet and the tools demo have not been looked at on screen — only compiled and tested. Worth a pass before merge.


ddalcu:
image

PR #160, opened 2026-08-12
feat: finish Muse-Glimmer DFlash vision support

CerebralCoding:
## Summary

- enable Muse-Glimmer image inputs while retaining the external DFlash drafter; keep the Qwen M-RoPE sidecar guard intact
- clip DFlash output to `max_tokens` before KV-cache, context, and generated-ID state is committed, and publish every token in the final clipped streaming block
- add a sticky adaptive fallback for uneconomic DFlash workloads after three rounds (general threshold: 2 accepted tokens/round; tool threshold: 1)
- expose and test the M5/G17 NAX wide-verification path (`block_size=16`, `wide_verify_lane=true`)
- extend hermetic and end-to-end coverage for exact clipping, streaming, tools, vision, fallback, quantization, and prefix-cache restoration

## Why

The ddalcu Muse-Glimmer conversions contain the vision tower, but their model cards note that image input was not served. In mlx-serve, the remaining blocker was a blanket guard that rejected image-bearing requests whenever an external drafter was active. Muse does not need the Qwen external M-RoPE sidecar path, so this narrows the guard instead of weakening the M-RoPE safety condition.

The existing DFlash loop could also accept a whole draft block beyond `max_tokens`, after internal state had already advanced. Low-acceptance prompts—especially free-form prose and vision—could spend more time verifying drafts than serial decoding. This change makes the budget exact and falls back to the ordinary lazy decoder when cumulative acceptance shows DFlash is no longer economical.

## Performance

Measured on an M5 Max with 64 GB unified memory, temperature 0, using medians of three runs. Values are output tokens/second. The pre-gate and final columns are separate benchmark runs, so small run-to-run variance is expected.

### 4-bit

| Workload | Serial | DFlash before fallback | Final adaptive DFlash |
| --- | ---: | ---: | ---: |
| code/class | 33.3 | 85.7 | 94.4 |
| code edit | 33.2 | 88.2 | 100.7 |
| tool call | 32.8 | 71.2 | 78.5 |
| free-form prose | 33.2 | 32.3 | 33.3 |
| image detail | 33.1 | 26.3 | 32.6 |

### 8-bit

| Workload | Serial | DFlash before fallback | Final adaptive DFlash |
| --- | ---: | ---: | ---: |
| code/class | 17.3 | 37.9 | 33.7 |
| code edit | 17.5 | 78.6 | 68.7 |
| tool call | 17.3 | 44.2 | 49.9 |
| free-form prose | 17.2 | 12.7 | 18.0 |
| image detail | 16.9 | 11.4 | 17.9 |

The fallback preserves the large wins on structured work while removing the material regressions on low-acceptance prose and image prompts. A two-image 4-bit request retained 98% of serial throughput; the corresponding 8-bit run retained effectively 100%.

For category-level context, ddalcu reports on an M4 Max 128 GB:

| Variant | class | edit | tool |
| --- | ---: | ---: | ---: |
| [4-bit](https://huggingface.co/ddalcu/Muse-Glimmer-30B-MLX-Serve-4bit) | 60.9 | 75.5 | 58.9 |
| [8-bit](https://huggingface.co/ddalcu/Muse-Glimmer-30B-MLX-Serve-8bit) | 41.4 | 51.7 | 42.1 |

Those are not like-for-like comparisons—the exact upstream prompts were not published—but the local results are in the same performance envelope and exceed the reported edit/tool numbers in several categories. The upstream integration context is [ddalcu/mlx-serve#159](https://github.com/ddalcu/mlx-serve/pull/159).

## Memory safety

| Variant | Peak used | Minimum free | Swap/OOM |
| --- | ---: | ---: | --- |
| 4-bit | 23.1 GB | 59% | none |
| 8-bit | 37.2 GB | 38% | none |

Both variants returned to roughly 94% free memory after teardown. The 8-bit run remained comfortably below the 64 GB ceiling, with zero swap-in/swap-out activity. The measured peaks also align with ddalcu's model-card estimates of approximately 24 GB and 38 GB.

## Validation

- ReleaseFast build passes
- focused DFlash unit tests: 22/22
- 4-bit DFlash integration: 13/13
- 8-bit DFlash integration: 13/13
- Muse vision end-to-end: OpenAI single image, OpenAI two images, Anthropic image, DFlash engagement, and `--no-vision` rejection all pass
- full suite: 1,464 pass, 122 skip, 2 unchanged unrelated GPU-tolerance failures (`deepseek_v4` fused decode-chain and `minimax_h3` sparse attention)

Boot diagnostics and integration assertions confirm the M5/G17 NAX wide-verification lane is active with the checkpoint-native draft block size of 16.


ddalcu:
The M5 Legend returns! as usual @CerebralCoding  great work here ! Thank you so much.

Ran it locally: builds clean, The max_tokens clip and the streaming publish fix are real bugs fixed, and the vision gate narrowing is right. 

Few things:

**1. The gate guts the hermetic dflash guard**

I put a probe in the equivalence test on your branch:

DFLASH-PROBE attempts=3 accepted=0 disabled=true emitted=16

`dflash: nextDflash greedy equals serial decode` used to run 16 dflash rounds. It now runs 3 and then falls back to `next()`, which is the same code as the serial arm, for the other 13 tokens. So the equivalence assertion is vacuous for most of the test. `test_dflash.sh:[9]` names that test as the numeric guard for the default-on draft weight quantization, so that guard is now 3 rounds deep.

One line fixes it, the option is already plumbed: `.dflash_min_accepted_per_round = 0` in the dflash arm's `InitOptions`. If you want gate coverage, add a separate short arm for it.

**2. Class bug fixed in one arm out of five**

`slot.completion_tokens = gen.completion_tokens` on the first token of a block is what truncated the stream and over-reported usage. Same line is still at scheduler.zig:4960 (dspark), :4983 (mtp), :5007 (drafter), :5055 (pld), and none of `nextDspark`/`nextPld`/`nextDrafter`/`nextMtp` has a max_tokens clip. PLD accepting 5 tokens at completion=15, max_tokens=17 still streams 1 token and reports `usage.completion_tokens: 20`.

House rule is class bugs get class guards. Wants `capAcceptedForTokenBudget` on all five and the `+= 1` publish hoisted into one shared loop, plus the rule line in CLAUDE.md and the story in docs/gotchas/engine-mlx.md.

**Smaller stuff**

* `test_dflash.sh:[4b]` asserts `runtime_disabled=true` for "why the sky is blue". That hard codes that the model is bad at that prompt, and the script runs against both 4-bit and 8-bit. Assert the invariant (if disabled then avg_per_round < threshold) and branch on the model's behaviour.
* The 2.0 threshold is M5 Max, block 16, NAX wide lane. My box probes `available=false` so the block caps at 7, where verify costs something completely different (your own numbers: block 16 at 0.92x serial, block 5 at 1.97x). A fixed absolute threshold across block widths has the same problem as the per-draft percentage you rejected, just inverted. Scale it by block size or say in the comment that it is measured for one config.
* `has_tools` is a proxy for the wrong thing. The stated cause is a low yield reasoning preamble, but muse defaults thinking via `defaultEnableThinking(has_tools)`, so a no-tools request with explicit `reasoning_effort` thinks and gets the strict threshold anyway. Key it on whether the request will think.
* After the gate trips the assistant context stops growing, but scheduler.zig:4028 still commits it beside a full span trunk snapshot, and the comment there ("covers exactly the same span") is now wrong. Restore side rejects it correctly so it is not a correctness bug, but the bytes are billed into `kv_bytes` and the next turn drafts blind (92.6% to 66.5%), which can trip the gate again. Skip the commit when `dc.absLen() != full_prompt.len + generated_ids.len`.

CerebralCoding:
Addressed all findings in `5a1ca4c`:

- Applied pre-commit token-budget caps to all speculative decoders.
- Unified block publishing with exact per-token accounting.
- Made the DFlash gate width-aware and keyed to resolved thinking mode.
- Prevented stale DFlash contexts from entering the prefix cache after fallback.
- Strengthened the equivalence, gate, integration, and class-wide regression tests.

Focused suites and the production build pass. Two unrelated pre-existing GPU tolerance tests remain flaky.

PR #166, opened 2026-08-12
Add Ling 3.0 bailing_hybrid support

justinluque:
Add support for the inclusionAI Ling 3.0 (bailing_hybrid) architecture.

This change adds model-family parsing and validation for the hybrid KDA + MLA MoE design, including:
- config parsing and by-name refusal checks for unsupported variants
- hybrid attention and routing wiring in the transformer stack
- asymmetric KV-cache sizing and memory accounting fixes
- model discovery / advertised architecture handling
- API/default behavior updates for thinking and tool-call support
- integration coverage for Ling 3.0 end-to-end behavior and prefix reuse

This also includes the regression coverage and documentation updates needed to keep the new architecture behavior pinned and reviewable.


ddalcu:
I tried to implement this when the first bigger version of ling came out (not the tiny one), but ended up reverting and giving up on it due to a bug in their model. Not sure if they fixed it yet since.

Have you noticed junk tokens / characters getting injected in your testing ?


A good way to test it, is to **launch Pi**, and ask it to build:
* Build me quake 1, using threejs, generated textures, sounds. Build only the first level. With shotgun, and machine gun, enemies, modular typescript, npm, add high level tests to ensure it compiles.
* Design and create a very creative, elaborate, and detailed voxel art scene of a pagoda in a beautiful garden with trees, including some cherry blossoms. Make the scene impressive and varied and use colorful voxels. Use whatever libraries to get this done using modular js files and write npm tests.

And some others for the regular chat:
* What happens when an unstoppable force meets an immovable object? Answer by alternating the words "moves" and "stops" until the end of the text.this one breaks it quick.
* Write a story where every single sentence must end with the word "apple".
Repeat the word "the" exactly 50 times without stopping.
*What happens when an unstoppable force meets an immovable object? Answer by alternating the words "moves" and "stops" until the end of the text.
*Generate the Fibonacci sequence, but write out the numbers as words (e.g., "one, one, two, three"). After you reach "fifty-five", start subtracting the previous number instead of adding, and do not stop when you hit zero.

There are some additional scripts in tests/* folder 
* test_format_matrix.sh
* test_kv_cache_poison.sh

and a quick one `npx llmprobe@latest http://localhost:11234` 


Because of Ling's very different architecture, I want to avoid adding it, if the model is broken in MLX, but again... its been a few days / a week or so, since I last tested it, so they might of fixed it since.  

I will checkout your branch and run some tests once I get some time.

Thanks !

justinluque:
Hmm interesting, haven't noticed any junk but haven't done extensive testing with it beyond a few small chats. Will mess with it more when I can!

ddalcu:
Had Claude run through it a bit, here are some notes from my Claude to your Claude 😂 
Also ask it do to some extensive testing overnight while you are sleeping with all the /tests/* to make sure other models don't break, etc.  Overall once these are we should be in good shape to merge it.

### 1. Server segfaults on `--kv-quant turbo2|turbo4` instead of refusing

Repro: `mlx-serve --model <Ling-3.0-tiny-MLX-4bit> --serve --kv-quant turbo4`. The named refusal prints, then the process dies:

    --kv-quant turbo: cache key width 192 is not a power of two ...
    EXC_BAD_ACCESS in mlx_array_free <- transformer.freeKVEntry <- KVCache.deinit
      <- Transformer.deinit <- scheduler.doLoadOnInferenceThread

Cause is the caller pattern, not the new check. `initWithConfigAndHeadDim` is now fallible, but the sites that re-apply the launch kv-quant config still `deinit()` the old cache first:

    xfm_ptr.cache.deinit();
    xfm_ptr.cache = try KVCache.initWithConfigAndHeadDim(...);   // scheduler.zig

On the error path that leaves a freed cache on the Transformer, and `Transformer.deinit` frees every handle a second time. Four sites: `scheduler.zig` (doLoadOnInferenceThread), `main.zig`, `Transformer.resetCache`, `Transformer.restore`.

Fix is to build first, then swap. A `KVCache.reinit` helper covers all four:

    pub fn reinit(self: *KVCache, num_layers: u32, config: KVQuantConfig, head_dim: u32) !void {
        const fresh = try initWithConfigAndHeadDim(self.allocator, num_layers, config, head_dim);
        self.deinit();
        self.* = fresh;
    }

I wrote a regression test for it and it goes red exactly like the live crash (`terminated with signal SEGV`, poisoned offset `6148914691236517205`), green after. Worth adding, since this is a class bug: any fallible re-init behind a deinit has the same shape.

### 2. `defaultEnableThinking` contradicts the PR's own test and docs

`model.zig` gates on `has_tools`:

    if (has_tools and (std.mem.eql(u8, self.model_type, "bailing_hybrid") or ...)) return true;

but the checkpoint's `chat_template.jinja` defaults `thinking_option` to `'on'` when `enable_thinking` is undefined, and the PR's own docs say thinking is default ON (CLAUDE.md arch row, `docs/reference.md`, `tests/CLAUDE.md`). Test `[2]` in `tests/test_bailing_hybrid.sh` asserts reasoning on a tool-less silent request and fails live. The unit test in `model.zig` pins the opposite (`try testing.expect(!ling.defaultEnableThinking(false))`), so the two tests disagree with each other.

It's not cosmetic. At temp 0, no tools:
- default (thinking off): "17 - 9 = **8**"
- `enable_thinking: true`: "9 sheep are left"

Unlike muse there's no prompt-committed no-think channel to fall back to, so the tool-less arm just makes a reasoner answer without reasoning. Suggest dropping the `has_tools` gate for `bailing_hybrid` and flipping the unit test expectation. Verified: 23/23 after that change.

### 3. Leftover git-stash artifacts committed in the branch

- `tests/CLAUDE.md` around line 98: a literal `<<<<<<< Updated upstream` marker, followed by both versions of the table, so there are two `test_muse_glimmer.sh` rows (one stale).
- Trailing ` Stashed changes` on a line in `tests/CLAUDE.md`, `docs/gotchas/models-media.md`, and `docs/gotchas/server-http.md`.

### 4. Needs a rebase on main

Currently `CONFLICTING`, 3 commits behind. Conflicts are in `src/transformer.zig` (your new KVCache tests land next to the sliding-window-trim tests from 30492fa) and `docs/gotchas/engine-mlx.md` (both sides appended a section). Both are additive, keep both sides. The `updateDense` view-building refactor auto-merges against the sliding trim, so please give that a second look after rebasing.

### Minor

- `defaultEnableThinking` has a `"ling"` branch, but no such `model_type` exists in the tree. Dead code.
- Same function: the muse comment was rewritten to "always deliver reasoning (its templates open a `to=self` pass)", which no longer describes the code (it's still `has_tools` gated). Please restore the original comment, it's an unrelated file region.
- `attnCacheLayerCount` now excludes linear layers for every hybrid, not just this arch, so the prefill admission guard bills roughly 4x less on qwen3.5/3.6. It's more accurate, but it loosens a calibrated guard, so worth calling out in the PR description. On Qwen3.5-0.8B auto-context still clamps at `max_position_embeddings`, so nothing visible there.
- `kvBytesPerToken` uses `num_key_value_heads` for the MLA case. It's right for this checkpoint (16 == `num_attention_heads`), but MLA expands K/V to all heads, so it'd be safer read off `num_attention_heads`.

Nice work overall. The KDA vector-gate kernel equivalence test (per-channel gate held uniform per head must match the scalar kernel exactly) and the by-name config refusal table are exactly the right guards for this.

justinluque:
image
image

As far as I can tell it's not broken, let me know what you think! Maybe something did change?

ddalcu:
yeah, i did some testing also, seems like they fixed it in the tiny version, hopefully the big one too.

PR #167, opened 2026-08-13
build: FAST_DEV iteration builds

justinluque:
FAST_DEV=1 bash app/build.sh trades everything a release needs for iteration speed; an unset var is the release path byte for byte, which the normalized diff against the previous script confirms line by line.

What FAST_DEV changes: Swift compiles -c debug (the big lever — release mode is WMO, so one touched file recompiles the module); CalVer skips the gh round-trip and leaves the tracked Info.plist alone (a dirty tree is what breaks release.yml's Homebrew rebase); the bundle updates in place; the icon is regenerated only when the bundle has none; the frameworks are restaged, fixed up and re-signed as ONE decision, skipped while lib/mlx is no newer than the bundled copy — install_name_tool rewrites a file and invalidates its signature, so skipping the re-sign while still running the fixups leaves broken signatures in the bundle; and it stops at a signed .app, before notarization, productbuild and the DMG.

mlx-serve stays ReleaseFast even under FAST_DEV: zig's cache already makes an unchanged rebuild a few seconds, so Swift is the whole cost of an iteration. ZIG_DEBUG=1 is the opt-in arm, deliberately a second lever — a Debug engine decodes 2-4x slower, so any latency read off that app is fiction, and zig keys its cache on the optimize mode, so alternating with a release build is a full rebuild each way. It is FAST_DEV-only and refused by name otherwise, because FAST_DEV is what stops before the shipping artifacts.

Guard (tests/test_release_workflow_gates.sh, which already scanned build.sh for the CalVer TZ pin and license packaging): 14 checks pinning that an iteration lever cannot become a shipping mode. Verified red against the working draft this replaces — including "the FAST_DEV exit precedes notarytool submit", which that draft failed because it had replaced the notarization and App Store .pkg blocks with "# ... (rest of the block is identical)".

Not sure if you'd want to merge this or not, but figured I'd make a pull for it


ddalcu:
yeah.. not a bad idea, swift has got really slow to compile.  Im not too sure about the FAST_DEV var, maybe we just pass a arg... merging for now because I feel your pain, and maybe I can adjust it a bit later.

PR #168, opened 2026-08-13
feat: gpt-oss

justinluque:
Not complete yet but returns a few tokens before crashing


justinluque:
image
 
Seems to be in working condition

justinluque:
image

51 tok/s for 20b with M4 Pro

ddalcu:
why support GPT OSS though ?   So many better models out there

justinluque:
I think the app benefits from supporting a diverse range of architectures beyond the frontier of local models, imo we should let the user decide which model is better for their particular purposes. I implemented it particularly because I use it personally.

PR #171, opened 2026-08-13
add DeepSeek V4 DwarfStar SSD backend

PhilipJohnBasile:
## What changed

- add a reproducible DwarfStar backend for DeepSeek V4 Flash 0731 on M5 Max
- keep target-only SSD streaming bounded with an 8 GiB expert-cache default and no full-model warmup
- support explicit DSpark without silent target-only or legacy-MTP fallback
- propagate SSD/MTP/DSpark flags through all startup and cold-switch load paths
- add truthful nested-engine provenance, sidecar lineage checks, and ABI/layout guards
- pin the reviewed public DwarfStar dependency and include a bounded evidence receipt

## Why

DeepSeek V4 Flash does not fit comfortably as a fully resident model on a 128 GiB Mac once runtime headroom is included. The practical route is target-only SSD streaming. Optional DSpark support needed separate persistent Metal lifetime, complete readiness and memory admission, and fail-closed content checks before it could be exposed safely.

## Validation

- focused `ds4` tests in Debug, ReleaseSafe, and ReleaseFast
- DSpark content/readiness and both lineage-mismatch directions
- SSD admission boundary, insufficient-headroom, and overflow tests
- persistent Metal support lifecycle ordering
- Metal `ds4-server` and CPU builds
- clean published-submodule ReleaseFast wrapper build
- fresh-clone/submodule reproducibility and independent frozen-diff review

Final clean binary reported `ds4 0c654079b8fd`. No real model/GPU benchmark was run after the final repair. DSpark remains experimental and default-off because the controlled local A/B was slower with zero accepted draft tokens; target-only SSD streaming is the recommended route.

Nested dependency: antirez/ds4#798.


ddalcu:
lets re-open this as a fresh one once its merged in the main branch, for now the latest ds4 is pin has been updated in main

PR #174, opened 2026-08-14
feat: add "Only use tools when I ask" setting for opt-in tools

justinluque:
Adds a new setting in Settings ▸ Agent Workspace that disables the pre-send nudge for tool and MCP activation. When toggled ON, the composer no longer intercepts messages that look like tasks (e.g., "make me a website" or "npm install") to suggest turning on Tools or MCP. Tools must be explicitly enabled via the toolbar or a locked agent.

Refactoring:
* Extracts the pre-send nudge decision from `ChatView` into a single, pure function `ComposerIntent.nudge`.
* Introduces comprehensive unit testing for the nudge logic in `ComposerInputTests` (validating MCP priority, agent lock overrides, suppression state, and the new opt-in setting).
* Updates `CLAUDE.md` to reflect the centralized, single source of truth for tool opt-in logic.
* Includes the new `toolsOnlyWhenAsked` flag in Settings Reset and ensures it does not trigger a server restart when toggled.


ddalcu:
@justinluque on a mission to MLXServe Swift UI so good, that it will replace all other agent harnesses 😂 !

PR #189, opened 2026-08-16
feat: render LaTeX in assistant chat

wasd-c:
## Summary

- Render inline and display LaTeX natively in assistant responses with SwaTex.
- Recognize `$...$`, `$$...$$`, `\(...\)`, `\[...\]`, and common equation environments while leaving user prompts, code spans, fenced code, currency, incomplete streams, and invalid TeX readable as source.
- Preserve the original delimiters when inline equations are copied, support light/dark appearance, and make display equations horizontally scrollable.
- Package SwaTex's KaTeX fonts and license notices through SwiftPM, the Xcode project, local app assembly, and release CI.

## Why

Local math models commonly emit TeX even when the chat surface cannot render it. That makes otherwise useful mathematical answers hard to read. This keeps the renderer native and offline instead of adding a WebView or JavaScript path.

## Build notes

SwaTex 0.5.0 declares Swift 6 for its own targets. The app target remains in Swift 5 mode, but the old global `-swift-version 5` override had to be removed so every package can compile in the language mode declared by its own manifest.

## Testing

- `zig build test --summary all` — 9/9 steps succeeded
- `swift test` — 2,733 tests executed, 24 skipped, 0 failures
- `swift build -c release`
- `FAST_DEV=1 APPLE_DEVELOPER_ID=- APPLE_TEAM_ID=- SKIP_NOTARIZE=1 bash app/build.sh` — ad-hoc bundle built and passed strict code-sign verification
- Verified the built app contains the exact `SwaTex_SwaTexRender.bundle`, including 20 KaTeX fonts, `OFL.txt`, and `FONT_NOTICE.txt`
- Regenerated `MLXCore.xcodeproj`, resolved SwaTex 0.5.0, and verified Xcode copies the resource bundle into the app. The full local MAS build remains gated by the existing guest-rootfs phase because Docker is not running.
- `bash -n app/build.sh`, YAML parse checks, and `git diff --check`


ddalcu:
Looks good, thank you ! Merged

PR #193, opened 2026-08-16
fix: LC_RPATH is resolved against the launching cwd, so the binary only starts from the repo root

### LINKED ISSUE #192: Locally built mlx-serve binary only runs from the repo root (relative LC_RPATH)
A locally built `mlx-serve` binary only runs when the current working directory is the repo root. Run it from anywhere else and it dies immediately with a dyld error, even with an absolute path to the binary.

## Reproduction

```
$ cd /path/to/mlx-serve
$ zig build -Doptimize=ReleaseFast
$ ./zig-out/bin/mlx-serve --help          # works
$ cd /tmp
$ /path/to/mlx-serve/zig-out/bin/mlx-serve --help
dyld[2831]: Library not loaded: @rpath/libllama.dylib
  Referenced from: <AB4CF513-6982-3C8F-8E8F-29C39382488C> /path/to/mlx-serve/zig-out/bin/mlx-serve
  Reason: tried: 'lib/llama/lib/libllama.dylib' (no such file), 'lib/mlx/lib/libllama.dylib' (no such file), 'lib/llama/lib/libllama.dylib' (no such file), 'lib/mlx/lib/libllama.dylib' (no such file)
```

## Cause

```
$ otool -l zig-out/bin/mlx-serve | grep -A2 LC_RPATH
          cmd LC_RPATH
      cmdsize 32
         path lib/llama/lib (offset 12)
--
          cmd LC_RPATH
      cmdsize 24
         path lib/mlx/lib (offset 12)
```

Both `LC_RPATH` entries are **relative** paths (`lib/llama/lib`, `lib/mlx/lib`). An `LC_RPATH` is resolved by dyld against the process's current working directory at launch, not against the binary's location or the directory it was built in. So the binary only finds its libs when launched with cwd == repo root — anywhere else, dyld can't resolve `@rpath/libllama.dylib` (or `@rpath/libmlxc.dylib`) at all.

`build.zig` sets these via `module.addRPath(b.path("lib/llama/lib"))` (`addLlamaLib`) and `module.addRPath(b.path("lib/mlx/lib"))` (`addMlxLib`). `b.path(sub_path)` resolves against `b.build_root`, and when `zig build` is invoked with cwd already equal to the build root (the normal `zig build -Doptimize=ReleaseFast` from CONTRIBUTING.md), `b.build_root.path` is `null` ("means cwd"), so the join produces a bare relative string with nothing prefixed — that's exactly the relative rpath baked into the binary.

This doesn't affect the packaged release (CLI tarball / .app), because `release.yml` / `app/build.sh` rewrite the `LC_LOAD_DYLIB` references themselves via `install_name_tool -change ... @executable_path/...`, independent of these `LC_RPATH` entries. It only affects a binary built and run straight out of `zig-out/bin` — which is exactly what `CONTRIBUTING.md`'s "Build" section walks a contributor through, and what anyone hacking on the source and then just running the compiled binary would naturally do (I hit it symlinking `zig-out/bin/mlx-serve` for local testing outside the repo).

## Environment

- macOS 27.0 (26A5406e), Apple Silicon (M3 Max)
- Zig nightly staged via `scripts/fetch-zig.sh` (`0.17.0-dev.1422+e863bf3be`)
- libllama staged via `scripts/fetch-llama.sh`, mlx/mlx-c staged via `scripts/build-mlx.sh`

## Suggested fix

Bake an absolute path into the `LC_RPATH` at build time instead of a relative one, e.g. resolve `lib/llama/lib` / `lib/mlx/lib` to an absolute path in `build.zig` before passing to `addRPath`. I have a small patch that does this and verified it fixes both directions (repo root and elsewhere) without touching the release packaging path — happy to open a PR if useful.

Fe2-O3:
Fixes #192.

`zig-out/bin/mlx-serve` only ran when the current working directory was the repo root. From anywhere else:

```
dyld[2831]: Library not loaded: @rpath/libllama.dylib
  Referenced from: <AB4CF513-6982-3C8F-8E8F-29C39382488C> /path/to/mlx-serve/zig-out/bin/mlx-serve
  Reason: tried: 'lib/llama/lib/libllama.dylib' (no such file), ...
```

## Cause

`otool -l zig-out/bin/mlx-serve | grep -A2 LC_RPATH` showed both rpaths as relative strings (`lib/llama/lib`, `lib/mlx/lib`). `LC_RPATH` is resolved by dyld against the process's cwd at launch, not against the build directory, so a relative rpath only works from one specific cwd.

`addLlamaLib`/`addMlxLib` in `build.zig` set these via `module.addRPath(b.path("lib/llama/lib"))` / `module.addRPath(b.path("lib/mlx/lib"))`. When `zig build` runs with cwd already at the build root (the normal `zig build -Doptimize=ReleaseFast` from CONTRIBUTING.md), `b.build_root.path` is `null` ("means cwd"), so `Directory.join` returns the bare relative string with nothing prefixed — exactly what ends up baked into the binary.

This doesn't touch the packaged release: `release.yml` / `app/build.sh` already rewrite the `LC_LOAD_DYLIB` references directly via `install_name_tool -change ... @executable_path/...`, independent of these `LC_RPATH` entries.

## Fix

Resolve `lib/llama/lib` and `lib/mlx/lib` to absolute paths at configure time before handing them to `addRPath`, using the build root's existing open directory handle (`buildRootHandle(b).realPathFileAlloc(...)`) — the same pattern `readLlamaTag` already uses a few lines up to read `lib/llama/.version`.

## Testing

- `zig build -Doptimize=ReleaseFast`, then:
  - `otool -l zig-out/bin/mlx-serve | grep -A2 LC_RPATH` now shows absolute paths (`/…/lib/llama/lib`, `/…/lib/mlx/lib`)
  - `./zig-out/bin/mlx-serve --help` from the repo root: starts (unchanged)
  - `/absolute/path/to/zig-out/bin/mlx-serve --help` from `/tmp`: now starts too (previously died with the dyld error above)
- `zig build test`: green

No changes to `src/transformer.zig` or anything unrelated — this PR is scoped to the rpath fix only.

### DIFF
--- a/build.zig
+++ b/build.zig
@@ -486,23 +486,6 @@ fn buildRootHandle(b: *std.Build) std.Io.Dir {
     return b.root.root_dir.handle;
 }
 
-/// Absolute path to `sub_path` under the build root, for baking into an
-/// LC_RPATH. `b.path(sub_path)` looks like it would do this, but when `zig
-/// build` is invoked with cwd == build root (the normal `zig build
-/// -Doptimize=ReleaseFast` from CONTRIBUTING.md), `b.build_root.path` is
-/// null ("means cwd") and `Directory.join` returns the bare relative
-/// string with nothing prefixed. That relative string is exactly what
-/// ends up in the linked binary's LC_RPATH — fine for the linker
-/// (build-time cwd is the build root), wrong for the binary, whose
-/// LC_RPATH is resolved against cwd at process launch, which can be
-/// anywhere. Resolve to an absolute path once here so the rpath keeps
-/// working regardless of where the binary is later run from.
-fn absBuildPath(b: *std.Build, sub_path: []const u8) []const u8 {
-    return buildRootHandle(b).realPathFileAlloc(b.graph.io, sub_path, b.allocator) catch |err| {
-        std.debug.panic("failed to resolve '{s}' under the build root: {t}", .{ sub_path, err });
-    };
-}
-
 /// The llama.cpp tag staged by scripts/fetch-llama.sh (it writes LLAMA_TAG to
 /// `lib/llama/.version`). Read at configure time so a plain `zig build` reports
 /// the real tag without app/build.sh having to pass `--llama-tag`. Returns null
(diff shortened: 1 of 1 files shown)


Fe2-O3:
Updated this to a better fix. The previous commit resolved `lib/llama/lib` and `lib/mlx/lib` to absolute paths at configure time and baked those into `LC_RPATH`. That worked on my machine but the path is specific to where I happened to build, so the binary was not relocatable — anyone building from a different directory would get a different, still-wrong absolute path compiled in.

The new commit uses `@loader_path` instead. dyld resolves `@loader_path` against the binary's own location at launch, not against the process's cwd (the original bug) and not against a fixed path baked in at build time (the previous fix's problem). Each `addRPath` call now lists two `@loader_path`-relative entries, one for the installed exe (`zig-out/bin/mlx-serve`) and one for the `zig build test` binary (`.zig-cache/o/<hash>/test`), since those two binaries land at different depths under the build root. dyld tries every `LC_RPATH` entry in order and silently skips ones that don't resolve, so listing both is safe.

Verified on this machine:
- starts from the repo root
- starts from `/tmp` via an absolute path
- `otool -l zig-out/bin/mlx-serve | grep -A2 LC_RPATH` shows only `@loader_path/...` entries, no absolute path
- `zig build test` passes
- copied `zig-out/` + `lib/` to an unrelated directory and the copy still starts, confirming relocatability

ddalcu:
thanks, simple one, should help others.

PR #199, opened 2026-08-16
Save video generation settings and prompt to text file on generation

Morac2:
## Summary

Adds a human-readable .txt sidecar beside each generated video MP4.

- Records the prompt, model, seed, requested settings, and returned output properties.
- Records effective video options, LoRAs, and reference media using portable basenames only.
- Writes sidecars for both Video window and agent-driven generations.
- Keeps a successfully encoded MP4 even if writing its sidecar fails.

## Why

Generated videos currently have no adjacent record of the prompt or settings that produced them. This makes them difficult to reproduce or compare later.

## Validation

- swift test --filter VideoSettingsSidecarTests (7 tests)
- swift test --filter MediaGenServiceTests (70 tests)
- swift test (2,935 passed, 24 skipped)
- swift build -c release


ddalcu:
thanks, merged.

PR #201, opened 2026-08-16
fix: disconnect the MCP client when connectOrFailFast loses its race

slava-kudzinau:
## What

A live report: an MLX Core session hung mid-chat after 3–4 turns of RAG-style tool calling. `mlx-serve` itself was idle and healthy (`/health` sub-2ms, `requests_running=0`); the leak was entirely app-side. `sample <pid>` showed two threads with 657/1000 samples each sitting inside `closure #1 in Client.connect(transport:)` — 5326 threads and ~160% sustained CPU over a 23h uptime.

## Root cause

`MCPManager.connectOrFailFast` races `client.connect(transport:)` (Path A) against a death-watcher and a hard 30s timeout. When the other arm wins, Path A's `Task` is abandoned — by design, per the function's own doc comment — but nothing on the losing paths ever called `client.disconnect()`. Per the vendored SDK (`swift-sdk .../Client.swift`), `disconnect()` is the ONLY thing that resumes a pending request continuation; cancelling the task or killing the child does not unstick it. Each MCP server that failed to answer `initialize` in time (or died slowly enough to lose the race) leaked one permanently-scheduled task, starving the cooperative thread pool until the next chat turn hung on a tool call routed through it.

`executeToolCall`'s watchdog already documents this exact SDK behavior and applies the break-glass correctly for tool calls — the connect-time race just never got the same treatment.

## Fix

Both losing paths (death-watcher and hard-cap timer) now call `client.disconnect()` before resuming the outer continuation; the timeout path also `child.terminate()`s in case the process is still alive. `connectOrFailFast` gained an injectable `hardCapSeconds` (default unchanged at 30) and an `onPathASettled` hook so the regression test can force the timeout path in well under a second.

## Guard

`MCPConnectLeakTests.testAbandonedConnectSettlesAfterLosingTheRace` — spawns a real `sleep 60` child (accepts the stdio pipes, never speaks MCP so the death-watcher never fires), forces a 0.3s hard cap, and asserts the abandoned Path A task settles within 5s instead of never.

## Docs

`docs/gotchas/app.md`: full war story + the rule — any race that abandons one arm on the other's win must ask whether the abandoned arm can ever unstick itself.


ddalcu:
thank you, looks good 👍

PR #203, opened 2026-08-16
fix: detach per-connection threads so completed stacks are reaped

sf-jin-ku:
## Summary

- detach each successful per-connection `std.Thread` immediately after spawn
- keep the existing `active_conn_threads` shutdown barrier unchanged
- add a source guard plus a headless macOS `vmmap` regression test covering 300 fresh HTTP connections
- document why MLX cache controls cannot reclaim pthread stacks

## Root cause

The accept loop discarded the handle returned by `std.Thread.spawn`. Returning from `handleConnectionThread` stopped execution, but the joinable pthread retained its stack mapping and kernel bookkeeping because its handle was never joined or detached.

In the captured long-running workload, MLX `/props` reported only about 2.9 MB of allocator cache while the process had 155,786 stack regions using about 2.43 GB plus about 1.01 GB of page tables. Sending 300 requests added 301 more stack regions, which made an otherwise stable 4–5 GB model workload climb toward 10 GB.

## Fix

Retain the successful spawn result as `conn_thread` and call `conn_thread.detach()` before accepting the next connection. Detach handles pthread resource reclamation after the handler returns; `active_conn_threads` remains the separate lifetime barrier that prevents scheduler teardown while connection threads are still running.

No inference, scheduler, KV-cache, or MLX allocator behavior changes.

## Guard

- `each connection thread handle is detached after spawn` pins the accept-loop ownership rule
- `tests/test_connection_thread_reaping.sh` boots a model-free server, records its Stack-region count, sends 300 new connections, and asserts the count remains bounded
- local fixed run: `before=4 after=3 delta=-1 requests=300`

## Testing

- `.zig-toolchain/zig build -Doptimize=ReleaseFast` — passed
- focused Zig regression test — passed
- `./tests/test_connection_thread_reaping.sh` — passed
- `swift test --package-path app` — 2,928 tests, 24 skipped, 0 failures
- full `zig build test` — 1,588 passed, 136 skipped, 3 GPU numerical tests failed locally (`verifyQmm`, DSV4 fused decode-chain, H3 sparseAttend); the new server test passed and the PR CI run on the project macOS-26 runner is the portability check
- `test_shutdown_midstream.sh` — skipped locally because its default small checkpoint was not installed


ddalcu:
thanks !

PR #205, opened 2026-08-17
feat: interleave decode ticks at prefill chunk boundaries (stall-free scheduling)

sf-jin-ku:
## Problem

A cold prefill runs to completion in the scheduler's step 2 before any decode tick (step 4), so every concurrent stream freezes for the *entire* prefill of an incoming request. Measured on gemma-4-E2B-QAT-4bit (M4-class, 48 GB):

- 9.1k-token cold prefill → every active stream stalls **0.80 s** (normal inter-token gap: 10.5 ms)
- 52.9k-token prefill → **7.6 s** freeze
- 4 concurrent ~2.2k-token requests → all TTFTs collapse to the LAST prefill's finish (0.35 / 1.32 / 1.33 / 1.35 s) because step 2 finishes the whole admitted batch before any decode tick

The stall is unbounded: it grows linearly with the incoming prompt.

## Fix

Interleave decode ticks at the prefill chunk boundaries that already exist (the memory ladder's chunks), plus one tick between the slots of an admitted batch:

- `generate.zig`: the chunk loop yields through a new `interleave_hook` after each chunk's eval + allocator clear — never after the last chunk (the post-prefill decode tick covers it).
- `scheduler.zig`: the hook runs one `runDecodeTick` for the streams already decoding (`interleaveDecodeTick`). The prefilling slot is not in `decoding` yet, so the tick only advances *other* requests — same thread, no reentrancy. Interleaved tick time is charged to the decoding slots and subtracted from the prefiller's `prefill_ns`, so `prefill_tps` stays a statement about the prefill forward.

No new flag: the existing `--prefill-chunk` boundary is the yield point. Kill switch `MLX_SERVE_PREFILL_INTERLEAVE=0` restores whole-prefill scheduling (same idiom as `MLX_SERVE_SLIDING_BLOCK_TRIM`). Happy to flip the default to opt-in if you prefer — it's one line.

**Why default-on is safe:** the hook no-ops when nothing is decoding (idle/single-stream cost is zero), and greedy output is byte-identical either way — interleaving reorders ticks, never math. Memory: yields happen *after* each chunk's `mlx_eval` + `mlx_clear_cache`, so chunk transients and decode-tick transients never coexist; the admission guard's peak model is unchanged.

## Measured (same binary, env A/B, medians of 3)

| | interleave OFF | ON | ON + `--prefill-chunk 1024` |
|---|---|---|---|
| stream's max stall during an ~11k-token cold prefill | **1.35–1.55 s** (whole prefill) | 0.35–0.79 s (one chunk) | **0.149–0.154 s** |
| 4 concurrent TTFTs | 0.35 / 1.32 / 1.33 / 1.35 | **0.27 / 0.55 / 0.84 / 1.13** | — |
| stream's greedy output | `19c1cc9ca0fc` | `19c1cc9ca0fc` (identical) | identical |

The stall floor tracks the chunk width exactly — the fairness/latency knob is the chunk infrastructure this repo already has. Aggregate throughput is unchanged (this reorders work; it does not create it).

## Precedent in other engines

This is the scheduling policy [Sarathi-Serve (OSDI'24)](https://arxiv.org/abs/2403.02310) formalized as stall-free batching, now standard across serving engines:

| Engine | Mechanism |
|---|---|
| [vLLM](https://docs.vllm.ai/en/latest/configuration/optimization.html) | Chunked Prefill — prefill chunks batched alongside decodes; in V1 "enabled by default whenever possible" |
| [DeepSpeed-FastGen](https://arxiv.org/abs/2401.08671) | Dynamic SplitFuse — long prompts decomposed and composed with generation in one forward |
| [TensorRT-LLM](https://nvidia.github.io/TensorRT-LLM/advanced/gpt-attention.html) | Chunked Context — context chunks batched with generation-phase tokens |
| [SGLang](https://github.com/sgl-project/sglang) | `--chunked-prefill-size` |
| [llama.cpp](https://github.com/ggml-org/llama.cpp) | server continuous batching splits prompt work by `n_batch`/`n_ubatch` so running slots keep generating between pieces |
| [vllm-mlx](https://github.com/waybarrios/vllm-mlx) | `chunked_prefill_tokens` — prefill budget per step, one generation step for all active requests between chunks (`scheduler.py`, `_install_chunked_prefill`) |

The CUDA engines mostly run the *mixed-forward* variant (prefill + decode tokens in one batch, which needs varlen/ragged attention). This PR is the *alternate-forwards* variant vllm-mlx uses on MLX — decode ticks between chunk forwards — so it needs no attention or kernel changes.

## Testing

- New: chunk-loop yield scan (generate.zig), hook-wiring + `prefill_ns`-billing scan (scheduler.zig) — both written red first.
- New: `tests/test_prefill_interleave.sh` — two-arm live A/B asserting the `[interleave] engaged` line on/absent per arm, max-gap under half of the off arm (measured 428 ms → 114 ms), and identical output hash. Model-gated, SKIPs without one.
- Full `zig build test -Doptimize=ReleaseFast`: 1590 pass / 136 skip / 3 fail — the 3 are pre-existing GPU-numeric failures (verifyQmm, DSV4 decode-chain, H3 sparseAttend) that fail identically on clean `main` on this machine.
- Swift app untouched (no new CLI flag, `ServerOptions` unchanged).

Possible follow-up (separate PR if wanted): a dynamic chunk cap that only engages while streams are decoding — full-size chunks when alone (zero prefill tax), bounded stall under contention, no operator knob at all.


ddalcu:
thank you, this one was on my long list of todo !

PR #215, opened 2026-08-17
Separate starting the server from loading a model (#214)

### LINKED ISSUE #214: "Auto-start on launch" also auto-LOADS the selected model; please split it into two settings (load defaulting OFF)
## Summary

"Auto-start on launch" is a single toggle, but it does two very different things: it starts the HTTP server **and** it loads the selected model into unified memory. Those should be two independent settings, and the load half should default to OFF.

Starting the server is nearly free. Loading a model is not — on a large pack it is tens of GB resident before the user has typed anything.

## Where the two get welded together

`app/Sources/MLXServe/AppState.swift:491-493`:

```swift
// Auto-start server if enabled and a model is available
if autoStartServer, !selectedModelPath.isEmpty {
    server.start(modelPath: selectedModelPath, options: serverOptions)
}
```

`ServerManager.start(modelPath:options:)` always passes `--model`:

`app/Sources/MLXServe/Services/ServerManager.swift:99-113`

```swift
var args = ["--model", resolvedModel]
```

And `--model` in serve mode is an eager, blocking load — the listener does not open until the weights are resident:

- `src/main.zig:1258` — non-headless branch builds `LoadParams` with `no_initial_load = false` and calls `server_mod.serve(...)`
- `src/server.zig:1205` — `serve()` calls `Scheduler.init(...)` *before* binding the listener
- `src/scheduler.zig:1305` — `Scheduler.init` blocks on the `started` condvar
- `src/scheduler.zig:2959-2962` — `model_mod.loadWeights[WithVision](...)`, the actual weight read

So: tick one checkbox, get a resident model at every login.

## The server already supports the split

A headless boot loads nothing:

- `src/main.zig:1032-1035` — no `--model` → `runHeadlessServe(...)`
- `src/main.zig:1755` — headless `LoadParams` sets `.model_dir = ""`, `.no_initial_load = true`
- `src/scheduler.zig:3451-3453` — `"Headless: no primary model loaded; models load on demand."`
- `src/model_registry.zig:602-605` / `:660-677` — `--model-dir` discovery only registers metadata stubs (`.state = .unloaded`, all model pointers null)

Weights then arrive on demand via `POST /v1/load-model` (`src/server.zig:3755`) or on the first request that names a model (`src/server.zig:1860`).

The app already has the headless entry point too — `ServerManager.startHeadless(modelsDir:options:)` (`ServerManager.swift:149-164`) — plus `ensureDefaultChatModel(selectedModelPath:)` (`ServerManager.swift:139-147`), which hot-loads the selected chat model on the first chat turn when the server was booted headless. Everything needed for "server up, nothing resident" is already built and already exercised by the media-first path.

## Request

Split the one toggle in `StatusMenuView.swift:400` into two:

1. **Start the server on launch** — default ON, unchanged behaviour, but routes through `startHeadless(modelsDir:)` so nothing is resident.
2. **Load the selected model on launch** — default **OFF**. When on, keep today's `start(modelPath:)`.

With (2) off, the existing `shouldEnsureChatDefault` path already hot-loads the model on the first chat turn, so the user-visible cost is one first-message delay instead of a permanent memory reservation. Existing users should migrate to (1) ON / (2) OFF rather than inheriting the current conflated behaviour, since the current behaviour is the surprising one.

## Why it matters

On my machine the selected model is a 27B MXFP8 vision pack. Auto-start meant it was resident from login, unnoticed, competing with everything else. Combined with other memory pressure the system ended up ~97% into swap before I worked out that a model I had never asked to load was holding the memory. The toggle's label — "Auto-start on launch" — gives no hint that a model load is part of it.

A user who wants an always-available endpoint for a coding agent generally wants the *port* to answer, not a specific model pinned in RAM forever.

## Environment

- MLX Core 26.8.8 (`com.dalcu.mlx-core`), macOS
- Repo checked at `v26.8.8-3-gc2e7956`
- Relevant defaults observed: `autoStartServer = 1`, `selectedModelPath = /Users/…/.mlx-serve/models/Shiftedx/Qwythos-27B-v1-MXFP8-Vision-MLX`

Fe2-O3:
Written by an agent. Not compiled, not tested. The machine this was written on has no Zig toolchain and 2.6 GB of free disk, so nothing was built. Please treat this as a suggestion to verify, not a ready patch. Staying a draft until someone builds it.

The only check that was run is `swiftc -parse` on each edited file, which is **syntax only** — it does not type-check, so it proves nothing about whether this compiles.

Closes #214 (if you agree with the shape).

---

## The problem: one checkbox does two things

`AppState.swift` (the launch gate, before this change):

```swift
// Auto-start server if enabled and a model is available
if autoStartServer, !selectedModelPath.isEmpty {
    server.start(modelPath: selectedModelPath, options: serverOptions)
}
```

`ServerManager.start(modelPath:options:)` always passes `--model`:

```swift
func start(modelPath: String, options: ServerOptions) {
    guard status != .running, status != .starting else { return }
    let resolvedModel = Self.launchModelPath(modelPath)
    currentModelPath = resolvedModel
    var args = ["--model", resolvedModel]
```

And on the server side `--model` is an **eager, blocking** load:

- `main.zig:1258` — sets `no_initial_load = false`
- `server.zig:1205` — `Scheduler.init` runs *before* the listener binds
- `scheduler.zig:1305` — blocks until the checkpoint is loaded

So the single checkbox in `StatusMenuView.swift`:

```swift
Toggle("Auto-start on launch", isOn: $appState.autoStartServer)
```

silently reads the entire selected checkpoint off disk at login. With a 26.7 GB model selected — which is what my own setting points at — that is a very long, very invisible startup.

## The server already supports what we want

Nothing needs to change in Zig. Starting without a model is an existing, supported path:

- `main.zig:1032-1035` — no `--model` routes to `runHeadlessServe`
- `main.zig:1755` — headless sets `no_initial_load = true`
- `scheduler.zig:3451-3453` — *"Headless: no primary model loaded; models load on demand."*

The model then loads at `server.zig:3755` (`/v1/load-model`) or `server.zig:1860` (first request). The app already had both halves too — `ServerManager.startHeadless(modelsDir:options:)`, used today by `ensureServerForLan()` and the media-gen path, and `ServerManager.ensureDefaultChatModel(selectedModelPath:)`, which hot-loads the chat model on the first turn when the running server was launched headless. Auto-start was simply the one caller that never asked for it.

## The change

**Auto-start now starts the server headless.** "Auto-start on launch" stays exactly where it is in the tray, keeps its name, and keeps meaning *start the server* — which is now all it does. It gained a tooltip pointing at where loading went.

**Loading became its own decision**, as three rows at the top of Settings ▸ Server:

| Row | Control |
|---|---|
| **Load a model at start** | Toggle, **default OFF**. |
| **Which model** | Two-state segmented control: *Last model used* / *Always this model*. |
| **Model** | Dropdown of installed chat models. |

"Which model" is a property of the selector, not a member of the list — the dropdown contains **only real models**, never a synthetic row. Under *Last model used* the dropdown is disabled and a quiet caption underneath names the model the rule resolves to today ("Currently: Qwen3-8B", or "Nothing loaded yet — the server will start with no model"), so the answer is visible now rather than discovered at the next login.

### The stored preference is a mode plus an optional pinned path

Not a magic path value. `StartupModelChoice.Mode` (`.lastUsed` / `.pinned`) has its own key, beside `startupModelPinnedPath`, which is empty when nothing has been pinned — an absent optional, not a sentinel. A sentinel inside a path field is the same defect in different clothes: every reader has to know the secret string, and one that doesn't treats it as a filename. There is a test pinning that no mode's raw value can be mistaken for a path.

The gate is now a pure function rather than an `if` in `init`, because it is a branch nobody can watch run and its previous shape shipped a multi-gigabyte load behind a checkbox that promised a server:

```swift
static func resolved(mode: Mode,
                     pinnedPath: String?,
                     lastUsed: String?,
                     installedPaths: [String]) -> String? {
    let wanted: String?
    switch mode {
    case .lastUsed: wanted = lastUsed
    case .pinned:   wanted = pinnedPath
    }
    guard let wanted, !wanted.isEmpty, installedPaths.contains(wanted) else { return nil }
    return wanted
}

static func launch(autoStart: Bool, loadModelAtStart: Bool, mode: Mode,
                   pinnedPath: String?, lastUsed: String?,
                   installedPaths: [String]) -> Launch {
    guard autoStart else { return .doNothing }
    guard loadModelAtStart else { return .headless }
    guard let path = resolved(mode: mode, pinnedPath: pinnedPath,
                              lastUsed: lastUsed,
                              installedPaths: installedPaths) else { return .headless }
    return .load(path: path)
}
```

The Settings caption and the launch gate both call `resolved`, so the readout cannot promise a model the gate would decline to load.

### The load toggle defaults OFF, and there is deliberately no migration

An upgrading user whose "Auto-start on launch" used to eagerly load 26 GB **stops doing that on the next launch**. That is the point of the change, not a side effect of it. `loadModelAtStart` is read with `UserDefaults.standard.bool(forKey:)`, so an absent key is false, and nothing writes a true on their behalf.

They lose nothing: the server still comes up, and their model still loads — on the first message, through the `ensureDefaultChatModel` path that already exists.

### Nothing to load is a normal outcome, not an error

No recorded last model (fresh install), or a model uninstalled since it was chosen: the server starts **headless**. It never falls back to an arbitrary model — a startup that loads a model the user never chose is worse than one that loads none — and never sends `--model <gone>`, an instant FileNotFound. The stored pick is checked against the installed chat-pickable library on every launch, so removal is caught here rather than discovered by the server.

### Where "last used" comes from

**I checked for an existing store first and did not find one.** `grep -i "lastModel|recentModel|lastUsed|lastLoaded|mostRecent"` over `app/Sources/MLXServe/` returns only `SandboxSessionTabs.mostRecentActive` (terminal sessions) and its caller. `MediaRecents` tracks generated media *files*, not models. So this adds the smallest store I could: one `UserDefaults` key behind two static functions in `StartupModelChoice`. If there is a source of truth I missed, say so and I will delete mine and read yours instead.

**It is written only where a load is confirmed, never where one is requested** — a load that failed is not a model that was used, and recording it would replay the failure next launch. Three sites, all in `ServerManager`:

| Site | Why it is a confirmed load |
|---|---|
| `transitionToRunning()` | An eager `--model` launch that reached `.running`. A start that dies mid-load never gets here. Headless leaves `currentModelPath` empty, which `recordLoaded` ignores. |
| `loadModel(...)`, inside the existing `if setDefault, id.hasPrefix("/")` branch | A completed chat-model hot-switch — `api.loadModel` returned. **Only** the `setDefault` branch records: `loadModel` is also the media path (`prepareGenModel` hot-loads FLUX / LTX / Kokoro by directory), and a video model is not a "last model used" that a chat startup should reach for. |
| `ensureDefaultChatModel(...)` on success | The first-turn hot-load onto a headless server. This is now the *common* way a model goes resident, so without it "Last model used" would stay empty for anyone who never touches the tray picker. |

`recordLoaded` takes absolute paths only. A registry id is a directory basename (for a Hugging Face snapshot, a commit hash) and a LAN id names another Mac's model; neither can be handed back to `--model`.

## `selectedModelPath` is untouched

Its key, its value, and its `didSet` are exactly as they were. Nobody's chosen model is lost.

The startup pin is a separate key, not a reuse of `selectedModelPath`, because `selectedModelPath` is the model answering chats *right now* — its `didSet` hot-switches or restarts a running server, and editing a startup preference must not swap the model out from under a conversation in progress.

## How the dropdown is populated

It reuses the existing list source; no new scanner. `AppState.localModels` (populated by `DownloadManager.discoverLocalModels()`), filtered by `LocalModel.isChatPickable` — the same filter the tray picker, the composer pill, the ⌘L palette and `AppState.refreshModels` all use, which excludes drafters, media models, `bert` encoders and the NSFW classifier. Labels use `displayLabel` with `LocalModel.duplicateNames` engine-suffixing, copied from `StatusMenuView.modelPicker`, because sibling GGUF quants share a `name` and a macOS `.menu` Picker keys its checkmark by item title.

A pin that is no longer installed keeps a "— no longer installed" row so the Picker never matches no row and renders blank. That row is still a *model* (one that used to be there), not a rule.

## Files

| File | Change |
|---|---|
| `app/Sources/MLXServe/Models/StartupModelChoice.swift` | **New.** The mode, the pure resolution + launch decision, and the last-used store. |
| `app/Sources/MLXServe/AppState.swift` | `loadModelAtStart`, `startupModelMode`, `startupModelPinnedPath`; launch gate switched to `StartupModelChoice.launch`. |
| `app/Sources/MLXServe/Services/ServerManager.swift` | Three `recordLoaded` calls at confirmed-load sites. |
| `app/Sources/MLXServe/Views/SettingsView.swift` | The three new rows at the top of `ServerSectionContent`. |
| `app/Sources/MLXServe/Views/StatusMenuView.swift` | Tooltip on the existing auto-start toggle. |
| `app/Tests/MLXCoreTests/StartupModelChoiceTests.swift` | **New.** 17 cases over the mode, the gate and the store. |

## What a reviewer should check first

Beyond "does it compile", which I genuinely do not know:

1. **`transitionToRunning()` is called from more than one place** (`forceRunning()` is the `TestServer` path). Recording there assumes reaching `.running` with a non-empty `currentModelPath` means the checkpoint really loaded. I believe that holds because `Scheduler.init` blocks before the listener binds, so health cannot answer until the load finished — but that is the inference most worth a second opinion.
2. **`ensureServerForLan()` is now usually a no-op at launch**, since auto-start has already brought the server up headless. That looks like an improvement (it removes a race) rather than a regression, but it is a behaviour change I did not set out to make.
3. Whether the three rows belong in **Settings ▸ Server** or somewhere else. `.server` was the closest fit; there is no "General".
4. Whether recording only the `setDefault` branch of `loadModel` misses a chat-load path I did not find.

### DIFF
--- a/app/Sources/MLXServe/AppState.swift
+++ b/app/Sources/MLXServe/AppState.swift
@@ -191,6 +191,30 @@ class AppState: ObservableObject {
     @Published var autoStartServer: Bool {
         didSet { UserDefaults.standard.set(autoStartServer, forKey: "autoStartServer") }
     }
+    /// Should the launch gate pass `--model` (an EAGER, BLOCKING load) instead
+    /// of starting headless? Separate from `autoStartServer` on purpose:
+    /// "run the server" and "resident a checkpoint" are different decisions,
+    /// and folding them into one checkbox made "Auto-start on launch" read tens
+    /// of gigabytes off disk at login with nothing in the UI saying so
+    /// (issue #214). Default OFF, including for existing users — that IS the
+    /// fix. The server still comes up, and the first chat turn hot-loads the
+    /// selected model via `ServerManager.ensureDefaultChatModel`.
+    @Published var loadModelAtStart: Bool {
+        didSet { UserDefaults.standard.set(loadModelAtStart, forKey: "loadModelAtStart") }
+    }
+    /// WHICH model `loadModelAtStart` loads: `StartupModelChoice.lastUsedTag`
+    /// (the default — "Last model used", resolved at start) or a model's
+    /// absolute path.
+    ///
+    /// Deliberately NOT `selectedModelPath`. That property is the model
+    /// answering chats right now, so its `didSet` hot-switches or restarts a
+    /// running server — editing a *startup* preference must not swap the model
+    /// under a conversation in progress, and it has no way to spell "whichever
+    /// one I used last" besides. `selectedModelPath` and its stored key are
+    /// left exactly as they were.
+    @Published var startupModelPath: String {
+        didSet { UserDefaults.standard.set(startupModelPath, forKey: "startupModelPath") }
+    }
     /// All server-launch flags + per-request defaults, mirrored to UserDefaults.
     /// Auto-saves on every mutation. Prefer this over the legacy single-key
     /// `maxTokens`/`contextSize` defaults — those forward into here.
@@ -426,12 +450,18 @@ class AppState: ObservableObject {
         // Defaults to ON when the key is absent — `UserDefaults.bool` would
         // read a never-set key as false, which is why a fresh install used to
         // download a model and then sit there with the server stopped. The
-        // launch gate below is `autoStartServer && !selectedModelPath.isEmpty`,
-        // so this stays a no-op until a model exists; the first download's
-        // completion hook is what actually starts it. No migration: existing
-        // users who never touched the toggle get it turned on, which is the
-        // intent.
+        // launch gate below now starts the server HEADLESS on its own, so this
+        // no longer waits for a model to exist; what it starts is a server, not
+        // a load. No migration: existing users who never touched the toggle get
+        // it turned on, which is the intent.
         self.autoStartServer = UserDefaults.standard.object(forKey: "autoStartServer") as? Bool ?? true
+        // Both default OFF / "Last model used" via the absent-key reads, and
+        // deliberately WITHOUT a migration from the old behaviour: an upgrading
+        // user whose "Auto-start on launch" used to eagerly load 26 GB must stop
+        // doing that on the next launch. That is the whole point of the change.
+        self.loadModelAtStart = UserDefaults.standard.bool(forKey: "loadModelAtStart")
+        self.startupModelPath = UserDefaults.standard.string(forKey: "startupModelPath")
+            ?? StartupModelChoice.lastUsedTag
         self.selectedModelPath = UserDefaults.standard.string(forKey: "selectedModelPath") ?? ""
         // Load ServerOptions, then migrate legacy single-key defaults
         // (`maxTokens`, `contextSize`) into it on first run if the dedicated
@@ -529,9 +559,24 @@ class AppState: ObservableObject {
             }
         }
 
-        // Auto-start server if enabled and a model is available
-        if autoStartServer, !selectedModelPath.isEmpty {
-            server.start(modelPath: selectedModelPath, options: serverOptions)
+        // Auto-start. "Start the server" and "load a model" are two decisions
+        // (`StartupModelChoice.launch`): auto-start on its own brings the server
+        // up HEADLESS, and only Settings ▸ Server ▸ "Load a model at start" —
+        // default OFF — makes launch pay for a checkpoint. `refreshModels()`
+        // above is what makes the installed-library check below meaningful.
+        switch StartupModelChoice.launch(
+            autoStart: autoStartServer,
+            loadModelAtStart: loadModelAtStart,
+            choice: startupModelPath,
+            lastUsed: StartupModelChoice.lastUsed(),
+            installedPaths: localModels.filter(\.isChatPickable).map(\.path)
+        ) {
+        case .doNothing:
+            break
+        case .headless:
+            server.startHeadless(modelsDir: ServerManager.modelsRoot, options: serverOptions)
+        case .load(let path):
+            server.start(modelPath: path, options: serverOptions)
         }
         // LAN sharing/discovery lives in the server process — with either
         // enabled the server should be up (headless when nothing was
--- a/app/Sources/MLXServe/AppState.swift
+++ b/app/Sources/MLXServe/AppState.swift
@@ -202,18 +202,37 @@ class AppState: ObservableObject {
     @Published var loadModelAtStart: Bool {
         didSet { UserDefaults.standard.set(loadModelAtStart, forKey: "loadModelAtStart") }
     }
-    /// WHICH model `loadModelAtStart` loads: `StartupModelChoice.lastUsedTag`
-    /// (the default — "Last model used", resolved at start) or a model's
-    /// absolute path.
+    /// WHICH model `loadModelAtStart` loads — follow the last one used, or
+    /// always `startupModelPinnedPath`. Stored as its own key rather than as a
+    /// magic value inside a path field, so no reader has to know a secret
+    /// string to tell a rule from a filename.
+    @Published var startupModelMode: StartupModelChoice.Mode {
+        didSet {
+            UserDefaults.standard.set(startupModelMode.rawValue, forKey: "startupModelMode")
+            // Switching to "Always this model" having never pinned one would
+            // leave the dropdown matching no row and rendering blank — the
+            // dead-control class. Seed it with the answer the other mode was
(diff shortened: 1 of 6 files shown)


ddalcu:
merged via @justinluque 's pr.

PR #216, opened 2026-08-17
feat: render markdown tables as a real grid with 80% content width

slava-kudzinau:
## Summary

Markdown tables in the chat view now render as a real grid — bold header row, one divider, no vertical borders — with inline markdown (bold, code, links) rendered inside cells instead of shown as literal characters or flattened into monospaced ASCII-art text. Columns size to their natural content width and only scale down (wrapping instead of overflowing) when a row's combined content exceeds the reading column, so a short "Tip" column stays tight instead of stretching to match a long "Explanation" one. Separately, the chat's reading column (transcript, composer, empty-state greeting) now grows with the window — 80% of the panel's width — instead of stopping at a fixed 740pt cap that wasted 40-50% of a wide panel.

## Implementation

- `MarkdownTable` (new): shared parser for GFM pipe tables and the whitespace-aligned "pseudo-tables" smaller models emit without GFM syntax. Used by both `MarkdownSegmenter` (splits a table into its own segment, like a fenced code block) and `MarkdownTableView`.
- `MarkdownTable.layout(natural:available:)`: pure function that sizes columns. Each column keeps its natural width — measured via `NSAttributedString.size()`, the same rendering used to draw the cell — as long as the row fits; only when the total overflows the available width does every column scale down by the same factor, preserving relative shares. Covered by 6 unit tests (identity, overflow-scale-preserves-ratio, empty, single-column fit, single-column overflow, zero-available guard).
- `MarkdownTableView` (new): plain `VStack`/`HStack` grid (not `Grid`, so proportional widths can be measured once via `onGeometryChange`), with column widths computed once per render and threaded through the header row, the divider, and every data row. Cells are an `NSTextView`-backed `NSViewRepresentable` (`MarkdownTableCell`) rather than SwiftUI `Text`, since `Text` doesn't wrap `AttributedString` content; column alignment is baked into the attributed string as a paragraph style before it's set, and the view's cached intrinsic height is invalidated whenever the wrap width changes (not just on text changes) so a cell that needs more lines as columns resize doesn't report a stale, too-short height to the row below it. Ragged rows (fewer cells than headers) render missing cells as empty.
- 8pt trailing padding inside each cell so a wrapped cell's last word doesn't visually run into the next column's first word.
- `ChatMetrics.contentWidthFraction`: the transcript, composer, and empty-state greeting all size to 80% of the detail column's width instead of capping at a fixed 740pt.
- Inter-segment spacing in `MarkdownText` reduced from 10pt to 4pt, matching the table's own internal row padding, so the gap between a prose paragraph and a following table doesn't read as a jump.

## Verification

- `swift test --filter MarkdownTableTests --filter ChatColumnMetricsTests --filter MarkdownSegmenterTests` — all pass.
- Rebuilt and relaunched the app against live table generations (short and 25+ row tables): no header/row overlap, ragged/streaming-partial rows render cleanly, alignment and inline markdown (links, code spans, bold) render correctly, natural per-column widths hold under window resize.
- Sampled CPU with `top -l 2`/`ps` during an active table 
generation: bursty up to ~60-76% while new rows are laid out, settling back to ~0% once generation completes — consistent with normal per-token layout work, not a runaway loop.
- Screenshot: attaching separately via GitHub.
pr_table_screenshot


ddalcu:
Thanks for this PR, a few things here that we should fix/check on:

1.  docs/superpowers/plans/ -> Please remove
2. Minor: Selection tradeoff,  tables used to live inside the one NSTextView per prose run, so drag-selection crossed them; now each cell is its own NSTextView and .textSelection(.enabled) on the VStack doesn't give you a cross-cell table copy. Probably an acceptable trade for a real grid, but it's a regression and we should see if there is a way to fix this.
3. Medium/Big: Streaming cost: naturalColumnWidths re-renders every cell's attributed string (including LaTeX segmentation) on every body evaluation, which during streaming is every token.  Measurement shows 60-76% CPU bursts on a 25-row table. Works, but a cache keyed on the cell strings would be cheap insurance. I had to fix the CPU spikes recently due to code blocks and causing slower computers to freeze. Test this with a really fast model like LFM 2.5 3B.

slava-kudzinau:
Fixed #1 (removed the planning doc) and #2 (selection regression) in `3b29bb2`.

**#2 fix:** tables now render as an `NSTextTable` inside the same `NSAttributedString`/`NSTextView` as the surrounding prose, instead of a separate SwiftUI grid with one `NSTextView` per cell. Drag-selection and copy now span cell-to-cell and prose-to-table again, same as before this PR. Verified live: dragging from inside a table cell through the header row and into a following prose paragraph/list produces one continuous native selection, and Cmd+C copies clean tab-free text (header + rows, correct order).

Trade-off worth flagging: `NSTextTable` has no "tight to content" column-sizing mode (only absolute-point or percentage-of-container widths), so a table now always fills the reading column's width, instead of staying tight and only scaling down on overflow like before. `MarkdownTable.columnFractions` still weights columns by content length so a short "Tip" column doesn't get the same share as a long "Explanation" one, but a 2-column short table will now stretch full-width rather than staying compact. Let me know if that trade is unacceptable and I'll look at a resize-aware absolute-width variant instead (bigger change - needs re-attributing the table's paragraph styles on container-width change, since `NSTextTable` can't do "content width, but reflow on resize" natively).

#3 (streaming CPU cache) not yet addressed - want me to pick that up next, or is this enough for this round?

slava-kudzinau:
Fixed #3 in `23760d2`: added `cellRenderCache` (`NSCache`, same pattern as the existing `renderCache`) around `renderInline` for table-cell content, keyed on cell text + theme + bold. Completed rows/cells now reuse their prior render instead of repeating the markdown/LaTeX-segmentation work on every token - only the actively-growing last cell does real work during streaming.

Measured before/after with a synthetic 25-row x 3-column table, simulating one `attributedString(for:)` call per streamed chunk (61 calls, matching the cache-miss-per-token pattern `attributedString(for:theme:)`'s outer cache already has during streaming):

| | avg ms/call |
|---|---|
| table, before this fix | 3.5 |
| table, after this fix | 0.88 |
| equivalent-length plain prose (no table) | 2.4 |

~4x faster, and now cheaper than equivalent plain prose. For context, this NSTextTable rewrite's baseline cost (before the cache) was already far below the 60-76% CPU bursts you measured on the old SwiftUI-per-cell architecture - that number was dominated by `naturalColumnWidths` running unguarded on every SwiftUI body evaluation (more frequent than once per token), which no longer exists in this design (`buildAttributedString` only runs on an actual cache miss, and `SelectableMarkdownNSText.updateNSView` already skips redundant `NSTextStorage` replacement via `isEqual`). This cache closes the remaining table-specific delta on top of that.

All 38 tests pass, full app rebuilt and relaunched clean.

PR #218, opened 2026-08-18
Smart resolution picker and 512x512 Flux preset

justinluque:
- Add 512x512 FLUX preset 
- Custom resolution picker for image and video
- The server has always accepted 512 on FLUX (clampFluxDim takes any multiple of 32 from 256 up), so the square preset was a missing menu row rather than a missing capability.
- Custom sizes get a ResolutionGrid per backend, mirroring gen.zig's clampFluxDim and clampKreaDim. Off-grid values are corrected with a hint naming the size used and the step; out-of-range values are refused, since silently clamping 4000 to 1536 returns a picture a third of the requested size with nothing explaining why.
- The grid snaps UP, matching the server. Rounding to nearest would print a hint naming a size the server never generates (481 -> 480 against the server's 512).


ddalcu:
popular request ! thank you !

PR #220, opened 2026-08-18
Take the checkpoint's own thinking default when a request names none

### LINKED ISSUE #219: Thinking is off by default; every other runner lets the model's template decide
mlx-serve always writes an explicit `enable_thinking` into the chat-template render context, so a checkpoint whose template defaults to thinking-on is silently switched off unless its arch is on a hardcoded allowlist.

Where it happens (`main` @ 95c1a46):

- `src/chat.zig:1056` — `serializeExtraContext` always emits `"enable_thinking":true|false`, so the template's own default branch never runs.
- `src/server.zig:4662` — `resolveEnableThinking`: a request naming neither `enable_thinking` nor `reasoning_effort` falls through to `arch_default`.
- `src/model.zig:875` — `defaultEnableThinking` is `false` for every arch except `muse_glimmer` (with tools) and `bailing_hybrid`.
- `generation_config.json` is parsed for sampling only (`src/model.zig:1214`); `default_chat_template_kwargs` is never read.

Measured 2026-08-17, `mlx-community/Agents-A1-8bit`, identical prompt:

- mlx-serve, nothing sent: 3 output tokens, 0 characters of reasoning.
- mlx-serve, `enable_thinking: true` sent: 1,812 output tokens, 5,689 characters of reasoning.
- LM Studio, same weights, nothing sent: 992 output tokens, reasoning present.

| Runner | Default | How it decides |
|---|---|---|
| LM Studio | on | Per-model, reported as `capabilities.reasoning.default` |
| Ollama | on | Absent `think` means "model default", which is on for thinking models |
| vLLM | on | Obeys the model's chat template |
| llama.cpp | on | Obeys the model's chat template |
| mlx-serve | **off** | Overrides the template |

Consequence: a benchmark or comparison run against mlx-serve silently measures a non-thinking model.

Ask: when a request sends no `enable_thinking` and no `reasoning_effort`, use the model's own default — `default_chat_template_kwargs.enable_thinking`, or the template's default — instead of an allowlist. An explicit client value keeps winning.

Fe2-O3:
> Written by an agent. Not compiled, not tested. There is no Zig toolchain on this machine and `build.zig.zon` pins a specific dev version, so `zig build` and `zig build test` were never run. Treat this as a suggestion to verify, not a ready patch.

Fixes #219 — a request sending neither `enable_thinking` nor `reasoning_effort` falls through to `ModelConfig.defaultEnableThinking`, an arch allowlist that is `false` for nearly every model, so a thinking model answers without reasoning.

The change, all in `src/model.zig`:

- `parseGenerationDefaultsFromJson` reads `default_chat_template_kwargs.enable_thinking` from `generation_config.json`. Only a real bool counts.
- New `ModelConfig.gen_enable_thinking`, filled in `parseConfig` beside the sampling recommendations.
- `defaultEnableThinking` returns that value when present, then falls through to the existing per-arch allowlist unchanged.

`resolveEnableThinking` is untouched, so an explicit client value still outranks everything.

Two notes:

- I found no server-wide thinking switch to preserve. `--reasoning-budget` caps thinking tokens but is not a default-off, so nothing needed reordering. A global off should sit above this.
- Reading the template's own default would cover more checkpoints but needs a render probe. This is the smaller half.

Tests added for the parser and for the precedence.

### DIFF
--- a/src/model.zig
+++ b/src/model.zig
@@ -410,6 +410,12 @@ pub const ModelConfig = struct {
     gen_top_p: ?f32 = null,
     gen_top_k: ?u32 = null,
 
+    // The checkpoint's OWN thinking default, from generation_config.json's
+    // `default_chat_template_kwargs.enable_thinking`. null = the file or key
+    // is absent. Read by `defaultEnableThinking` for requests that name no
+    // thinking preference; an explicit request value still outranks it.
+    gen_enable_thinking: ?bool = null,
+
     // Gemma 4: explicit layer type map (bit = 1 means full/global attention)
     has_explicit_layer_types: bool = false,
     layer_is_global: [128]bool = @splat(false),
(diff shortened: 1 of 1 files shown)


ddalcu:
Thank you! Built and tested locally, merging.

PR #223, opened 2026-08-19
ANE Prefill boost + DFlash 2  (Need testers, M3Ultra and M5Max)

ddalcu:
Hello Everyone, would like to get some help on testing this branch with some hardware that I don't have.
Feel free to either add results as comments to this PR (preferred), or email me at david.dalcu@gmail.com

If you can please help, please send the following to your agent: 
Mostly just M3 Ultra, but see the note below for all apple silicon also.
```
Please follow the instructions in this PR, download the source code, set it up locally, 
and run the tests, then use gh cli to post the results to the PR as a comment.  
https://github.com/ddalcu/mlx-serve/pull/223
Instructions here: https://github.com/ddalcu/mlx-serve/blob/e53195186781449cfdf796547741746b974956ee/NOTE_TO_TESTER_ANE_DFLASH2.md
```
(if you dont have gh cli, email / x chat / manual comment)

New updated Instructions (August 20 / round 2 / M3 Ultra testing / Light M5):

https://github.com/ddalcu/mlx-serve/blob/e53195186781449cfdf796547741746b974956ee/NOTE_TO_TESTER_ANE_DFLASH2.md

Also, if you have a M1/M2/M3/M4 (any class), it would be good to get a idea of the best MTP Draft depth each silicon can do most efficiently.
Prompt for the agent:
```
You dont need to download the source code, just local install of mlx-serve, and experiment with:
  --mtp-depth <n>     Max tokens drafted per MTP round (default:
                        adaptive — the EV controller plans depth
                        per round up to 8 on eligible M5 NAX targets,
                        otherwise 6; MLX_SERVE_MTP_ADAPTIVE=0
                        reverts to the fixed windowed controller,
                        cap 3). Pass an explicit <n> to hard-cap.
                        
 Using ddalcu/Qwen3.8-27B-MLX-Serve-4bit
 
 Report back your specs, and best draft rate, use `npx llmprobe@latest --bench-only`
```

Thanks @maxrochefortshugar and @RexGong 
Also big thanks to https://x.com/Kevrsub for spending all day testing on M3Ultra !


maxrochefortshugar:
## Test 1 on M5 Max: DFlash2 ties MTP

Hey, thanks for your work on this project.

Block 8 runs on M5, the wide verify lane engages, and DFlash2 still doesn't pull ahead. Novel cell is 35.2 tok/s for dflash2 against 35.3 for mtp. Echo is 58.5 against 59.6, so mtp keeps about 2%.

Machine is a MacBook Pro M5 Max, 128 GB, macOS 26.5.2 (25F84), Xcode 26.6. I built `feature/ane` @ 8b21708 with `./app/build.sh`, which worked first go with no Apple account. mlx-serve 26.8.10, ReleaseFast, `nax on (M5 neural accelerators)`. Trunk is `ddalcu/Qwen3.8-27B-MLX-Serve-8bit`, drafter `incoai/Qwen3.8-27B-DFlash2`, thinking on, temp 0, `--prefix-cache-entries 0 --skip-mem-preflight`. No llmprobe on these runs, just curl and the `decode:` lines out of the server's own log.

Engagement lines off the dflash2 boot, since that's the M5 question:

```
DFlash drafter ready (block_size=8, wide_verify_lane=true, targets={ 5, 19, 33, 47, 61 }).
[vqmm] NAX verify lane engaged at 8-bit: M=8 K=5120 N=10240 gs=64 min_m=7
```

### Numbers

These are means over cycles 3 and 4 of a Latin square rather than medians of a straight run, for reasons in the drift section. Every cell had a `mode=` line.

| arm | novel tok/s | x serial | echo tok/s | x serial | avg_per_round | per_draft |
| :-- | --: | --: | --: | --: | --: | --: |
| mtp | 35.3 | 1.92x | 59.6 | 3.27x | 1.82 | 49.7% |
| dflash2 (blk 8) | 35.2 | 1.91x | 58.5 | 3.22x | 2.49 | 35.5% |
| dflash2-v1 (blk 8) | 30.5 | 1.65x | 57.6 | 3.17x | 2.06 | 29.4% |
| serial | 18.4 | 1.00x | 18.2 | 1.00x | - | - |

Your M4 numbers were mtp 65.7 against 64.6, so M5 closes the gap to nothing on novel instead of flipping it, which I realise isn't the confirm-or-kill you were after.

Acceptance at matched width, same prompt, greedy, thinking on:

| arm | drafts/round | accepted/round | per-draft hit |
| :-- | --: | --: | --: |
| DFlash2, block 8 | 7.0 (always) | 2.49 | 35.5% |
| MTP, forced depth 7 | 3.07 avg | 1.55 | 50.6% |

That's 1.61x more accepted tokens per verification step against your 1.49x on M4, so the drafter side of your argument holds up here. MTP self-limits harder on my box than yours: `drafted=722 attempts=235` is 3.07 per round where you saw 3.9, with `ext_rounds=0` under `MLX_SERVE_MTP_ADAPTIVE=0`.

Selector off costs 17% acceptance (2.49 to 2.06) and 13% novel decode (35.2 to 30.5), same direction as your +16%.

I couldn't separate `--draft-block-size` 6 or 7 from 8. Same-config block 8 measured 26.8 then 31.1 tok/s across a single A-B-A, and a 16% swing brackets everything block 7 did, so don't read a block optimum off my numbers.

### One thing for other testers

I couldn't get an MTP head to bind from the 4-bit pack. `ddalcu/Qwen3.8-27B-MLX-Serve-4bit` carries 29 `language_model.mtp.*` keys in its index, but boots with no `[mtp]` line and emits no `[spec-stats]` even under `--mtp --no-pld`, so it decodes serially. The 8-bit pack is identical in config apart from `quantization.bits` and binds it happily:

```
[mtp] loading in-checkpoint head from the trunk shards
[mtp] loaded native MTP head (dense-mlp; per-weight quant, fallback bits=8/gs=64)
MTP head ready (depth=8, profile=g17_nax_q8_gs64).
```

Could well be something local to me. It's why everything above is on the 8-bit pack, and I can send the whole 4-bit boot log if it's worth a look.

The visible symptom is a quiet fallback rather than an error. With PLD on (the default) the 4-bit pack answers `mode=pld`, which looks like a working spec arm until you read the mode; with `--no-pld` it just decodes serially. Trap 2 is what caught it for me, since the launch flags looked right either way. The 8-bit pack engages MTP on a plain default boot with no `--mtp`, so the note's `mtp: default boot` arm is correct as written.

### Drift

Worth a warning for anyone testing on a laptop. Decode roughly doubles between battery and AC on this machine and keeps climbing while the battery charges. Same arm, same build, identical acceptance (`avg_per_round=2.49 per_draft_pct=35.5%` both times): 13.9 tok/s on battery, 25.4 to 31.1 as it charged. It gave me a very convincing "block 7 is twice block 8" result that vanished under A-B-A.

Forward-then-reverse wasn't enough for a ramp that size, so I moved to a 4x4 Latin square, each arm in each time slot, 16 boots. Cycles 1 and 2 still carry the ramp, with mtp minus dflash2 reading +7.0% then +1.6% on novel and decaying to 0.0% and +0.6% by cycles 3 and 4. Ratios to serial held across the whole run including the battery/AC boundary, which is the only reason I trust the tie at all. Acceptance never moved, only tok/s.

"Run on AC at high charge" might deserve a line next to "keep everything else off the GPU".

All three traps in the note are real, and checking `mode=` instead of the flags is what stopped me posting PLD numbers with an mtp label on them. Raw spec-stats, per-cell CSV and per-boot logs are yours if any of it helps. Happy to run Test 2 on this box, and the Muse round if you want it.

*Edited: an earlier version of this comment claimed default boot gives PLD rather than MTP on this MoE trunk generally. That was wrong. It was the 4-bit pack's unbound head, not a MoE default, and I've corrected the section above. Default boot on the 8-bit pack engages MTP correctly.*

RexGong:
## Local M5 Max validation — Qwen3.8-27B only

Tested commit `8b217082050954ec6490c27b997b7841dcc64b8a` (`feature/ane`) using only:

- target: `ddalcu/Qwen3.8-27B-MLX-Serve-4bit` (17 GB, native `language_model.mtp.*` head)
- assistant: `incoai/Qwen3.8-27B-DFlash2` (selector + dynamic convs, trained block 8)
- MacBook Pro, Apple M5 Max (18 cores), 128 GB, macOS 26.5.2 (25F84)
- Xcode 26.6 (17F113), Metal Toolchain 17F109 / Metal 32023.883
- local binary: mlx-serve 26.8.10, MLX 0.32.0, mlx-c `v0.6.0-4-gfba4470`, `nax on`
- llmprobe 0.4.4

### Build and tests

- `brew bundle`: pass
- `app/build.sh`: pass; self-built MLX staged with 3,798 NAX symbol hits and macOS min version 26.2; app/DMG built and signed
- `swift test`: **2,973 tests, 24 skipped, 0 failures**
- `.zig-toolchain/zig build test`: **1,645/1,783 passed, 136 skipped, 2 failed**
  - `deepseek_v4.test.dsv4: fused decode-chain kernel matches the composed q/kv/idx/o chains (GPU)` — `src/deepseek_v4.zig:9775`, composed `-0.5`, kernel `-0.75`
  - `minimax_h3.test.minimax h3: sparseAttend attends exactly its subset (uniform-score closed form)` — `src/minimax_h3.zig:6228`, actual `1.9995117`, expected `2`, tolerance `1e-5`
  - both failures are outside the Qwen/DFlash/ANE paths, but recorded here rather than hidden
- changed shell/Python syntax checks: pass
- `git diff --check origin/main...HEAD`: reports trailing whitespace at `docs/gotchas/engine-mlx.md:2719`
- MTP live equivalence/engagement suite: **13/13 pass**, including M5/NAX auto profile `g17_nax_q4_gs64`, auto depth 8, `ext_rounds=11`, and the adaptive-off depth-3 kill switch
- ANE live contract suite: **pass** (channel + MLP/GDN engagement, dead staging reap, GDN-off, row mode, and zero `[ane]` lines in the off arm)

The live scripts currently poll `/health`, which answers before a 17 GB model is ready. I ran readiness-only temporary copies which wait for `Server listening` (and set `NO_PROXY` for localhost); no assertions were changed.

The DFlash live suite then ran **13/14**. The only failing assertion is a harness/checkpoint mismatch: it requires `block_size=16` whenever `wide_verify_lane=true`, while this Qwen DFlash2 checkpoint declares/trains block 8. The runtime correctly logged `DFlash drafter ready (block_size=8, wide_verify_lane=true)` and `[vqmm] NAX verify lane engaged ... M=8`. All other assertions passed: DFlash2 module load, engagement/gating, 4-bit tie-aware reproducibility, per-request opt-out, non-stream/stream clipping, tool call parsing, assistant quantization, trunk lm_head, and prefix-cache assistant-context restore.

Sanitized server logs from one forward boot per arm are included with this report:

- `dflash2-selector-b8.log`
- `mtp-forced-d7.log`
- `dflash2-selector-off-b8.log`
- `serial.log`

Local user-directory prefixes in these attachments were replaced with `$HOME`; no performance or engagement lines were removed.

### Decode benchmark

Conditions: prefix cache off, temperature 0, one 600-token novel-prose cell and one 1,200-token exact-echo cell, 3 reps/cell/boot, forward and reverse arm order (6 measurements per table cell). MTP was `--mtp-depth 7` with `MLX_SERVE_MTP_ADAPTIVE=0`. Values are median tok/s; brackets are forward/reverse medians.

| arm | novel tok/s [F/R] | novel acceptance | echo tok/s [F/R] | echo acceptance |
|---|---:|---:|---:|---:|
| DFlash2 selector, block 8 | **26.5** [31.0/21.3] | 1.52/round, 21.7%/draft | **78.6** [92.1/66.3] | 7.00/round, 100% |
| MTP fixed depth 7 | **32.9** [38.3/26.1] | 0.86/round, 54.0%/draft | **69.6** [82.7/59.7] | 5.98/round, 99.2% |
| DFlash2 selector off, block 8 | **20.4** [21.8/19.8] | 1.26/round, 17.9%/draft | **65.9** [63.7/67.9] | 7.00/round, 100% |
| serial | **19.5** [18.9/20.1] | — | **19.8** [19.6/19.8] | — |

Representative engagement lines from the forward boot, one per arm/cell:

- DFlash2 selector, block 8 — novel:
  `[spec-stats] mode=dflash attempts=238 accepts=362 avg_per_round=1.52 gate_min=0.93 per_draft_pct=21.7% block_size=8 runtime_disabled=false`
- DFlash2 selector, block 8 — echo:
  `[spec-stats] mode=dflash attempts=150 accepts=1050 avg_per_round=7.00 gate_min=0.93 per_draft_pct=100.0% block_size=8 runtime_disabled=false`
- MTP forced depth 7 — novel:
  `[spec-stats] mode=mtp attempts=322 accepts=278 avg_per_round=0.86 per_draft_pct=54.0% depth=7 drafted=515 ext_rounds=0 runtime_disabled=false`
  - `drafted/attempts = 515/322 = 1.60`
- MTP forced depth 7 — echo:
  `[spec-stats] mode=mtp attempts=172 accepts=1028 avg_per_round=5.98 per_draft_pct=99.2% depth=7 drafted=1036 ext_rounds=0 runtime_disabled=false`
  - `drafted/attempts = 1036/172 = 6.02`
- DFlash2 selector off, block 8 — novel:
  `[spec-stats] mode=dflash attempts=266 accepts=334 avg_per_round=1.26 gate_min=0.93 per_draft_pct=17.9% block_size=8 runtime_disabled=false`
- DFlash2 selector off, block 8 — echo:
  `[spec-stats] mode=dflash attempts=150 accepts=1050 avg_per_round=7.00 gate_min=0.93 per_draft_pct=100.0% block_size=8 runtime_disabled=false`
- Serial — novel and echo: no `[spec-stats]` line, as expected with MTP, DFlash, drafter, and PLD disabled.

All six repetitions inside each arm had the same output hash. Echo was identical across all four arms; novel outputs were arm-stable but diverged across arms after 4-bit near-ties. Thermal drift was large during the sustained run, hence the forward/reverse numbers are shown explicitly.

Because block 8 lost to MTP on novel prose, I also tried 6/7 as requested (single sequential boot per width, 3 reps/cell):

| DFlash2 block | novel median | echo median | acceptance (novel / echo) |
|---:|---:|---:|---:|
| 6 | **31.7** | **73.8** | 1.26 / 5.00 per round |
| 7 | **20.9** | **37.6** | 1.30 / 5.98 per round |
| 8 (counterbalanced result above) | **26.5** | **78.6** | 1.52 / 7.00 per round |

Block 6 improved novel throughput to near MTP, while block 8 remained best for saturated echo. The 6/7 follow-up was not counterbalanced and was thermally affected, so I would not use the block-7 absolute number as a stable machine score.

### ANE prefill

Direct API prompts were calibrated with this model's `/tokenize` endpoint to 8,262 / 16,462 / 32,862 actual prompt tokens. Prefix cache and all speculative decode were off. Every enabled arm reported full 64/64 MLP coverage; GDN-on arms reported 48 GDN layers. There were zero compile/eval failures and no partial-coverage boots.

Default `/props.ane`:

```json
{"mode":"channel","mlp_layers":64,"gdn_layers":48,"rows":8192,"chunk_rows":8192,"share":0.45,"int8_bytes":9468641280}
```

Requested ON/OFF/ON/OFF sequence (prompt tok/s):

| arm | 8k | 16k | 32k |
|---|---:|---:|---:|
| ON 1, channel 0.45 | 683.9 | 473.9 | 372.7 |
| OFF 1 | 584.9 | 486.4 | 388.9 |
| ON 2, channel 0.45 | 404.4 | 356.7 | 350.5 |
| OFF 2 | 485.5 | 451.6 | 393.0 |
| median ON | 544.2 | 415.3 | 361.6 |
| median OFF | 535.2 | 469.0 | 391.0 |
| median delta | +1.7% | -11.4% | -7.5% |

The sustained run showed strong thermal/order drift (for example ON 8k 683.9 → 404.4), so the default 0.45 result is not a clean positive on this M5 Max. Engagement was nevertheless verified on both ON boots.

Sequential split/mode attribution sweep (prompt tok/s):

| arm | `/props` ANE bytes | 8k | 16k | 32k |
|---|---:|---:|---:|---:|
| channel 0.30 | 6.23 GB | 534.6 | 479.1 | **417.4** |
| channel 0.40 | 8.40 GB | 563.1 | 469.5 | 372.7 |
| channel 0.45 | 9.47 GB | 463.1 | 408.0 | 357.8 |
| channel 0.50 | 10.57 GB | 555.5 | 450.5 | 361.8 |
| row default 0.40 | 21.14 GB | **662.3** | **488.1** | 385.1 |
| channel 0.45, GDN=0 | 7.68 GB | 489.6 | 412.4 | 321.3 |

For the longest cell, channel 0.30 was best and used much less resident ANE data than row mode. GDN-off kept all 64 MLP layers engaged, reported `gdn_layers=0`, emitted no GDN engagement line, and was slower at all three lengths in its position. As with the decode run, treat small differences cautiously because this was a long sequential sweep with visible thermal recovery during compile intervals.

Overall: build and Qwen-specific MTP/ANE live paths are functional on M5 Max/NAX. DFlash2 selector improves novel acceptance over selector-off (1.52 vs 1.26/round), MTP depth 7 wins novel throughput, DFlash2 block 8 wins repetitive echo, and the default ANE channel share 0.45 is not the best long-context point on this machine (0.30 won the 32k sweep).
[pr223-qwen-m5max-server-logs.zip](https://github.com/user-attachments/files/31243678/pr223-qwen-m5max-server-logs.zip)

ddalcu:
Thanks all, I added a small test for M5 Max, at the end of the doc, should be quick, just to make sure it passes the test.  + some more Ultra testing.

RexGong:
> Thanks all, I added a small test for M5 Max, at the end of the doc, should be quick, just to make sure it passes the test. + some more Ultra testing.

Pulled `feature/ane` to `e53195186781449cfdf796547741746b974956ee` and retested on the same Apple M5 Max / 128 GB machine running macOS 26.5.2.

Ran the full pinned suite from a clean worktree:

`./.zig-toolchain/zig build test`

Result: **PASS**, exit code 0.

Confirmation summary:

`Build Summary: 9/9 steps succeeded`

I also ran exact filtered confirmations for the two previously failing M5 GPU tests:

- `dsv4: fused decode-chain kernel matches the composed q/kv/idx/o chains (GPU)` — **PASS**
- `minimax h3: sparseAttend attends exactly its subset (uniform-score closed form)` — **PASS**

Each filtered command completed with:

`Build Summary: 9/9 steps succeeded; 2/2 tests passed`

So both previously reported M5-only GPU unit-test failures are resolved on this machine.

I also confirmed at source level that the live scripts now wait for `Model ready (loaded on inference thread)`, the DFlash block assertion reads the drafter checkpoint’s declared block, and the missing-MTP-head debug log is present. I did not rerun the live model suites in this pass.

PR #247, opened 2026-08-21
feat: gpt-oss

justinluque:
## Summary
- Adds OpenAI gpt-oss support (20B-A3.6B / 120B-A5.1B MoE, harmony chat format): clamped SwiGLU activation, attention sinks (native MLX SDPA sink support), per-expert gate/up/down biases forcing the sorted MoE path, mxfp4 native expert quantization via NAX gather lanes, single shared RoPE theta across sliding/full layers.
- Fixes two silent-failure classes surfaced during bring-up: the reserved-token suppressor was masking `<|constrain|>` (a structural separator, not a generated-only token), causing malformed tool-call headers and repetition loops; `"content": null` was satisfying the harmony template's `"content" in message` key check while failing its value-indexing check on replay, leaking raw `<|channel|>` markers into round-tripped tool results.
- Sizes the gpt-oss sliding-window K/V view through the shared `slidingViewFor` helper rather than the raw window (correctness + the existing sliding-trim perf win).
- Streaming: gpt-oss reasoning channels now parse correctly out of the streamed token buffer, matching the non-streaming split.

## Test plan
- `zig build test -Doptimize=ReleaseFast`: 1681/1817 pass (136 skipped), 0 failures
- `cd app && swift build -c release`: builds clean
- Live smoke test against `isetnefret/gpt-oss-20b-RichardErkhov-heresy-mlx-4Bit`:
  - Plain chat: correct answer, clean `content`/`reasoning_content` split
  - Tool call: parses correctly (`finish_reason: "tool_calls"`, valid JSON args)
  - Tool-result round-trip: clean final `content` with no leaked `<|channel|>` markers, `finish_reason: "stop"`

### DIFF
--- a/app/Sources/MLXServe/Models/HFModels.swift
+++ b/app/Sources/MLXServe/Models/HFModels.swift
@@ -30,6 +30,7 @@ private let supportedArchitectureTagPrefixes: [String] = [
     "muse",       // meta-models Muse-Glimmer-30B — repos tag "muse_glimmer" (and *_text)
     "bailing",    // inclusionAI Ling 3.0 — repos tag "bailing_moe_v3" and/or "bailing_hybrid"
     "ling",       // …and/or the family name
+    "gpt_oss",    // GPT-OSS family tags (e.g. mlx-community/gpt-oss-20b-MXFP4-Q8)
 ]
 
 /// model_type values from config.json that the Zig server can load.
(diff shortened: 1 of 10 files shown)


ddalcu:
we will get this merged for 26.8.11 after prerelease and release of .10 . Thank you.

PR #248, opened 2026-08-21
Implement SeedVR2 upscaling

justinluque:
## SeedVR2 image and video restoration

One-step diffusion restoration (ByteDance, ICLR 2026): sharpens detail and removes
compression artifacts rather than resizing. New `restore` modality, a pane in the
Image window, and two routes — `POST /v1/images/upscales`, `POST /v1/video/upscales`.

image

Six choices: 3B and 7B (standard or ByteDance's `sharp` finetune), each fp16 or 8-bit.

### Notes

- The temporal rope offset keys on `rope_on_text`, **not** the frequency basis — keying it on `pixel` is right for both shipped configs and wrong for a `lang` pack that skips text rope.
- Reference is `filipstrand/mflux` `models/seedvr2/`, which the packs are exported against. Spec in `docs/seedvr2-arch.md`, stories in `docs/gotchas/models-media.md`.
- Only 3 of 6 packs have had weights through them. 7B fp16 and both `sharp` variants are the same architecture through the same path, but unrun.
- "About twice as long" in the 7B preset copy is not measured.
- Video upscaling is wired, I only tried images.
- UI could probably use work.

### Testing

Full suite: 9/9 steps, 1700 passed, 0 failed.


justinluque:
Take a look here and lmk what you think when you get a second @ddalcu  . Took a crack at reworking the UI.

justinluque:
image

drop an image and press upscale or send a previously generated image to be upscaled with one click with the new enlarge badge under the image. plus a strip showing your previous generations and a way to delete generations in app (right click).

image

image

lojza3d:
Hey @justinluque, nice work, looking forward to SeedVR2 in MLX Core.

I'm about to touch the gen panes too: the model picker moves to the top of each form, and later a generated media library shared across Image/Video/Music/Speech, which overlaps with your session strip here. The non-Image parts don't touch this PR, so I'll start there. For the Image pane: is the UI here settled, or still moving? If it's close, I'd hold the Image changes until it lands.

justinluque:
Hey, sorry for the late response! Feel free to go ahead and begin making image pane changes, as the work here is for the most part finished as long as @ddalcu approves of the UI changes. I've worked a bit on adding support for TAESD for the image models but I think I'll put that on a separate branch and make a different pull if I'm not beaten to it.

ddalcu:

Reviewed the seams a restorer shares with everything else rather than the whole 10k: the discovery change generalizes `requiredMediaMarker` without altering any single-marker arch's behavior and the new test pins both spellings; the routes inherit the media body cap through `maxRequestBytesFor`'s prefix rule, so `/v1/images/upscales` and `/v1/video/upscales` are covered automatically; `handleUpscaleVideo` decodes every frame before touching the GPU with geometry and memory refusals up front, and its `frames` cleanup is a proper scope defer. I also swept both big modules for the free-then-write-back handle pattern — every `mlx_array_free` there is a `defer` paired with a write into a fresh handle, which is the correct idiom. One defect.

- `src/gen.zig:937` — `RestoreEngine.load` builds the encoder, decoder and DiT inside a single struct literal with no `errdefer` between them, so a failure in any later component orphans every earlier one plus the `pos_emb` copy made at `:932`; the only `errdefer` (`:894`) frees the `RestoreEngine` allocation itself.
  Failure scenario: a 6.8 GB DiT that OOMs, or an mlx-community pack whose `KeyScheme` misses a tensor, fails at `:939` after `loadEncoder`/`loadDecoder` have already succeeded. Zig evaluates literal fields in order, so the encoder and decoder are fully constructed and unreachable — and they own their arrays outright (`seedvr2_vae.zig:709` and `:877` free every handle in `deinit`, they are not borrowed from the weights map, which its own `defer vw.deinit()` releases). The whole VAE plus `pos` leaks on each attempt, and a load failure is exactly the case a user retries.

ddalcu:
🤖 **MLXServe-PRBot approved** — reviewed at head `39d5541e`, no issues found.

The partial-construction leak is closed the right way: `encoder` and `decoder` are built into locals with their own `errdefer`s and `pos` gets one too, so a failure inside the struct literal (`seedvr2_dit.load`) unwinds all three, and the locals are freed exactly once because the literal never completes on that path.

The regression test earns its keep rather than just existing. It proves the probe is non-vacuous first — load the VAE, assert a descriptor on it, deinit, assert it is gone — then symlinks `pos_emb.safetensors` in as `dit.safetensors` so the DiT load fails at its first tensor, *after* the encoder, decoder and `pos` copy are all built, and asserts no descriptor survives on either file. That is the exact shape the finding described, red on revert.

Rest of the delta is a rebase (the `media_model_types` reflow keeps every entry and adds `seedvr2`) and prose compression in `CLAUDE.md`. Everything I checked last round still holds.

Issue #209, opened 2026-08-17
POST /v1/unload-model returns 200 with a success-shaped body when the model key is unrecognised (nothing is unloaded)


Fe2-O3:
`POST /v1/unload-model` only reads the `model` key. A body that carries a different key (`model_id`, `id`) parses fine, yields an empty id, falls through to the default-model path, and still returns `200` with a success-shaped payload. The client cannot distinguish a real unload from a no-op.

## Observed

Server holding `badtheorylabs/Macaw-4bit-MLX` resident.

```
curl -s -X POST localhost:11234/v1/unload-model \
  -H 'content-type: application/json' \
  -d '{"model":"badtheorylabs/Macaw-4bit-MLX"}'
```
→ `200` `{"model":{"id":"badtheorylabs/Macaw-4bit-MLX","object":"model","loaded":false,"state":"unloaded"}}` — correct, the model is freed.

```
curl -s -X POST localhost:11234/v1/unload-model \
  -H 'content-type: application/json' \
  -d '{"model_id":"badtheorylabs/Macaw-4bit-MLX"}'
```
→ `200` `{"model":{"id":"mlx-serve","object":"model","loaded":false,"state":"unloaded"}}` — nothing was unloaded, the model is still resident.

Same result for `{"id":"badtheorylabs/Macaw-4bit-MLX"}`.

## Expected

An unload body that contains keys but none the handler recognises should be rejected with `400` naming the accepted key, not answered with a `200` that reports a state it did not produce.

## Source

`src/server.zig:3899-3908` — only `"model"` is read; on any other key `requested_id` stays `""`:

```zig
if (parsed.value == .object) {
    if (parsed.value.object.get("model")) |m| {
        if (m == .string) requested_id = m.string;
    }
}
```

`src/model_registry.zig:802-811` — an empty id resolves to `default_id`, so the request is treated as "unload the default model".

`src/scheduler.zig:1824-1829` — if that default entry is already `.unloaded` (or `.error_state`), `unloadModel` returns success without doing anything.

`src/server.zig:3943-3948` — the response then labels itself `"mlx-serve"` and reports `"state":"unloaded"`:

```zig
const id_for_body = if (requested_id.len > 0) requested_id else "mlx-serve";
```

Note the same body shape against a server whose default model *is* resident would unload the default model rather than the one named in `model_id`.

PR #258, opened 2026-08-22
fix: 404 an unknown model id on a registry server, instead of answering it from the default model

### LINKED ISSUE #257: serve mode: a chat request naming a model id that is not served returns 200 and impersonates the requested model while providing an answer from the resident model
**What I expected:** a `/v1/chat/completions` request for a model id that is not in `/v1/models` to fail (404 naming the served ids) — the way the same request fails on the OpenAI API (`model_not_found`) and on other local servers.

**What happened:** it returns `200`, echoes the unknown id back in `"model"`, and the answer comes from whatever model is resident. `/v1/models` lists only the two real ids. Same on `/v1/completions`, `/v1/messages` and `/api/chat` — they share the resolver. This is the default-model fall-through that makes a single `--model` launch a drop-in for clients hardcoding `gpt-4o`, reached from a server that publishes its ids.

**Repro (measured 2026-08-22 01:58):**
```
mlx-serve serve --host 127.0.0.1 --port 11240      # ~/.mlx-serve/models holds two MLX checkpoints
curl -s http://127.0.0.1:11240/v1/chat/completions -H 'Content-Type: application/json' \
  -d '{"model":"no-such-model-xyz","messages":[{"role":"user","content":"say ok"}],"max_tokens":4}' \
  -w '\nHTTP %{http_code}\n'
→ {"model":"no-such-model-xyz", ... "choices":[{"message":{"role":"assistant","content":"OK"} ...   HTTP 200
```
Server log for that request:
```
POST /v1/chat/completions (1 msgs, max_tokens=4, temp=1.00, top_p=0.95, top_k=20, stream=false, thinking=false, sys=0b, user=6b, tools=0b, tool_msgs=0)
  prompt=14 tokens, max_gen=4, ctx=262144
  <- 14+1 tokens (283ms) [prefill: 86.7 tok/s, decode: 12.4 tok/s] [stop]
```

**Environment:** mlx-serve 26.8.9 (brew tap ddalcu/mlx-serve); macOS 26.5.2; M5 Max 137 GB. Models resident: `Ornith-1.5-35B-A3B-MLX-6bit` (answered), `Qwen3.8-27B-MLX-6bit` (unloaded at the time).

**Why it matters:** for an automated caller — a seat record, a bench, an agent harness with a typo'd or renamed id — the call succeeds and a different model answers, and the record looks correct. We hit it moving seats between servers and now check the id against `/v1/models` client-side before sending; a 4xx from the server lets every client fail closed for free.

**Proposed fix:** PR #258 (https://github.com/ddalcu/mlx-serve/pull/258) — in registry mode (`serve` / `run` / `--model-dir`) reject an id not present in the registry with a 404 that names the served ids; a single `--model <dir>` launch keeps the fall-through. Discriminator is `registry.discovery` (the launch shape).

Disclosure: written by Syne, an AI agent that John Sampson runs on this machine; measurements are from his box and he reviewed this before it was posted.

johnwsampson:
Fixes #257.

`POST /v1/chat/completions` with `"model":"no-such-model-xyz"` returns HTTP 200, echoes the unknown id back in `"model"`, and is answered by whatever model is resident — while `/v1/models` lists only the real ids. Same on `/v1/completions`, `/v1/messages` and `/api/chat`; they share the resolver. A client cannot tell that from a real answer: it is another model's reply under the id it asked for, which is the unload-200 class again (#209 / #211).

A server handed model roots (`serve`, `run`, `--model-dir`, every headless launch the app makes) publishes its ids on `/v1/models`, so a request naming one is routing. It now refuses an id it does not hold with a 404 that names the ids it does serve, in the shape the calling surface speaks (OpenAI `model_not_found`, Anthropic `not_found_error`, Ollama `{"error"}`). Refused above `ensureLoaded`, so a typo still loads nothing — pinned by a source-order test, same rule as the route-existence 404.

A single `--model <dir>` launch is unchanged: one model, the id is advisory, and answering a hardcoded `gpt-4o` is the whole drop-in. The discriminator is `registry.discovery` — the launch shape, not a live model count, so what a request resolves to cannot change under a client when a second model loads. Omitted `model` and the `"mlx-serve"` alias still mean the default everywhere.

Verification: `zig build test` 9/9 (and a mutation control — inverting the mode check fails exactly the new test); `tests/test_multi_model_load_failure.sh` 6/6 (its header documented the old fallback and is corrected, plus two new assertions); `cd app && swift test` — Executed 3048 tests, with 24 tests skipped and 0 failures (0 unexpected) in 49.755 (49.908) seconds; verified live on both launch shapes on macOS 26.5.2 / M5 Max with mlx-serve 26.8.9+ (this branch): registry mode → 404 naming the served ids on `/v1/chat/completions`, `/v1/completions`, `/v1/messages`, `/api/chat`, with the served id, the omitted field and `"mlx-serve"` all still 200 and `/v1/models` unchanged (the refusal loaded nothing); single-`--model` mode → `no-such-model-xyz`, `gpt-4o`, `claude-opus-4-5` all still 200.

Two things worth a maintainer's eye: (1) the resolver is shared, so in registry mode `/v1/embeddings`, `/tokenize` and the media endpoints also 404 an unknown id — I read `app/` (`prepareGenModel` uses the id `/v1/load-model` returned; `AgentEngine` sends the exempt `"mlx-serve"`) and believe the app is unaffected, but that is a reading, not a run against the app. (2) `tests/integration_test.sh`'s two `Logprobs` assertions fail on HEAD and on this branch identically — `assert_contains` pipes a body whose `max_tokens:3` tail is a truncated emoji through `grep -q` under a UTF-8 locale, which exits 1 on invalid UTF-8; `LC_ALL=C` fixes it. Separate from this PR.

Small doc note: CONTRIBUTING's build steps omit `scripts/fetch-llama.sh`; without it a fresh clone fails at link (`_ggml_commit` undefined, `lib/llama/lib` missing). Happy to send the one-line fix separately.

Disclosure: written by Syne, an AI agent that John Sampson runs on this machine; measurements are from his box and he reviewed this before it was posted.

### DIFF
--- a/src/server.zig
+++ b/src/server.zig
@@ -1537,6 +1537,76 @@ fn handleConnectionThread(args: *ConnThreadArgs) void {
     args.allocator.destroy(args);
 }
 
+/// Must a `model` id this registry does not hold be REFUSED, rather than
+/// answered by the default model?
+///
+/// Only on a server that was handed model ROOTS (`serve`, `run`, `--model-dir`,
+/// every headless launch the app makes). That registry publishes its ids on
+/// `/v1/models`, so a client naming one is ROUTING — and answering a typo from
+/// whatever else is resident echoes the unknown id back inside a 200 and hands
+/// over another model's reply. Same class as the unload that answered 200
+/// without unloading (#209): success-shaped, and no client can tell it from
+/// the real thing.
+///
+/// A `--model <dir>` launch keeps the fallback exactly as it was: it serves ONE
+/// model and the id is advisory, which is what makes it a drop-in for clients
+/// that hardcode `gpt-4o` (`registry.discovery` is null there — the launch
+/// shape, not a live count, so what a request resolves to cannot change under
+/// it when someone loads a second model). The empty id and the `"mlx-serve"`
+/// alias mean "the default" on every server and are never refused.
+fn unknownModelIdRefused(registry: *ModelRegistry, id: []const u8) bool {
+    if (registry.discovery == null) return false;
+    if (id.len == 0 or std.mem.eql(u8, id, "mlx-serve")) return false;
+    return registry.peek(id) == null;
+}
+
+/// The refusal for `unknownModelIdRefused`, in the shape the calling surface
+/// speaks. It NAMES the ids this server does serve: a refusal that quotes what
+/// it compared against is the difference between "fix the id" and "the server
+/// is broken", and the id list is exactly what the client got wrong. Bounded —
+/// a typo must not be answered with a 4 KB body — with `/v1/models` named for
+/// the tail.
+fn sendUnknownModelError(
+    allocator: std.mem.Allocator,
+    stream: *Conn,
+    registry: *ModelRegistry,
+    path: []const u8,
+    id: []const u8,
+) !void {
+    var ids: std.Io.Writer.Allocating = .init(allocator);
+    defer ids.deinit();
+    {
+        registry.mutex.lockUncancelable(stream.io);
+        defer registry.mutex.unlock(stream.io);
+        var n: usize = 0;
+        var it = registry.entries.valueIterator();
+        while (it.next()) |ep| {
+            if (n == 32) {
+                try ids.writer.writeAll(", ...");
+                break;
+            }
+            if (n > 0) try ids.writer.writeAll(", ");
+            try ids.writer.writeAll(ep.*.id);
+            n += 1;
+        }
+        if (n == 0) try ids.writer.writeAll("(none registered)");
+    }
+    const msg = try std.fmt.allocPrint(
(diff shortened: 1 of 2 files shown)


ddalcu:
@johnwsampson this is by design. It follows other engine conventions also.  Not saying your change here is in-correct, it is actually more correct. But trying to stay compatible with the existing expectations.

Did you test this with other engines to see how they behave ?

ddalcu:
🤖 **MLXServe-PRBot approved** — reviewed at head `cafe17f`, no issues found.

ddalcu:
closing this for now, we can revisit if multiple people agree on it.

PR #261, opened 2026-08-22
feat(chat): export, rename, search, and per-tab drafts

Fe2-O3:
Four chat-window basics main does not have: **Export as Markdown**, **Rename**, **Search chats**, and **per-tab composer drafts**.

### What changed since this PR was opened

It originally carried seven features. While it was open, main shipped three of them — Regenerate, Edit & Resend, and Branch Chat From Here (`Models/ChatFork.swift`). This branch had built all three independently, which is what produced 15 conflicts across three files instead of three tidy ones. Merging that would have left the app with two Branch commands and two regenerate paths.

Main's versions also do more. Its fork carries per-chat settings to the new thread, trims the transcript to a boundary the model can be handed, and hides the menu item when forking would create an empty chat.

So the branch has been merged with main and cut down to what main genuinely lacks. `ChatExport.swift`, `SidebarSearch.swift` and `ComposerDrafts.swift` survive unchanged — they were clean and self-contained. `ChatRewind.swift`, the `rerunLastTurn` engine seam, and the fork primitives added to `ChatSession`/`ChatMessage` are gone as duplicates.

Diff is now 6 files, +398/−4, down from 9 files and +934/−31. No conflicts.

### The four

| | |
|---|---|
| **Export as Markdown** | Sidebar right-click. Roles labelled, code fences verbatim, machinery left out — tool rows, failure cards and the reasoning scratchpad stay out, because an export is a record of the conversation, not of the plumbing. |
| **Rename** | Sidebar right-click. A whitespace-only title is refused rather than blanking the row. |
| **Search chats** | Titles *and* transcript text, case- and diacritic-insensitive. Hidden rows excluded. |
| **Per-tab drafts** | `ChatDetailView` is reused across tabs with no `.id(sessionId)`, so a half-typed message rode along on a switch — one Return from being sent in the wrong chat. |

### Merged, not replaced

Where main had moved, both sides were kept: the sidebar keeps its ⌘-numbering state and its multi-select delete, and search feeds the existing `SidebarSessionGroups.split` so shift-click still ranges across the Agents/Chats boundary.

The repo's own `DialogDefaultActionTests` caught the rename alert giving Return nothing to do. Its affirmative button now declares the default action.

### Tests

3059 run, 2 failures — both `AgentIntegrationTests` reaching huggingface.co. Verified failing identically on untouched main, so they are not from this branch.

---

## Follow-ups

Each is independent and can be picked up on its own.

1. **Message action tray.** A hover row of icons under each bubble: copy, regenerate, edit, branch, delete. Every one of these already exists and is already wired — `MessageBubble` receives `onRegenerate`, `onFork`, `onEdit` and `onDelete` along with the conditions for when each should be offered. They are only reachable by right-click today, so this is surfacing, not building. Keep the context menu too.

2. **Copy button on that tray, with a "Copied" flip.** `CodeBlockView` already does exactly this for code blocks — a working pattern to follow.

3. **Drafts surviving relaunch.** `ComposerDrafts` is an in-memory dictionary. Make it `Codable` and persist it beside chat history.

4. **Search result context.** Search filters the list but does not show what matched. `SidebarSearch.matches` already finds the matching message and discards it. Surface that line under the title and jump to it on click.

5. **Export to JSON.** `ChatExport.markdown` is a pure function over messages. A JSON sibling is small, and it is the format you would want for re-importing.

6. **Export more than one chat.** Main's sidebar multi-select (`appState.sidebarSelection`, the `Delete N Chats` item in `ChatView.swift`) already provides the selection this needs. Note this landed on **main** during this PR's life and was not on the original branch — which is why it may look absent if you are reading the pre-merge diff.

1 and 2 are the same file and the same pattern, and are worth doing as one task.


Fe2-O3:
## Handoff list implemented on this branch (5 commits, all tested)

Rebased onto current `main` first — which matters, because main shipped its own Regenerate / Edit & Resend / Branch while this PR was open. The merge keeps main's versions of those three and drops this branch's duplicates; what survives from the original PR (export, rename, search, drafts) is what the handoff items build on.

**1+2. Message action tray + Copied flip**
Hover tray under every USER bubble: copy / edit / branch / delete (`MessageActionTray`), copy carrying the code block's checkmark flip. Deliberate scope note: assistant rows keep their always-visible footer — this repo has a rule that hover-revealed footer actions read as awkward mid-stream — and **Branch joins that permanent footer** so it is no longer context-menu-only there. Context menus unchanged.

**3. Drafts survive relaunch**
`ComposerDrafts` is Codable; the store moved onto `AppState` and persists to `~/.mlx-serve/composer-drafts.json` beside chat history (loaded at init, saved on stash/clear). Tab-switch stash/restore wiring unchanged in shape.

**4. Search result context**
`SidebarSearch.firstContentMatch` returns row index + a one-line snippet. While filtering, each matching row shows the snippet under its title; clicking jumps: sidebar publishes `AppState.pendingSearchJump`, the reused detail view consumes it once its rows exist — releases auto-follow via a new `ChatScrollState.messageTargeted` case (otherwise the scroll core would snap you back to the bottom mid-conversation), then `scrollTo(id:, anchor: .center)`. Rows are `.scrollTargetLayout()`-ed so an offscreen id still resolves.

**5. Export to JSON**
`ChatExport.jsonData`: title, exportedAt, and per visible row role/content/timestamp (+ reasoning and media paths for re-import fidelity). Machine rows excluded, same rule as markdown. Menu now offers both formats.

**6. Export more than one chat**
Uses the multi-selection that IS on main: right-clicking inside a selection of N>1 offers "Export N Chats…" — one markdown file, one exported-at line up top, each chat under its display title in sidebar order. (For the record: at the time I wrote the handoff text, my branch had no multi-select because my base predated it; main did. Both things were true.)

**Tests:** new/updated pins in `ChatBasicsSlimTests` + `ChatScrollTests` (snippet units, jump end-to-end scan, drafts roundtrip, JSON/multi shapes, `.messageTargeted` follow-release). Suite: 3,083 executed, 24 skipped; failures are exactly the 7 environmental ones already failing on untouched main here (2 network AgentIntegration, 5 LaTeX font-layout under bare `swift test`, which `app/build.sh` staging satisfies).

Fe2-O3:
**Correction to my earlier note + one more fix pushed** (2376bd9)

I had filed the 5 LaTeX test failures as environmental ("font layout satisfied by build.sh staging"). That was wrong. Digging in: `LaTeXFonts.locate` probed `<bundle>/Fonts/KaTeX_Main-Regular.ttf`, while every real layout — SwiftPM resource bundle AND sealed .app copy — keeps fonts at `<bundle>/Contents/Resources/Fonts`. The unit tests stubbed the same flat path, so they passed vacuously while `isAvailable` was false **everywhere**: native equations silently declined to source text in the shipped app too, not just in tests.

The fix accepts both sealings; a new pin uses the real Contents/Resources shape. Full suite on this branch:

```
Executed 3084 tests, with 24 tests skipped and 0 failures
```

Zero failures. The 2 AgentIntegrationTests network failures from my earlier run also cleared when re-run with HF reachable — if they're red on CI, they were already red on main for the same network reason.

Also rebuilt the FAST_DEV bundle from this branch, so the local app carries the fix.

Fe2-O3:
@ddalcu ready for review.

_(An earlier comment here was posted and then deleted — my agent posted it before I'd approved it. Sorry for the inbox noise.)_

## What this adds

Four things the chat window didn't have: **Export** (Markdown or JSON, one chat or several), **Rename**, **Search** across titles and message text, and **per-tab drafts** that survive a relaunch. Plus a hover row of icons under each message — copy, edit, branch, delete — for actions that were previously right-click only.

## What changed since this PR opened

It started with seven features. Three of them — Regenerate, Edit & Resend, and Branch Chat From Here — landed on `main` while this was in flight. This branch had built its own versions, which is what produced 15 conflicts instead of three. Those are dropped in favour of main's, which do more: main's fork carries per-chat settings over, trims to a valid boundary, and hides itself when forking would make an empty chat.

## Bugs found and fixed during review

Seven. None of them were visible to the unit tests, and most lived in the wiring between correct logic and the view:

- **Search jump did nothing in its main case.** Clicking a hit in a chat you weren't looking at flipped the session, and that handler re-pinned to the bottom — the centred scroll landed and was corrected straight back down. Same-session clicks worked, which is why it looked fine.
- **Drafts were lost on quit.** They were only saved on a tab switch, so the tab you were actually in never got saved.
- **Search indexed rows the transcript doesn't draw.** Agent tool summaries put raw tool output into sidebar captions, and produced hits pointing at a message id no row carries — so the click scrolled nowhere.
- **Agent chats couldn't be found by name.** Search read the stored title; the row displays the agent's name.
- **A snippet could lose its own match.** The offset was measured on the raw text but applied to the trimmed line.
- **Drafts outlived deleted chats**, and deleting the active chat resurrected one.
- **A failed export was silent.** A full disk or a bad write produced no file and no message, and the save panel closes either way.

One more worth calling out on its own: `LaTeXFonts.locate()` probed `<bundle>/Fonts/…`, but every layout that ships keeps fonts under `<bundle>/Contents/Resources/Fonts`. The tests stubbed the same wrong path, so they passed while `isAvailable` was `false` in every environment — equations silently fell back to source text everywhere. A stub that mirrors the code's own wrong assumption confirms nothing.

## Verification

- 3136 tests, 2 failures — both `AgentIntegrationTests` reaching huggingface.co, verified failing identically on untouched `main`.
- 24 skips, all environment-gated: live network, sandbox interop, App Store build.
- Merged up to current `main`. No conflicts.
- Every fix has a guard, and each guard was verified red on revert before being committed green.

---

**Review credits.** The LaTeX font-probe bug was caught by **Ox Alpha**. The other six were found by **Fable** in a dedicated review pass over the diff, then verified and fixed against the source before landing.

ddalcu:
🤖 **MLXServe-PRBot approved** — reviewed at head `6bde2a2`, no issues found.

ddalcu:
closing this PR as its pretty out of date by now, please re-open if you rebase from main.
Some things have been added like rename.

PR #262, opened 2026-08-22
Add support for safety, lora_paths, and lora_scales for image edits

ghost:
Untested because I can't figure out how to compile the app. But I would really like to see this added to the app because as of right now all my image edit requests are getting blocked by content filter despite passing `"safety":false`. This pr should fix that, as well as add support for loras for image edits. But untested so can't be sure


ddalcu:
I removed NSFW all together in the main branch.

PR #263, opened 2026-08-23
Let the H3 pane reach the step and frame counts the server accepts

### LINKED ISSUE #254: Lower H3 REF2VA step count minimum to 4
I'm currently playing with https://huggingface.co/lightx2v/Minimax-h3-Turbo/blob/main/minimax_h3_ref2v_turbo_4step_v0.1_comfyui_bf16.safetensors which is the Minimax H3 4-stop turbo for REF2VA.    I've found 4 steps generates okay-ish video, but messed up audio.  I plan to try higher steps.

Currently I have to either use the API or use a locally built app, because the app has a hard lock minimum of 16 steps for H3 REF2VA.  Since the server accepts values as low as 4, the app should allow that.  There can be a warning that generations lower than 16 without a Turbo LoRA won't work, but it shouldn't restrict the steps to 16 or up.

justinluque:
Closes #254.

The Video pane hard-locked MiniMax-H3 Steps at 16 and Frames at 124 — the lowest values we have a quality verdict on — while the server has no such floors: `gen.handleVideoH3` only refuses steps below 4 *when turbo is on*, and `minimax_h3.temporalShape` aligns `max(5, length)` onto the 17k+5 ladder. The pane was strictly tighter than the API it drives.

That mattered most on REF2VA, which is exactly where @Morac2 hit it: that pack has no Turbo toggle, so a community 4-step distillation (e.g. `lightx2v/Minimax-h3-Turbo`'s REF2VA adapter) loaded through the pane's own **Style LoRAs** slot was the only way to sample cheaply there — and the Steps slider would not go there. The workaround was hand-writing the HTTP request or rebuilding the app.

## What changed

- **Steps range 16…50 → 4…50** on all three H3 presets. 4 is the server's own floor.
- **Frame ladder floor 124 → 5**, the engine's floor. The server's *default* clip length (56 frames) was previously off the slider entirely, and a cheap 1-second test clip — how you find a step count without paying for a 20-minute render — was unreachable at any setting.
- **The floors survive as advice**, not as walls: new `testedStepsFloor` / `testedFrameFloor` on `VideoModelPreset` (with per-preset note text; `0` = fully tested, so LTX stays silent) render an orange line under the control.
  - Under 16 steps: *needs a distilled few-step adapter — the Turbo LoRA, or a community one attached under Style LoRAs.*
  - Under 107 frames: *below MiniMax's stated 4-second minimum.*
- The steps advisory takes `distilled` — the Turbo toggle **or any attached Style LoRA** — because warning someone off 4 steps right after they attached a 4-step LoRA is advice against their own setup.

## What deliberately did not move

- **Quality tiers** still default to 124 frames / 16 steps. A preset is a recommendation; the short end is somewhere you steer to, never somewhere you land.
- **Chat-generated videos** (`MediaToolArgs.videoFrames`) now floor at `testedFrameFloor`. That path has no slider and no reader, so an unattended preview stays inside the trained range.

One incidental improvement: with a real ladder floor, `H3Plan.frameCap` is honest at the bottom. It used to bottom out at 124 and answer "124 frames fits" on a Mac that could not load the pack at all.

## Testing

`H3ReachableRangeTests` (new, 9 cases): server-matching ranges on every H3 preset, both advisories firing and staying silent in the right places, LTX unaffected, tiers and chat previews unchanged, and a source scan that the pane actually renders both. `MediaGenServiceTests.testMiniMaxH3FrameLadderIs17kPlus5` updated — it pinned the old policy.

Full app suite: 3114 tests, 0 new failures (the 8 LaTeX-font failures reproduce on a clean `origin/main` on this machine and are unrelated). No Zig changes — the server already accepted all of this.

Not smoke-tested against live weights: I don't have a 69 GB H3 pack on this machine. The request shape is unchanged; only the values the pane will let you put in it moved, and they moved to values the server already documents as legal.


ddalcu:
Thank you!

PR #264, opened 2026-08-23
feat(server): --api-key-strict and --api-key-env

uxsmedjan:
## What

Two additive flags on the API-key gate:

- `--api-key-strict` drops the loopback exemption, so a configured key is required from 127.0.0.1 too (`/health` and CORS preflight stay open). Use case: an embedding application that generates a fresh key per start and wants *only the key holder drives inference* on a shared machine — with the exemption, any local process (and any browser page via a simple no-preflight POST) is a key holder. No effect unless `--api-key` is set.
- `--api-key-env <VAR>` reads the key from a named environment variable instead of argv, which the world-readable process table exposes.

## How

The gate condition is extracted into `apiKeyGateApplies(key_set, strict, peer_loopback)` and the `if` in the request path now calls it; the truth table is pinned by a hermetic test at the bottom of `server.zig`. Flag parsing and `--help` follow the surrounding style.

## Test evidence

Compiled this tree and ran the suite on a real Apple Silicon Mac in this session:

```
zig build test --summary all
run test 20 pass (20 total)
zig build -Doptimize=ReleaseFast  # + packaged CLI smoke: ./mlx-serve --version → 26.8.10
```

Motivation from the embedder side: Styrverk pins mlx-serve as a managed sidecar and its security review flagged the open loopback port as its one unauthenticated local surface; this closes it without changing any default.

### DIFF
--- a/src/main.zig
+++ b/src/main.zig
@@ -308,6 +308,13 @@ fn printUsage(io: std.Io) void {
         \\                        Accepts Authorization: Bearer, x-api-key, HTTP
         \\                        Basic (key = password), or ?api_key=. /health
         \\                        stays open. Unset = no auth (default).
+        \\  --api-key-strict    Require the key from loopback too (localhost is
+        \\                        exempt by default). For embedders that want
+        \\                        "only the key holder drives inference" on a
+        \\                        shared machine. No effect without --api-key.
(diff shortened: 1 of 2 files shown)


ddalcu:
Thank you!

Issue #266, opened 2026-08-23
SIGSEGV in KV cache teardown (transformer.freeKVEntry) during long agent sessions on Mac Studio M4 Max


d-b:
## Description

Recurring SIGSEGV crashes in `mlx-serve` 26.8.9 during heavy agent workloads. The one crash with a full macOS crash report is a **use-after-free in the KV-cache teardown path**: the inference thread frees a KV cache entry whose underlying MLX array was already freed.

Three ungraceful exits in the current server log (Aug 21–23), all during long multi-message agent sessions (100k–150k token contexts, tool calls, MTP + PLD speculative decoding). One has a crash report; the other two are identified from log context.

This signature is distinct from the two fixes mentioned in #265 (huge images, math LaTeX renderer) — it is in the KV-cache lifecycle, not the media/LaTeX path.

## Environment

- **Hardware:** Mac Studio M4 Max, 128 GB unified memory
- **macOS:** 26.6.2 (25G83)
- **mlx-serve:** 26.8.9 (MLX 0.32.0), bundled in MLX Core.app (signed, team S77TW432YG)
- **Model:** `ddalcu/Qwen3.8-27B-MLX-Serve-8bit` (29.05 GB), ctx 262144, MTP depth 6 (in-checkpoint head), PLD (draft_len=5, key_len=3), hot prefix cache on (4096 MB cap / 16 entries), `max_resident_models=3`, `--max-resident-mem` 86 GB, concurrency 1 (MoE fallback)
- **Serving:** MLX Core menu-bar app; clients are agent harnesses over the LAN

## Reproduction

No deterministic repro yet. All three crashes share the same workload shape:

1. Long agent sessions — 90+ messages, 100k–150k token prompts, sustained tool-call loops
2. KV-state churn — client disconnects mid-generation, hot-cache LRU eviction under the byte budget, and (in crash 1) a full model switch with eviction
3. Speculative decoding (MTP + PLD) active in every session

The three crashes died in three different churn patterns:

| # | Last activity before death |
|---|---|
| 1 | Model switch: Qwen3.8-27B evicted (29.05 GB), `poolside/Laguna-S-2.1-NVFP4-mlx` (66.96 GB) loaded and warmed up, then died at the first prefill of the new model (`Prefill chunk: 4096 tokens (memory-sized down from 8192)`) |
[...]

PR #272, opened 2026-08-25
feat(server): POST /v1/messages/count_tokens

Fe2-O3:
## What

Implements `POST /v1/messages/count_tokens` — the Anthropic Messages token-counting endpoint. Returns `{"input_tokens": N}` for the prompt that `/v1/messages` would actually build from the same body.

Closes the gap llmprobe reports as a missing endpoint on the Anthropic surface.

## Why it reuses the real prompt path

A token count is only worth anything if it describes the prompt the server will actually build. A second, independent counting path would drift from `/v1/messages` the first time either side changed a template, a tool serialization, or a thinking default — and the drift would be silent, because a plausible number looks exactly like a correct one.

So this PR factors the prompt-building half of `handleAnthropicMessages` into `buildAnthropicPrompt`, which now owns everything derived before generation: message parsing, system / `tool_result` / image blocks, tool serialization, `tool_choice`, thinking + `output_config.effort` resolution, the schema instruction, the template render, and the vision splice. Both handlers call it. The count handler reaches no scheduler and no `Generator`; a source scan pins both halves so a future edit cannot quietly reintroduce a second path.

Because the render goes through `cachedFormatChat`, all three tokenizer arms (ds4, llama, MLX) are covered without a special case.

## Verification

Full suite on this branch: `Build Summary: 9/9 steps succeeded; 1725/1866 tests passed (141 skipped)` — 0 failures.

Live against Macaw-4bit-MLX, `tests/test_count_tokens.sh` 12/12. The load-bearing assertion is not "200 OK" — it is that `count_tokens` equals `usage.input_tokens` from a real `/v1/messages` call on the identical body, checked across three shapes:

| body | count_tokens | /v1/messages `input_tokens` |
|---|---|---|
| plain | 10 | 10 |
| with `system` | 28 | 28 |
| with `tools` | 84 | 84 |

`tests/test_anthropic_api.sh` 42/42, unchanged by the refactor. `tests/test_index_page.sh` 24/24. `html_console_test.mjs` 72/0.

## Contract decisions worth a reviewer's eye

Four places where the Anthropic contract left room, and what was chosen:

1. **Context overflow is not enforced.** `/v1/messages` 400s a prompt over `ctx_size`; this endpoint returns the number anyway. A client counts *precisely in order to discover* it is over — refusing would defeat the purpose.
2. **`model` is required.** Anthropic's schema says so, and dispatch has already resolved it, so an omitted `model` is a named 400 rather than a silent count against the default model.
3. **Images are counted, not skipped.** A body carrying an image runs the vision encoder, exactly as `/v1/messages` would. That is real GPU work, so the endpoint is only cheap for text bodies. The alternative — estimating image tokens separately — is precisely the second code path this design exists to prevent.
4. **LAN route class is `model_gated`, not `denied`.** Claude Code calls `count_tokens` every turn; `denied` would 403 the client this surface exists for.

## Also found while in here

- The vision-billing source guard was keyed on a local variable *name* rather than on the predicate, so an unrelated rename silently took it from 3 matches to 2. Retargeted at `visionPrefillUnchunked(` and both spellings of the raw bool are now banned. This was a live no-op guard, not something this PR introduced.
- Pre-existing and **not** fixed here, flagged rather than folded into a refactor: message `images[].pixels` are never freed on the success path in either `handleChatCompletions` or the Anthropic path. Behavior is unchanged by this PR.

## Base

Branched from `main` rather than `v26.8.10-pre-release.1`. Main is three commits past that tag, all same-day (#266's double-free fix, the LTX FLF fix, and issue templates), so this merges clean and picks up the double-free fix.


ddalcu:
Thanks, the shape is right (one prompt builder for both handlers) and I want this in. It cant land as is though:

- Conflicts in src/server.zig, right in the code you refactor. Main (#273, cfbf839) moved request image/video/audio ownership onto server.RequestMedia after a leak, and there is a source scan now that a request handler never owns its own media list. Your AnthropicPrompt.vision_encoded + errdefer shape will trip it, so please rebase and route the media through RequestMedia rather than carrying it on the prompt struct.
- Please make tests/test_count_tokens.sh assert the count equals usage.input_tokens of the same body sent to /v1/messages, on three shapes: plain, tools + thinking, and one with an image. Plus one run on a GGUF (llama arm).
- Full suite green after the rebase.

Then its a quick re-review.

PR #273, opened 2026-08-25
fix(server): request image/video/audio buffers leaked on every request

Fe2-O3:
## What

Request media buffers — image pixels, video frames, and audio samples — were never freed on the success path of `handleChatCompletions` or the Anthropic `/v1/messages` handler. Every multimodal request permanently retained its payload.

## Scope: larger than "image pixels"

It is all three modalities, on both surfaces:

| Path | Verdict |
|---|---|
| `handleChatCompletions` — images | leaked on every path |
| `handleChatCompletions` — videos | leaked on every path |
| `handleChatCompletions` — audio (`samples`) | leaked on every path |
| Anthropic `/v1/messages` — images | leaked on success and on every return after the parse loop |
| Anthropic — error path between decode and `messages.append` | already freed (`errdefer`) |
| `parseImageUrlContent` / `appendImageUrlContent` / `appendLfm2Tiles` internal failures | already freed |
| `/v1/responses` (`responses.ParsedInput.owned_images`) | already correct |

Both handlers ran only `defer messages.deinit(allocator)`, which frees the `Message` array and nothing it points at. There is no per-request arena — the allocator is the process GPA handed down from `handleConnectionThread` — so the retention is permanent, not per-connection.

The Anthropic path had an `errdefer` covering the error path but nothing covering success. That is the inverse of the usual shape, which is probably why it read as handled.

## Measured

Same machine, same model (`Ling-3.0-tiny-MLX-4bit`), same probe: 40 `/v1/chat/completions` requests each carrying a 3 MB `x-mlx-pixels` payload (120 MB total).

| build | RSS before → after | delta |
|---|---|---|
| base commit | 4413 → 4534 MB | **+121 MB** |
| this branch | 4405 → 4405 MB | **+0 MB** |

The retained amount is the payload, to within a megabyte.

## The fix

One owner, `RequestMedia`, installed beside the `messages` list in both handlers, rather than a `defer` added at each of the exit paths. Scattering frees across N returns is exactly how the N+1th return gets added without one — and that is the shape this bug already had.

Specifics worth reviewing:

- A media list is obtainable **only** from `openImages` / `openVideos` / `openAudio`, and is owned from its first append — so a `try` between decode and message-append cannot leak either.
- `Message` borrows the slice; ownership is not shared.
- Slots are **indices, not pointers**, because opening another slot can move the list headers.
- A fourth modality that spells its buffer something other than `pixels` / `samples` fails to compile in `mediaBuffer` rather than silently leaking.

Guarded by a source scan (`a request handler never owns its own media list — RequestMedia does`) plus a `std.testing.allocator` unit test. The unit test was mutation-checked: deleting the single free in `MediaBag.deinit` reports `4 leaks` with the exact allocation sites, so it is a real guard rather than a tautology.

## Verification

`Build Summary: 9/9 steps succeeded; 1704/1845 tests passed (141 skipped)` — 0 failures. `zig build -Doptimize=ReleaseFast` succeeds.

The source-scan guard was confirmed red before the fix existed:
```
error: 'server.test.a request handler never owns its own media list — RequestMedia does' failed:
src/server.zig:10426:55: return error.HandlerOwnsItsOwnMediaList;
```

## Honest gaps

- **No integration shell script covers this.** The existing vision scripts need a 34 GB+ checkpoint. The live A/B above is the substitute, not an equal.
- **The LFM2-VL tiling path was not exercised live** — no LFM2-VL checkpoint on hand. Its internal frees were already correct and are unchanged by this PR.

Per repo policy the rule is added to `CLAUDE.md` and the full story to `docs/gotchas/server-http.md`.


ddalcu:
Thank you!

PR #276, opened 2026-08-26
fix(app): Cover is offered with no sign that fsq.safetensors is missing

Fe2-O3:
## The bug

An ACE-Step pack downloaded before Cover mode shipped has no `fsq.safetensors`. The Music tab still shows **Cover**, and nothing tells the user why it will not work — the only signal is a 400 after they have already chosen a source track, written a caption and pressed Generate.

The fix is not to hide Cover. It is to say, where the user chooses it, exactly what is missing and offer the one file.

## Why this is worth a PR rather than a docs line

The gap between "this pack has the file" and "this pack does not" is **420 MB, not 5 GB**. `model.safetensors` is byte-identical between an old pack and the current repo — I verified the content lengths match exactly (5,081,359,469 both sides) against `ddalcu/ACE-Step-1.5-XL-Turbo-MLX-Serve-8bit`, which gained `fsq.safetensors` (420,148,465 bytes) on 2026-08-23. A user who cannot tell those apart re-downloads the whole pack, or gives up on the feature.

## No server change

`src/` is untouched, deliberately.

`fsqAvailable()` lives on a loaded `acestep.Engine`, so exposing an FSQ capability over `/v1/models` or `/props` would mean loading a multi-GB pack to answer a UI question — and would advertise less capability for an unloaded pack than its stub does, which the project rules forbid.

The app can answer locally instead: `ServerManager.resolveModelDir(repo:)` already resolves through `ModelRoots.readRoots()`, so packs in LM Studio and custom folders are covered, and the question is one `fileExists` on `<packDir>/fsq.safetensors`.

The filename is a three-way contract (`src/acestep.zig` `FSQ_FILE`, `src/gen.zig`'s cover gate, and now the app) with no compiler between the halves, so a read-only source-scan guard pins it — the same idiom `H3PlanningTests` uses for `turbo_lora.safetensors`.

## No second download mechanism

Reuses `DownloadManager.startPackFile(repoId:fileName:)`, the existing "fetch one file into a pack you already have" path that `startTurboLora` was built on, including its surgical cancel that cannot kill an enclosing full-pack download.

## The four states, and the copy

`CoverWeightsFetch.decide(...)` is a pure function returning one of:

**`.fetch`** — the file is missing and the folder is writable:
> Cover needs fsq.safetensors, a 420 MB tokenizer this pack predates. It downloads into the pack you already have — not the whole model again.
>
> `[ Download fsq.safetensors (420 MB) ]`

**`.downloading`** — progress bar, percentage, Cancel:
> Fetching fsq.safetensors (420 MB) into this pack. Cover works as soon as it lands.

**`.missingUnwritable(dir:)`** — never offer a button that cannot work:
> Cover needs fsq.safetensors (420 MB), and MLX Core cannot write to this pack's folder. Download it from the model's Hugging Face page and put it in:
> `<pack dir>` *(selectable, so it can be copied)*

**`.unavailableRemotely`**:
> Cover runs on the Mac hosting this model; it needs fsq.safetensors in ITS copy of the pack.

The mode picker segment also reads `Cover — fsq.safetensors missing` while missing, so the reason is legible without clicking anything.

No dialog is added, so `DialogDefaultActionTests` is unaffected.

## Vocal to BGM is unaffected — verified in source, not assumed

`src/acestep.zig:1939` switches on `req.task`: `.complete` calls `contextLatents(src_lat.?)` directly, and only `.cover` reaches `audioTokenize`/`audioDetokenize`, the sole `ensureFsq()` callers. `src/gen.zig:2249` gates on `task == .cover` alone. `CoverWeightsFetch.decide` returns `.notApplicable` for `.complete` and `.text2music`, pinned by `testDecisionOnlyAppliesToCoverOnASourceAudioModel`.

Disabling both modes would have been a worse bug than the one being fixed.

## No reload needed

`acestep.Engine.fsqAvailable()` re-stats the directory whenever `fsq_w` is null, so a resident model picks the file up as soon as it lands. Noted in a code comment; the UI deliberately does not promise a restart that is not required.

## Verification

`swift build` clean (warnings pre-existing). `swift test`:

```
Executed 3116 tests, with 24 tests skipped and 8 failures (0 unexpected)
```

**The 8 failures are pre-existing and unrelated** — all `LaTeXFontsTests` / `LaTeXRenderingTests`. Confirmed by stashing every change and re-running the clean tree: same 8, zero added here. The new suite is 39/39.

Tests were written first. The initial red was `cannot find 'CoverWeightsFetch' in scope`. The source-scan guard was separately proven to bite rather than pass vacuously — stashing only `MusicGenView.swift` produces 5 failures, one per needle.

## Honest gaps

- **The real 420 MB download was not exercised.** Both states are covered by temp fixtures; the transfer path itself is unchanged `startPackFile` code already covered by `TurboLoraFetchTargetTests`.
- **Writability uses `FileManager.isWritableFile`.** That is the honest in-process check, but under the App Sandbox it can be optimistic for a folder reached through a scoped bookmark. If a real unwritable-pack report appears, that predicate is where to look. The unwritable path is tested against a `0o555` temp dir.
- The existing Generate-time auto-fetch in `MusicGenService` is left in place as the net behind this upfront declaration.


ddalcu:
Thank you!

PR #277, opened 2026-08-27
test: pin the hd-256 fused prefill arms instead of reading them off the host

lojza3d:
## What

Four `fusedSdpa256Prefill` tests fail on NAX-class silicon (M5). They are test-only failures - the engine is correct, the tests just assume the machine they run on.

## Why they fail

`fusedSdpa256Prefill`'s causal arm deliberately yields to the stock fused sdpa when NAX is available (`transformer.zig:2583`):

```zig
} else {
    if (fused256CausalMode() == .off) return null;
    // The band arm stays ours (mlx's NAX kernel takes no band); the
    // causal arm yields to the stock fused kernel on NAX.
    if (naxSdpaPreferred()) return null;
}
```

`naxSdpaPreferred()` defaults to whether the host has NAX, so on an M5 the function returns null and the tests that require it to engage fail:

- `causal parity vs composed SDPA (GQA, ragged shapes, chunk offset)` - `error.FusedDeclined`
- `budgeted kv chunking engages and is exact vs single dispatch` - `error.FusedDeclined`
- `causal parity at Qwen 24q/4kv geometry (gqa 6, ragged 64-row tile)` - `error.FusedDeclined`
- `causal and band both default FUSED (budgeted-dispatch flip)` - `causal != null`

## Fix

The seam for exactly this already existed and had no callers:

```zig
pub var nax_sdpa_override: ?bool = null; // test seam
```

The three parity tests pin it to `false`, so the kernel under test runs on every machine rather than only where the runtime policy happens to select it.

The defaults test claimed "No override, no env: BOTH arms engage", which is true only off NAX. It now pins both branches: off NAX both arms engage; on NAX the causal arm declines while the band arm stays ours. That asymmetry was not pinned by anything before - a decline is invisible from the outside, since the composed fallback is output-equivalent. Renamed accordingly.

## Verification

M5 Max, macOS 26.6, `[nax-probe] arch=applegpu_g17s macos=26.6 available=true`

Before:

`Build Summary: 7/9 steps succeeded (1 failed); 1715/1863 tests passed (144 skipped, 4 failed)`

After:

`Build Summary: 9/9 steps succeeded; 1719/1863 tests passed (144 skipped)`

And with the runtime preference forced off, so the pinning holds either way:

`MLX_SERVE_NAX_SDPA=0 zig build test -Dtest-filter="fusedSdpa256Prefill" --summary all`
`Build Summary: 9/9 steps succeeded; 8/8 tests passed`

Engine code untouched — `src/transformer.zig` test blocks only.

### DIFF
--- a/src/transformer.zig
+++ b/src/transformer.zig
@@ -33447,6 +33447,11 @@ test "fusedSdpa256Prefill: causal parity vs composed SDPA (GQA, ragged shapes, c
     const s = mlx.gpuStream();
     fused256_override = true;
     defer fused256_override = null;
+    // The causal arm yields to the stock fused kernel on NAX silicon, so on an
+    // M5-class Mac this parity check would DECLINE instead of running — the
+    // kernel under test would never be exercised. Pin the arm, not the host.
+    nax_sdpa_override = false;
+    defer nax_sdpa_override = null;
     var prng = std.Random.DefaultPrng.init(0x256256);
     const rnd = prng.random();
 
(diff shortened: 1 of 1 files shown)


ddalcu:
Thank you!

PR #279, opened 2026-08-28
fix: align M5 SDPA routing

PhilipJohnBasile:
M5 NAX now owns plain head-dimension 256 causal SDPA above the vector width, but the custom prefill and split tests still assumed the pre-NAX routing.

This makes test overrides authoritative, routes query length 9 causal verify blocks to stock NAX, keeps query lengths 6 to 8 on the split vector path, and keeps sliding-band attention on the custom kernel.

Validation:
- `zig build test -Dtest-filter="fusedSdpa256Prefill"`: 8 passed
- `zig build test -Dtest-filter="splitCausalSdpa"`: 4 passed
- `zig build test -Dtest-filter="sdpa force_fused"`: 2 passed
- `zig build test`: 1,723 passed, 145 skipped, 0 failed

### DIFF
--- a/src/transformer.zig
+++ b/src/transformer.zig
@@ -2534,6 +2534,10 @@ pub fn fused256CausalMode() Fused256CausalMode {
     return mode;
 }
 
+fn customCausalPrefillOwns(nax_preferred: bool, override_active: bool) bool {
+    return override_active or !nax_preferred;
+}
+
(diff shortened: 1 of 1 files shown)


ddalcu:
Superseded by #277 which pinned the hd-256 prefill arms on main. If the q=9 split decline on NAX is a measured win on its own, open it as a small PR with a same-boot number.

PR #282, opened 2026-08-28
mtp: measured G17 cost surface for the qwen4_exp head (M5 Max)

holtsway:
M5 NAX plan, item 1 (`qwen-flash-next_M5_NAX_plan.md`, Phase 1).

## What

`MtpHeadRef.costProfile`'s `.qwen4` arm hard-coded `.generic`, so Flash-Next
planned MTP rounds on M5 with the M4-fitted default costs (lo .10/hi .26)
while the real marginals are 3-5x steeper. This adds a measured
`g17_nax_qwen4_q4_gs64` surface, selected by full runtime fingerprint per the
house discipline, everything else stays `.generic`.

- `MTP_EV_G17_NAX_QWEN4_Q4_GS64_COSTS`, fitted at ~8.5k context like every
  shipped prior. Forced-depth saturated echo, depths {1,2,3,4,6}, two reversed
  passes, 3 reps, round-cost persist off, medians of per-request `round_ms`:
  T(1)=25.41, T(2)=31.18, T(3)=35.70, T(4)=41.33, T(6)=53.58 ms → 20.34 ms
  floor, composite (draft + per-position) marginals .257/.303 (SSE 0.28) —
  shipped house-style as `.draft = 0.02` plus per-position `.237/.283`,
  composite minus draft, the split every calibrated surface carries.
  Short-context sweep for the record (incl. depth 8):
  23.61/28.12/33.91/38.87/50.16/62.70 → floor 18.24, composite .283/.310/.344
  (the .344 ships as per-position `.325`, inert under cap 6).
- **Cap stays 6, not the NAX cap 8**: a verify row on this arch is bytes at
  every width — routed expert banks never ride the vqmm NAX lane — so there is
  no flattening past position 6 to open depths 7-8 for. (Positions 7-8
  measured .344/pos against sub-60% tail acceptance on saturated echo. This
  table is the grouped-expert kernel's before-number — plan item 2.)
- Selection validates the measured runtime only: uniform affine-4/gs-64
  config, both head fc packs at that width (K derived from the scales), NAX
  lane available and not env-killed. One-shot engagement log.
- New lever `MLX_SERVE_MTP_QWEN4_PROFILE=0` revokes the PROFILE only (compute
  lanes untouched) — the isolation arm for cost-surface A/Bs, where
  `MLX_SERVE_VERIFY_QMM_NAX=0` would change the very costs under test.

## Why it matters live

With generic costs the controller overreaches on marginal content in the
cold window before the in-run round-cost table matures. The reproducible
failure — BOTH clean pairs, first prose request: **20-24% BELOW serial**
(53.4 vs 66.3, and 50.0 vs 66.1 tok/s, same boot). With the measured surface
the worst prose cell across both pairs is 0.908x serial. Once the table
matures the arms converge by design; the prior owns exactly the window where
generic fails. Net: equivalent steady-state throughput, cold-start tail risk
removed.

## A/B (same binary, `MLX_SERVE_MTP_QWEN4_PROFILE=0` as the generic arm,
`MLX_SERVE_ROUND_COST_PERSIST=0` both arms, counterbalanced boot order,
within-boot MTP/serial ratios; M5 Max 40-core / 128 GB, AC)

Medians of 3 within-boot MTP/serial ratios; prose worst-rep in parens (the
cold cell). Absolute serials drifted 0.5-7% between boots, both directions.

| pair (boot order)          | arm     | code  | prose (worst)   | 8.5k  |
|----------------------------|---------|-------|-----------------|-------|
| c (generic→profile)        | generic | 1.475 | 1.018 (0.805)   | 1.106 |
| c — short-ctx-fitted prior | profile | 1.447 | 1.036 (0.973)   | 1.064 |
| d (profile→generic)        | profile | 1.492 | 0.983 (0.908)   | 1.077 |
| d — the shipped 8.5k fit   | generic | 1.418 | 1.042 (0.756)   | 1.104 |

Code is spec-cell variance (lead flips with boot order). Prose medians are
equivalent within variance; the profile halves prose spread and removes the
below-serial disasters. 8.5k keeps a persistent small generic edge
(~2.5-4% relative, equal floors) — the kv term is learned online and the
in-run table takes over; a kv-aware prior is future work, noted in the
gotcha story.

Engagement: `[mtp] cost profile g17-nax-qwen4-q4-gs64 engaged` present in
every profile-arm log, absent in every generic-arm log; `[spec-stats]
mode=mtp` in all MTP arms.

## Two A/B traps documented (docs/gotchas/engine-mlx.md + CLAUDE.md rules)

1. A cost-profile A/B with the persisted round-cost table live measures the
   TABLE, not the profile (`MtpCostSource` hands planning to a matured
   bucket) — persist off on both arms.
2. The within-boot serial cell is the drift meter: same-evening boots drifted
   5-12% on profile-independent serials; only ratios compare.

## Tests

- Pure classifier + degraded-condition fallbacks (mtp.zig).
- Profile → costs/cap selection, override fallbacks, cap loops (generate.zig).
- Behavioral: warmup EMAs keep base depth 3 but never extend to depth 8;
  cold-start plan stays at one exposed position.
- Full `zig build test`: 0 failures on M5 Max, default env (the four
  pre-existing M5 kernel-test failures were fixed upstream in #277 and run
  green here).
- llmprobe before/after (bench.sh row, medians of 3, same box; NOTE these
  boots load the persisted round-cost table, so they exercise shipping
  defaults rather than isolating the prior — the ratio tables above are the
  controlled evidence):
  before `| qwen38-flash-next | 91 mtp | (prefill 1509, 3.00 tok/step)`,
  after `| qwen38-flash-next | 98 mtp | (prefill 1565, 3.00 tok/step)`;
  spec predictable 105.5→106.6, novel 64.6→66.5, 16.3k decode 82.1→80.9.
  No cell regresses beyond the variance the drift meter showed.

Measured on the exact target: M5 Max 40-core, 128 GB, macOS 27.0, Xcode 27,
`ddalcu/Qwen3.8-Flash-Next-MLX-Serve-4bit`.

## Folded from #284 (per review)

The grouped-expert NAX kernel is **dropped from the tree**; only its
CLAUDE.md rule and gotcha story land here. Summary: the per-block profiler
confirms the MoE owns ~60% of the verify-row marginal, but a grouped m16
tensor-ops kernel — parity-perfect on every group shape, 11/11 on
test_mtp_equivalence armed — loses ~9% on the forward in both variants
(S=2 20.30/20.29 vs 18.66/18.62 ms; S=4 27.12/26.96 vs 24.91/24.88):
verify-width runs are 1-2 rows so the tile never amortizes, and sorted
gather_qmm already sits near the expert-byte floor (~1.7 ms/row). No opt-in
lane ships; the working kernel + hermetic parity test + ubench arms stay
findable at
[holtsway/mlx-serve@feat/qwen4-moe-verify-nax](https://github.com/holtsway/mlx-serve/tree/feat/qwen4-moe-verify-nax),
linked from the gotcha story.


ddalcu:
Thanks for both PRs, Could you fold them into one, keeping the diff minimal:

- Keep the profile this PR shipps. The shipped constants (.237/.283/.325) dont match the PR and doc comment (.257/.303/.345), pick one and make the gotcha story match the 8.5k fit.
- From #284 keep only the CLAUDE.md rule and the gotcha story, drop the kernel. Link the branch in the story so its findable.
- No opt-in paths for things that lost. If its not the default, it isnt in the diff.
- Please attach the llmprobe JSON reports and the server logs for each arm (before/after), plus the per-rep raw numbers behind the drift claim. 

Net speed change on M5 is about zero, so lets keep going. The MoE verify row being bytes is the wall. Next worth measuring: wider verify (bigger draft blocks or batched verify) so expert runs get wide enough to tile, and a kv-aware prior so 8.5k stops losing 3-4%. Same bar as before: same boot A/B, 3 reps, engaged lines from the log.

If you can, try Claude Fable 5 for this, it does noticeably better on kernel and controller work.

**Important !** You will most likely need to use "Macs Fan Control" app for this, and set it to full blast, M5Max gets really hot on the Mackbook pro's, and it can make your results look bad, when they are actually much better due to thermal throttle.

holtsway:
Done, all three folded and pushed (346c456):

- **Constants unchanged** — the shipped `.237/.283/.325` are per-position; `.257/.303/.345` was the composite (draft .02 + per-position), the same split the other calibrated surfaces carry. The doc comment, gotcha story and PR body now state the mapping explicitly, and the story leads with the 8.5k fit (short-context kept for the record, labeled, with the −4% lesson).
- **#284's kernel dropped from the tree**; its CLAUDE.md rule + gotcha story land here, with the working kernel, hermetic parity test and ubench arms linked at [holtsway/mlx-serve@feat/qwen4-moe-verify-nax](https://github.com/holtsway/mlx-serve/tree/feat/qwen4-moe-verify-nax) so the null stays reproducible.
- **No opt-in lane ships.**

Full `zig build test` green after the fold.

holtsway:
Evidence attached: https://gist.github.com/holtsway/1880b25cbf2225757d301b2a3ce0473d — llmprobe JSON reports + bench server logs (before/after), the per-arm server logs for both clean A/B pairs plus the table-contaminated first pair (labeled), and the per-rep raw numbers behind the drift claim, with the boot-order/serial tables laid out.

On the fan advice — confirmed, and thank you: our absolute serials sagged across the evening (code 72.5 → 66.8 across boots, 8.5k 64.3 → 58.7, partial recoveries between — timeline in the raws doc). The within-boot ratio methodology was the defense, but the absolute cells in those logs are thermally depressed. Macs Fan Control is installed and future arms run fans full-blast with macmon telemetry logged beside the engaged lines, per your logs' convention.

On next steps, aligned, in this order:

1. **kv-aware prior** — starting now. `MtpEvCosts.kv_ms_per_token` already exists (learned online, prior seeds 0); our two-context fits bracket it: floor 18.24 ms at kv≈120 vs 20.34 ms at kv≈8.6k → B ≈ 0.25 ms/1k tokens for this pack. Plan: add a mid-context sweep point to check linearity, seed B in the qwen4 surface, same bar — same-boot A/B, 3 reps, engaged lines, fans pinned.
2. **Wider verify** after that: at depth-cap widths runs stay 1-2 rows, so the interesting regime is batched verify (multiple slots' rows in one MoE pass) — that's where cross-slot expert overlap finally gives the tile something to amortize, and the parked kernel branch becomes the starting point. Bigger scheduler surface (the MTP slot is exclusive today), so we'll scope it against your conventions before writing anything.

And yes — this whole series has been driven with Claude Fable 5 end to end. It wrote the kernel, caught its own null, and is typing this sentence.

holtsway:
kv term follow-up, per "a kv-aware prior so 8.5k stops losing 3-4%":

**B is real and measured; the 8.5k deficit turns out not to be prior-addressable.** Three-context floors (fans pinned this time, 44-51°C under load — night and day, thank you: prefill cells went 1509 → 1840): 18.24 ms @ ~120 tok, 19.47 @ ~4.2k, 20.34 @ ~8.6k — linear, r² 0.991 → **B = 0.248 ms/1k tokens**. The marginal composites stay context-stable across all three fits; the floor alone carries the kv dependence, exactly the T(k,L) = W + B·L + C(k) the field docs describe.

The catch: `kv_ms_per_token`/`kv_ref_tokens` were unwired (nothing consumes them), and once wired + seeded, the term is ZERO at its own reference by construction — a prior fitted AT 8.5k cannot move 8.5k plans. Measured accordingly: same-boot 8.5k pair, seeded vs not = no change (1.053 vs 1.069 within-boot MTP/serial ratios, within variance, engaged lines in both arms). The residual ~3% vs generic at 8.5k is table-mediated — both arms converge onto the in-run table; the gap lives in cold rounds and trial cadence, not in a missing prior term.

Two more findings from the runs: (1) the below-ref deflation (physically real) measurably hurts — it codifies the fitted model's known conservatism exactly where wider plans win — so the wired term is **inflation-only**, byte-identical below ref (llmprobe 512 cell 102.9 vs 102.7 unseeded). (2) cross-boot llmprobe resolution is ±5% even fans-pinned (four near-identical binaries read 91.6–105.4 on the top decode cell); at this effect size only within-boot ratios resolve. One harness note: `qwen4_ab.sh`'s kill pattern (`mlx-serve --model`) misses a renamed `MLX_SERVE_BIN` — a stale server answered one arm's port and produced a phantom arm; the engaged-lines rule caught it.

The wiring + `floor_ref_ms` anchor + measured seed + behavioral tests (zero-at-ref, inflation-only, non-narrowing plans, unconditional T(k,L)/T(1,L) compression) are on [holtsway/mlx-serve@feat/qwen4-kv-aware-prior](https://github.com/holtsway/mlx-serve/tree/feat/qwen4-kv-aware-prior), full suite green. At 16k the term adds +0.09 floor units (sub-noise); it becomes principled at 32k+. Happy to open it as a small PR (default-on, provably inert ≤ ref) or leave it as the recorded answer — your call. Next on our side: scoping batched verify, where expert-run widths finally reach the parked tile's regime.

ddalcu:
merged, thank you 🫡

holtsway:
Sure! I've used a bit of your work and am happy to contribute where I can.

PR #284, opened 2026-08-28
moe: grouped-expert NAX lane for verify widths — a measured null (M5 plan item 2)

holtsway:
M5 NAX plan, item 2 (`qwen-flash-next_M5_NAX_plan.md`, Phase 2) — executed
to a conclusive **measured null**, shipped per the plan's own rule
("otherwise document as measured null and leave disabled") and the crossrow
lane's precedent: the lane lives in-tree OPT-IN default-OFF with the numbers,
so nobody re-runs this experiment blind.

## What the measurements say

**The premise was half right.** Per-block profiling (deltas-only, the M4
methodology) confirms the MoE owns the verify-row marginal on M5 Max: +4.31
ms/row of +7.12 total at S1→S2 (61%), +2.41 of +4.13 at S2→S4 — hc/gdn/attn
all small after the row-batched kernels. But the missing piece is not NAX:
the SORTED gather_qmm already sits near the expert-byte floor (~1.7 ms/row =
10 experts x 2.46 MB x 48 layers / ~700 GB/s), and verify-width routing
(S<=9, top-10-of-many) produces runs of 1-2 rows — the m16 tile's row
amortization never engages while its costs are fully paid.

## The kernel (correct, and slower)

Grouped-expert m16 tensor-ops tile over the pair-sorted activations: one
threadgroup per equal-expert run (in-kernel group discovery — a host
readback mid-layer is a GPU barrier), gate and up banks streamed through the
tile with the swept-table SwiGLU fused in-kernel at the sort path's exact
rounding sites, sorted [N, inter] output so the down path is untouched.
Host wrapper with full eligibility declines (4-bit/gs64, K%256, inter%32,
rows<=16, NAX probe + lane env), 16-row zero padding, one-shot engagement
log, `moe_vnax_override` test seam.

- Hermetic parity: mirrored-fp32-truth max-diff < 0.02 across duplicate-run /
  all-distinct / 16-row-single-group / mixed routings; kill-switch decline leg.
- Live: `test_mtp_equivalence.sh` 11/11 with the lane armed; engagement
  lines present in armed logs, absent otherwise.
- Forward cost (ubench, same methodology both arms, `[fwd-ubench]` lines):

| variant | S=2 on | S=2 off | S=4 on | S=4 off |
|---|---|---|---|---|
| v1 (two K passes, NSG 8)          | 20.30 | 18.66 | 27.12 | 24.91 |
| v2 (single pass, dual acc, NSG 4) | 20.29 | 18.62 | 26.96 | 24.88 |

~−9% both variants; halving barriers/passes moved nothing, so the cost is
structural, not synchronization.

## Why it ships anyway (OFF)

`MLX_SERVE_MOE_VERIFY_NAX=1` arms it; anything else is byte-for-byte the
stock path. The declining mlp marginal at wider S (4.31 → 2.41 ms/row) is
cross-row expert overlap slowly arriving — a wider-batch regime (S=16+,
larger draft blocks, batched verify) could revisit this with the amortization
the tile needs, from these numbers instead of from zero. Rule line in
CLAUDE.md, full story in docs/gotchas/engine-mlx.md (beside the multi-row
gather +38% and per-row gatherQmv −11% entries — same lesson, other angles).

Full `zig build test` green including the new hermetic test. Measured on
M5 Max 40-core / 128 GB, macOS 27.0, Xcode 27,
`ddalcu/Qwen3.8-Flash-Next-MLX-Serve-4bit`, AC power, 45 s settles.


ddalcu:
closing this PR, please fold it into #282  so we can keep it all together.

PR #296, opened 2026-08-29
feat: FLUX.1 dev and FLUX.1 schnell support

justinluque:
Adds the FLUX.1 architecture: T5-XXL text encoder + tokenizer, CLIP text tower, dedicated VAE, and the flux1 DiT/pipeline, shared plumbing factored into flux_common.zig. Wires into gen.zig model discovery, lora.zig stacking, and the app's media bundle/model catalog. (sorry for the incoming barrage of pull requests😄)


ddalcu:

- `src/flux1.zig:978` — `clipTokenize` runs the CLIP-L prompt through `tokenizer.zig`'s generic byte-level BPE, but the CLIP tokenizer is a different scheme: `Lowercase` normalizer, a whitespace-removing `Split` pre-tokenizer, and a BPE model with `end_of_word_suffix: "</w>"` (word-final vocab entries are `cat</w>`, merges are keyed on it). `tokenizer.zig` has none of that — no lowercasing, GPT-2 pretok that keeps the leading `Ġ` space on each word, and `bpeMerge` never appends `</w>`, so word-final merges never fire.
  Failure: "A cat" tokenizes to roughly `[bos, Ġ?, A, Ġ, c, a, t, eos…]` instead of `[bos, a</w>, cat</w>, eos…]` — the pooled CLIP vector fed into `conditioning()` is computed from the wrong ids on every prompt. The pipeline still produces a finite image (the env-gated smoke test only checks range/finiteness), so this degrades prompt adherence silently rather than erroring. A parity fixture against `CLIPTokenizer` ids (like `tests/dump_t5_fixtures.py` does for T5) would catch it.

ddalcu:
🤖 **MLXServe-PRBot approved** — reviewed at head `448090b1`, no issues found.

ddalcu:

PR #297, opened 2026-08-29
feat(anima): Anima image generation (Cosmos-Predict2 DiT + LLMAdapter)

justinluque:
Native support for circlestone-labs/Anima — anime/illustration text-to-image, served by a new `anima` media backend alongside the existing image engines (FLUX.2, Krea, Mage-Flow).

Architecture: a Cosmos-Predict2 2B DiT (28 blocks, AdaLN-modulated self/cross-attention + MLP) conditioned by a 6-block Perceiver-style LLMAdapter (T5-vocab query tokens cross-attending to Qwen3-0.6B hidden states), decoded by the Qwen-Image causal VAE. Two text tokenizers run per prompt: Qwen BPE feeds the text encoder, a new T5 SentencePiece unigram/Viterbi tokenizer feeds the adapter's t5xxl_ids. Sampling is rectified flow (fixed shift=3.0, CONST schedule).

Feature-complete: text-to-image, img2img via the VAE's own encoder, runtime LoRA (stacked, same grammar as the other image backends), and both the turbo (10-step, CFG 1.0) and base (32-step, CFG 4.5) variant schedules — including the CFG dual-forward path, which needed a fix for an empty-prompt tokenization edge case that only base's classifier-free guidance branch exercises.

Wired into the app: Image pane model picker, capability declarations, architecture recognition for locally-discovered checkpoints, and two downloadable catalog presets (Turbo/Base) pointing at published mlx-serve mirrors. Conversion script + fixture dumpers + integration test suite included.


ddalcu:

- `src/anima.zig:2324` — the empty-prompt fallback for the Qwen encoder uses `self.qwen_tok.eos_id orelse 0`, but the pack's `tokenizer/` is `vocab.json`+`merges.txt` only (both published mirrors `justintime47/Anima-*-MLX-Serve` ship no `tokenizer.json`), and `loadTokenizerSlow` builds no `added_tokens`, so `eos_id` is null and the empty string encodes as token id 0 (`!`). ComfyUI's `Qwen3Tokenizer` (`has_end_token=False, min_length=1, pad_token=151643`) pads the empty string to a single `<|endoftext|>` (151643) instead.
  Failure: any Base-pack request (`recommended_cfg` 4.5) without a `negative_prompt` builds its CFG unconditional branch from a `!`-token hidden state rather than the reference's `<|endoftext|>` state, so outputs diverge from ComfyUI for the default case.
- `src/gen.zig:2062` — `negative_prompt` is read with `extractJsonString` and passed through raw, whereas `prompt` at line 1837 goes through `jsonUnescape`.
  Failure: a negative prompt containing JSON escapes (`\"`, `\n`, or the `\/` that Swift's JSONSerialization emits for `/`) is tokenized with the literal backslash sequences.

ddalcu:
🤖 **MLXServe-PRBot approved** — reviewed at head `ab3d84e1`, no issues found.

ddalcu:

cellsheet:
I would really like this support as well! I didn't realize this already had a PR existing for it, Anima is specifically trained for anime images on danbooru unlike any of the currently available models (FLUX.2, Mage, Krea) which are just general purpose. There just isn't an all in one solution for running Anima currently as it requires ComfyUI and this would fit perfectly. There's also a v1.1 for the aesthetic and turbo versions of these models in the upstream repo https://huggingface.co/circlestone-labs/Anima/tree/main/split_files/diffusion_models

Edit: Actually I would literally donate for this PR to go thru

justinluque:
Hi @cellsheet, for now I would be happy to produce a build for you with this change

cellsheet:
@justinluque I found your branch I can make a build from 😁 thanks for working on that, hopefully this can get re-evaluated for merging later on.

Edit: I went ahead and forked ddalcu's repo then merged your PR commits into my fork and got the code base up to date, this is also info for anyone else who wants justinluque's Anima model's supportted on mlx-serve. I also put in a github action runner to periodically merge in any changes into my repo and I'll try my best to continue to resolve merge conflicts.

PR #298, opened 2026-08-29
feat: classifier-free guidance for FLUX.2 klein base checkpoints

justinluque:
Distilled klein has guidance baked in and needs no CFG, but the undistilled "base" checkpoint requires it: encode a negative prompt at the same fixed FLUX_SEQ_LEN, run cond/uncond DiT forwards sharing img_ids/txt_ids, and blend v = uncond + guidance_scale*(cond-uncond). guidance_scale defaults to 1.0, which collapses to the no-CFG path.


ddalcu:
🤖 **MLXServe-PRBot approved** — reviewed at head `6e8b7d7`, no issues found.

PR #301, opened 2026-08-29
feat: stablediffusion

justinluque:
This pull request introduces native support for Stable Diffusion image generation, including SD1.5 SD-turbo, and SDXL in both base and Turbo variants, into the server and the app. It adds comprehensive stable diffusion model handling, negative prompt support, advanced configuration options, and ensures cross-component consistency and documentation. The changes also improve the handling of model downloads, configuration, and media model type detection.

**Stable Diffusion and Features:**

- Added native SDXL image generation support, including both SDXL 1.0 and SDXL Turbo, with detailed documentation and changelog entries describing the architecture, LoRA adapter support, negative prompt handling, download behavior, and parity with diffusers. [[1]](diffhunk://#diff-06572a96a58dc510037d5efa622f9bec8519bc1beab13c9f251e97e657a9d4edR109-R117) [[2]](diffhunk://#diff-6ebdb617a8104a7756d0cf36578ab01103dc9f07e4dc6feb751296b9c402faf7R42-R43) [[3]](diffhunk://#diff-6ebdb617a8104a7756d0cf36578ab01103dc9f07e4dc6feb751296b9c402faf7R398-R406)
- Negative prompt and classifier-free guidance controls are now available as advanced, sticky settings for SDXL in `ImageGenSettings`, with proper persistence and decoding logic. [[1]](diffhunk://#diff-f968622daf2de88944fc0ecf936401c3c96028b754fec6c08d823538fce34d66R71-R79) [[2]](diffhunk://#diff-f968622daf2de88944fc0ecf936401c3c96028b754fec6c08d823538fce34d66R166-R167)
- The app and backend now recognize SDXL and SD1 as media model types, ensuring correct architecture gating and UI support. [[1]](diffhunk://#diff-7cd7542704941b56eac1445cd997cdb1c4aa1e255aeba0189225def1382e14bfL70-R70) [[2]](diffhunk://#diff-7cd7542704941b56eac1445cd997cdb1c4aa1e255aeba0189225def1382e14bfL86-R86) [[3]](diffhunk://#diff-7cd7542704941b56eac1445cd997cdb1c4aa1e255aeba0189225def1382e14bfR189)

**Model Loading, Download, and Configuration Improvements:**

- Enhanced model type detection and classification logic to support SDXL/SD1 based on diffusers' `model_index.json`, and clarified how media types are mapped and loaded.
- Improved download manager logic to correctly handle flat and nested quantization variants for SDXL, ensuring only the necessary files are kept based on the variant's structure.

**Documentation and Legal Notices:**

- Added extensive documentation and technical notes in `CLAUDE.md` on SDXL quirks, tokenizer handling, quantization, single-file checkpoint conversion, and more. [[1]](diffhunk://#diff-6ebdb617a8104a7756d0cf36578ab01103dc9f07e4dc6feb751296b9c402faf7R398-R406) [[2]](diffhunk://#diff-6ebdb617a8104a7756d0cf36578ab01103dc9f07e4dc6feb751296b9c402faf7R247)
- Updated `NOTICE` to include OpenAI CLIP tokenizer license and embedding details, as these are now bundled for single-file SDXL checkpoint support.

These changes collectively enable robust SDXL support, advanced prompt controls, and improved model management in both the backend and the app, while ensuring legal compliance and clear documentation.


ddalcu:
this is for the 3 year old model ? why ?

justinluque:
They're fast, and have a smaller memory footprint so they're more accessible for people with older devices. Also the massive ecosystems of loras and finetunes.

ddalcu:
I mean... Flux works on my phone using https://apps.apple.com/us/app/mlx-chat-local-ai/id6787603518

Smaller than that ? 

I want to avoid adding bloat to mlx-serve, another path for adding stable diffusion, would be to simply include the SDCPP library, you can see what i did in https://github.com/ddalcu/zig-ai

justinluque:
Flux is awesome, but SD1.5/Turbo (2–4GB VRAM) and SDXL still fill a specific niche for lots of users: sub-second generation speeds, lighter memory draw, and native compatibility with the huge library of community fine-tunes and LoRAs (like Pony and Illustrious). If the path forward is just to wrap GGML libraries like sd.cpp, wouldn't that logic also apply to using llama.cpp for text, which kind of defeats the purpose of an MLX-specific server entirely?

I totally get wanting to prevent scope creep, though, so if you'd prefer to keep mlx-serve strictly focused on current LLM/DiT architectures, I'm more than happy to maintain the SD pipeline on my fork.

ddalcu:
good point, one thing on my roadmap is to add a plugin system...  and this SD could be part of the plugin, much like Pi coding agent works and its plugins.  Let me think about it, because this can solve other old models also.

justinluque:
that's a great idea!

ddalcu:

- `src/cli.zig:127` — the new diffusers-layout aliases (`sdxl`, `sdxl:turbo`, `noobai`, `noobai:vpred`, `sd1`, `sd1:turbo`) have no `subdir`, so `wantedFile` falls through to `shouldDownload`, which (line 322) rejects every nested path except `mtp/…`. Nothing under `unet/`, `vae/`, `text_encoder*/`, `tokenizer*/`, `scheduler/` is ever fetched.
  Failure: `mlx-serve pull sdxl` downloads `model_index.json` plus the root single-file checkpoints (`sd_xl_base_1.0.safetensors` and `sd_xl_base_1.0_0.9vae.safetensors`, 6.9 GB each; `sdxl:turbo` is 13.9 GB + 6.9 GB, `sd1` is 4.3 GB + 7.7 GB) and none of the subdirs. `Engine.loadAuto` then sees `model_index.json`, takes the diffusers `load()` path, and fails on the missing `text_encoder/`/`unet/` — the single-file fallback is only tried when the index is absent. The cli test covers only `pony` (flat) and `illustrious` (subdir), so this is untested. Verified against the HF file listings for all six repos.

- `src/gen.zig:2168` — `negative_prompt` is passed straight from `extractJsonString` without the `jsonUnescape` that `prompt` gets three lines above at 1933.
  Failure: `{"negative_prompt": "text, \"watermark\", caf\u00e9"}` reaches the CLIP tokenizer with literal `\"` and `\u00e9` sequences, so the encoded unconditional branch does not match what the client sent.

- `src/sdxl.zig:157` — `trailing` spacing uses the integer `n / steps` stride, but diffusers `EulerDiscreteScheduler` uses the float ratio: `round(arange(1000, 0, -1000/steps)) - 1`.
  Failure: for any step count that does not divide 1000 the ladders diverge (e.g. 6 steps: diffusers `[999, 832, 666, 499, 332, 166]`, here `[999, 833, 667, 501, 335, 169]`), so the Lightning/Turbo `trailing` path only matches the reference at 1/2/4/5/8/10/20/25… steps; the schedule tests exercise only 8 and divisors.

ddalcu:

PR #304, opened 2026-08-29
Feat/html js chat renderer

justinluque:
This pull request introduces a new feature for rendering HTML and SVG code blocks in chat, giving users control over whether these blocks open as live previews or as source code by default. It also adds a new "Chat" settings category where this preference can be configured, and ensures the new setting is properly encoded, decoded, and reset with other settings. The main changes are as follows:

**HTML/SVG Preview Feature:**
- Added support for rendering closed ```html and ```svg code blocks as live, interactive previews within the chat transcript, using a secure `WKWebView` and a toggleable Preview/Code header. The default behavior is to show the preview, but this can be changed in settings. [[1]](diffhunk://#diff-cb82e5c2bd201ba350a51a9ba43c9ccd760d7de7d64f37e39728880e0207fabeR39) [[2]](diffhunk://#diff-cb82e5c2bd201ba350a51a9ba43c9ccd760d7de7d64f37e39728880e0207fabeR76-R77)

**Settings Model and Persistence:**
- Introduced a new `htmlPreviewsByDefault` Boolean property to `ServerOptions`, which determines whether HTML/SVG blocks open as previews or as source code. This is a rendering preference and does not require a server restart.
- Updated the `ServerOptions` decoder to handle the new `htmlPreviewsByDefault` property, ensuring it is loaded from user settings.

**Settings Organization:**
- Added a new `chat` case to the `SettingsCategory` enum, with appropriate display name and icon, to group chat rendering preferences separately from request defaults. [[1]](diffhunk://#diff-7adc715067ec2833eb52dca66b9d99f596c4c9a1edafd28e0f74fc08aee421cdR26-R28) [[2]](diffhunk://#diff-7adc715067ec2833eb52dca66b9d99f596c4c9a1edafd28e0f74fc08aee421cdR52) [[3]](diffhunk://#diff-7adc715067ec2833eb52dca66b9d99f596c4c9a1edafd28e0f74fc08aee421cdR84)
- Updated the settings reset logic to include `htmlPreviewsByDefault` under the new "Chat" category, so it is reset when chat settings are restored to defaults.


ddalcu:
cool, will test this out, was one of the things deep on my todo list !

justinluque:
A warning though that is is still a WIP😆, been on my todo list for a while too

justinluque:
image

ddalcu:
because of html pages being huge, and we already have the existing built in Browser, I would prefer to leverage that, as some sort of "canvas" as in ChatGPT.

I added the ability to "show" the browser via tool call, so technically we can use that going forward, versus doing it inline. It needs some more love, but I want to go that direction versus inline in the chat, since HTML needs to be able to be stretched and resized, and so on.

Closing this PR for now, but lets think of a way to combine the built in browser, with what you did here.

PR #307, opened 2026-08-29
fix(prefix-cache): bound SSM checkpoints on the replace path, keeping them spread

### LINKED ISSUE #306: Hot prefix cache: ssm_checkpoint_max and the byte budget are not enforced on the replace path
## What happened

Long agent sessions on a hybrid recurrent model eventually fail with:

```
Prompt (224530 tokens) requires ~6492MB GPU memory but only ~6326MB available.
Reduce prompt size or use a smaller model.
```

on a machine with ~42 GB free. Retrying the same prompt can never succeed. Only a
*different, smaller* request clears it.

## What I expected

`--ssm-checkpoint-max 8` to bound the SSM checkpoints an entry retains, and
`--prefix-cache-mem 12GB` to bound hot-cache residency. Neither holds for an entry
that is extended in place.

## Environment

- mlx-serve 26.8.11-pre-release.1 (MLX 0.32.2)
- macOS 26.6.2, Mac17,6 (M5), 128 GiB unified memory
- Model: `Qwen3.8-Flash-Next-Uncensored-MLX-Serve-4bit` — arch `qwen4_exp`, 4-bit affine
  (group 64), 125B total / ~6B active MoE, 48 layers = 36 `linear_attention` (GDN) +
  12 `full_attention`, `mamba_ssm_dtype: float32`
- Flags: `--mtp --prefix-cache-mem 12GB --prefix-cache-disk 20GB --ssm-checkpoint-max 8 --kv-quant turbo4`
- Client: a coding agent driving `/v1/chat/completions`, appending to the same
  conversation every turn (660+ messages, 46 KB of tool definitions)

## Evidence

The eviction log reports SSM bytes separately, which makes the cause visible:

```
[hot-cache] evicted LRU entry (byte budget; was  83116 tokens, 20967.12 MB; ssm 18885.12 MB)
[hot-cache] evicted LRU entry (byte budget; was 204835 tokens, 23282.37 MB; ssm 21886.81 MB)
[hot-cache] evicted LRU entry (byte budget; was 224475 tokens, 32740.68 MB; ssm 31237.11 MB)
```

31237 MB of SSM in a single entry is ~34 checkpoints under a cap of 8. The KV part is a
steady ~24 KB/token (~6.9 KB/token under `--kv-quant turbo4`); the SSM part does not
track token count at all, only how many checkpoints the entry accumulated.

Residency against a 12288 MB budget, at **one** entry:

```
[hot-cache] resident=30484.23 / 12288.00 MB (2/32 entries)
[hot-cache] resident=32740.68 / 12288.00 MB (1/32 entries)
```

266% of the budget. Growth was a steady ~700 MB per turn, i.e. exactly one new
checkpoint per turn, none dropped.

## Cause

`HotPrefixCache.commitWithState` has two paths. The insert path enforces the count cap
and the byte budget. The **replace** path — taken whenever the new tokens extend the
existing entry, which is every turn of an agent conversation — does neither:

1. It merges the previous entry's checkpoints with this turn's, deduplicating by
   position, and nothing re-applies `ssm_checkpoint_max` to the merged list.
   `generate.zig` applies the cap per prefill, which no longer bounds a list spanning
   many turns. `HotPrefixCache` has no `ssm_checkpoint_max` field at all.
2. It `return`s before the byte-budget eviction below it.

The `--ssm-checkpoint-max` help text offers the byte budget as the fallback for
`0 = unlimited`, but on this path both are inactive, so there is no bound left.

## Why retrying cannot recover

Admission reserves memory against hot-cache residency, so a bloated cache turns away the
next large prompt. Eviction only runs on insert, and a rejected request never inserts —
so the request that would trigger the eviction is exactly the one being refused. In my
logs a small unrelated request broke the cycle and dropped residency from 32740 MB to
57 MB in 1.5 s, after which the next real request still matched 99% of its prefix.

## Fix

PR follows. Give `HotPrefixCache` its own `ssm_checkpoint_max`, wired from the scheduler,
apply it to the merged list oldest-first exactly as the capture path does, and run the
byte budget on the replace path too.

Measured with the patch, same workload and flags:

| | before | after |
|---|---:|---:|
| SSM per entry (~225k tokens) | 31237 MB | 7047 MB |
| checkpoints per entry | ~34 | 8 (= cap) |
| peak residency | 32740 MB | 11056 MB |
| as % of the 12288 MB budget | 266% | 90% |
| growth per turn once at the cap | ~700 MB | ~11 MB |
| rejections | 4 in one day | 0 |

kartalbas:
Fixes #306.

## The bug

`HotPrefixCache.commitWithState` has two paths. The insert path enforces the count cap and
the byte budget. The **replace** path — taken whenever the new tokens extend the existing
entry, i.e. every turn of an agent conversation — does neither:

1. It merges the previous entry's checkpoints with this turn's, deduplicating by position,
   and nothing re-applies `ssm_checkpoint_max` to the merged list. `generate.zig` applies
   the cap per prefill, which no longer bounds a list spanning many turns. `HotPrefixCache`
   had no `ssm_checkpoint_max` field at all.
2. It `return`s before the byte-budget eviction below it.

`--ssm-checkpoint-max`'s help offers the byte budget as the fallback for `0 = unlimited`,
so on this path both bounds are inactive at once.

## The change

- `HotPrefixCache.ssm_checkpoint_max`, wired from the scheduler next to the existing
  `entry.ssm_checkpoint_max` assignment.
- After the merge, drop oldest-first while over the cap — the same `orderedRemove(0)` the
  capture path uses, for the same reason: the newest positions sit closest to
  end-of-prompt, where warm multi-turn requests match.
- Run the byte budget on the replace path. The refreshed entry has just been bumped to the
  newest `last_used`, so `evictOneLru` cannot select it while another entry exists; the
  loop is guarded on `entries.items.len > 1`.

88 insertions, no deletions, no new plumbing.

## Tests

Hermetic test added at the bottom of `prefix_cache.zig`: two turns, cap 2, four checkpoints
merged, asserting both the count and that the surviving positions are the newest.

Red before green — with the trim removed:

```
error: 'prefix_cache.test.HotPrefixCache: replace path re-applies ssm_checkpoint_max across turns' failed:
       expected 2, found 4
Build Summary: 7/9 steps succeeded (1 failed); 1/2 tests passed (1 failed)
```

With the fix:

```
Build Summary: 9/9 steps succeeded; 2/2 tests passed
+- run test 2 pass (2 total) 417ms MaxRSS:25M
```

Full suite on this tree: `Build Summary: 9/9 steps succeeded`.

I did not add a new script under `tests/` — the change is visible over HTTP only through a
long multi-turn session, which I ran against a real client rather than scripted. Happy to
add one if you want it in the suite.

## Human testing

macOS 26.6.2, Mac17,6 (M5), 128 GiB. `Qwen3.8-Flash-Next-Uncensored-MLX-Serve-4bit`
(`qwen4_exp`, 36 GDN + 12 full-attention layers), flags
`--mtp --prefix-cache-mem 12GB --prefix-cache-disk 20GB --ssm-checkpoint-max 8 --kv-quant turbo4`,
driven by a coding agent over a long session (660+ messages, 46 KB of tool definitions),
built with `zig build -Doptimize=ReleaseFast` and served for several hours.

Same workload, same flags, before and after:

| | before | after |
|---|---:|---:|
| SSM per entry at ~225k tokens | 31237 MB | 7047 MB |
| checkpoints retained per entry | ~34 | 8 (= cap) |
| peak residency | 32740 MB | 11056 MB |
| as % of the 12288 MB budget | 266% | 90% |
| growth per turn once at the cap | ~700 MB | ~11 MB |
| prompt rejections | 4 in one day | 0 |

The eviction line that shows the cap holding, at a context length where the old build was
carrying 31 GB:

```
[hot-cache] evicted LRU entry (byte budget; was 228698 tokens, 8604.60 MB; ssm 7047.04 MB)
```

7047.04 / 8 = 881 MB per checkpoint, against 874 MiB predicted for that position — so the
list is at exactly the cap. The byte budget also demonstrably fires now, at 11056 MB of the
12288 MB budget when a second entry arrives, where the old build ran to 32740 MB at one
entry.

Prefix reuse is unaffected: 99.77% cache hit across the session, no `hybrid miss`, and no
extra cold prefills.

### DIFF
--- a/src/prefix_cache.zig
+++ b/src/prefix_cache.zig
@@ -155,6 +155,13 @@ pub const HotPrefixCache = struct {
     /// Enforced on `commit`: evict LRU entries (in addition to the count
     /// cap) until `current_kv_bytes + new_entry_bytes <= max_kv_bytes`.
     max_kv_bytes: u64,
+    /// Cap on SSM checkpoints kept per entry, mirroring `generate.zig`'s
+    /// `ssm_checkpoint_max`. That one bounds a SINGLE prefill; this one bounds
+    /// the replace path's merge, which concatenates the previous entry's
+    /// checkpoints with this turn's. Without it an entry extended in place —
+    /// every turn of an agent conversation — gains one checkpoint per turn for
+    /// the life of the session. 0 = unlimited.
+    ssm_checkpoint_max: u32 = 0,
     /// Running total of `kv_bytes` across all live entries. Updated on
     /// commit/evict/invalidate.
     current_kv_bytes: u64,
@@ -731,6 +738,17 @@ pub const HotPrefixCache = struct {
                 }
                 self.allocator.free(old);
                 self.allocator.free(new);
+                // The merged list spans BOTH turns, so the per-prefill cap in
+                // generate.zig no longer bounds it. Re-apply it here, dropping
+                // oldest-first exactly as the capture path does: the newest
+                // positions sit closest to end-of-prompt, which is where warm
+                // multi-turn requests match.
+                while (self.ssm_checkpoint_max > 0 and
+                    merged.items.len > self.ssm_checkpoint_max)
+                {
+                    var oldest = merged.orderedRemove(0);
+                    oldest.deinit(self.allocator);
+                }
                 break :blk try merged.toOwnedSlice(self.allocator);
             };
 
@@ -767,6 +785,18 @@ pub const HotPrefixCache = struct {
             e.mtp_bytes = new_mtp_bytes;
             e.last_used = self.bumpCounter();
             self.current_kv_bytes += e.kv_bytes;
+            // The insert path below enforces the byte budget; this path used to
+            // return without it, so an entry that only ever grows could hold the
+            // cache above `max_kv_bytes` indefinitely. `e` was just bumped to the
+            // newest `last_used`, so `evictOneLru` cannot pick it while another
+            // entry exists.
+            if (self.max_kv_bytes > 0) {
+                while (self.current_kv_bytes > self.max_kv_bytes and
+                    self.entries.items.len > 1)
+                {
+                    self.evictOneLru("byte budget");
+                }
+            }
             if (self.disk != null) self.disk_dirty = true;
             self.logResident();
             return;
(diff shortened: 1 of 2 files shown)


kartalbas:
Updated: the cap no longer drops oldest-first.

Running the first version against a real session surfaced the flaw. Bounding the merged
list fixed the size, but every survivor then sat at the end of the prompt, so a request
that diverged earlier found nothing at or below its match:

```
[hot-cache] hybrid miss (no checkpoint <= 16382 of 178509); cold prefill
<- 178509+63 tokens streamed [prefill: 430.5 tok/s, decode: 28.8 tok/s]
```

178509 / 430.5 = **415 s** of cold prefill. Two of those in one session, plus a 110 s and a
61 s one. Traded a memory failure for a latency failure — not a good deal.

Oldest-first is the wrong policy here for a second reason: a checkpoint costs roughly a
constant plus a term linear in its position, so it discards the cheap early ones and keeps
the large late ones. For 8 checkpoints over a 230k-token prompt:

| retention | cost | coverage |
|---|---:|---|
| 8 newest | ~7.0 GB | last ~16k tokens |
| 8 spread | **~4.3 GB** | whole prompt |

The spread is both cheaper and more useful. So the cap now thins the interior — always
keeping the first and the newest, dropping whichever checkpoint sits between the closest
pair of neighbours, i.e. the one whose removal widens the coverage gap least. `n` is at
most `ssm_checkpoint_max`, so the quadratic scan is irrelevant.

The regression test now asserts the spread rather than a raw count: two turns, cap 4, eight
checkpoints merged, expecting 100/300/500/800. Oldest-first reports `expected 100, found
500`.

Full suite on this tree, actually executed (not a cached step):

```
Build Summary: 9/9 steps succeeded; 1759/1905 tests passed (146 skipped)
```

kartalbas:
Heads-up: the merged state carries a latency regression, and it is my fault for not getting the revision in before the merge.

Bounding the list is right, but dropping **oldest-first** collapses every survivor onto the end of the prompt. A request that diverges earlier then finds no checkpoint at or below its match and pays a full cold prefill:

```
[hot-cache] hybrid miss (no checkpoint <= 16382 of 178509); cold prefill
<- 178509+63 tokens streamed [prefill: 430.5 tok/s, decode: 28.8 tok/s]
```

That is 415 s. I measured two of those in one session on the merged behaviour, plus a 110 s and a 61 s one. It is also the expensive choice — a checkpoint costs a constant plus a term linear in its position, so oldest-first keeps the large late ones and discards the cheap early ones.

#310 replaces the policy: thin the interior, always keeping the first and the newest. At 8 checkpoints over a 230k-token prompt that is ~4.3 GB spread over the whole prompt instead of ~7.0 GB bunched at its end, with 0 cold prefills over 39 requests where the merged policy produced 4.

Sorry for the split — I pushed the revision seven minutes after you merged.

Issue #208, opened 2026-08-17
[Feature Request] Per-step preview frames on `POST /v1/video/generations` SSE (H3 / LTX)


Rhystic1:
## Summary

When `stream: true` on `POST /v1/video/generations`, mlx-serve already emits one SSE `progress` event per sampler step. Those events carry only `stage` / `step` / `total`. There is no image.

A MiniMax-H3 clip at 1344×768 / 124 frames can take ~50 minutes on an M4 Max. Clients can draw a progress bar, but they cannot show the video forming. That is a long time to stare at a percentage.

Please add an **opt-in** per-step preview: a small JPEG (or a short animated WebP) on each `progress` event. Full VAE decode at the end stays as it is.

## What happens today

Request:

```http
POST /v1/video/generations
{"model":"…MiniMax-H3…","prompt":"…","stream":true,"steps":30,…}
```

Each step writes roughly:

```json
{"type":"progress","stage":"Generating","step":8,"total":30}
```

Then a terminal `complete` with the full `rgb8` payload.

That matches `src/gen_sse.zig` (`StreamCtx.cb`) and the H3 / LTX emit sites:

- `src/minimax_h3.zig` — `p.emit("Generating", i + 1, req.steps)` after each denoise step
- `src/ltx_video.zig` — same pattern (`ProgressWindow`)
- `src/flux.zig` — same for images

The progress callback is only `(stage, step, total)`. The SSE writer uses a **256-byte** stack buffer, so a frame cannot be attached without changing that path.

## Why a client cannot fake this

A Gradio / HTTP client never sees the 24-channel H3 latent. The only pixels mlx-serve produces today are the **final** VAE decode after the last step.

ComfyUI can show live H3 frames because a custom node wraps the sampler in-process and decodes `x0` every step with a tiny autoencoder. See [simsim9-stack/ComfyUI-MiniMaxH3-PreviewOverride](https://github.com/simsim9-stack/ComfyUI-MiniMaxH3-PreviewOverride). That node ships `taeh3_decoder.safetensors` (TAEHV, 24 latent channels, ~9.9M params, ~40 MB). Holdout PSNR ≈ 25.5 dB. Decode is milliseconds; a full H3 video VAE every step is not safe.

mlx-serve is the only place that has the latent during sampling. The preview has to be produced there.

## Proposed API (opt-in, backward compatible)

Default behaviour unchanged. Only when the client asks:

```json
{
  "stream": true,
  "preview": true,
[...]

PR #308, opened 2026-08-29
feat: opt-in per-step JPEG previews on video SSE (#208)

Rhystic1:
This implements #208 

Built successfully with all tests passing on my local M5 Max MacBook Pro.

```
FAST_DEV=1 bash app/build.sh
=== Fast build complete (v26.8.10, swift debug, mlx-serve ReleaseFast) ===
App: /Users/rhystic1/Documents/repos/mlx-serve/app/MLX Core.app
MLX staged successfully with 5,772 NAX symbol hits.
codesign: valid on disk
health: {"status":"ok"}
Three live chat API smoke tests passed.
```


ddalcu:
Thanks for this PR, a good feature !

Well-built plumbing, but the image it ships isn't a preview.

Blocking:
- The "Latent2RGB" is invented. preview.zig maps channel i to a golden-angle hue (cos(i*2.4)), then mean/std-stretches per frame. That's not a fitted latent-to-RGB projection (LTX's real one has per-channel coefficients; H3 has none published). Output will be a colourful noise blob that doesn't resemble the video. There's no oracle or perceptual test proving it looks like anything, and the shell test only checks JPEG magic bytes. For LTX, port the real coefficients; for H3, fit a 24x3 map against VAE decodes or drop it.
- App forces it on for every video gen (body["preview"]=true always), so every step now pays an x0 sub/mul + unpatchify + eval + f32 host copy. LTX copies the whole 128-ch volume each step. Not measured. Should be a toggle, or at least measured before default-on in the app.

Should fix:
- previewJPEG reads 4 key aliases ("older clients also look for image/b64_json/preview_image") that never existed. YAGNI.
- docs/building.md / build.zig claim preview-test is the Linux-only-valid graph, but it's fine; on a Mac it still demands staged mlx (verifyMlxStage runs first). Minor.
- No CLAUDE.md Layout row for preview.zig/jpeg.zig (growth policy).
- Merge latest main.

If possible please make sure everything here is tested and built on a Mac with Fable if possible.

Rhystic1:
Hi @ddalcu and thank you for your comment.

I have pushed another commit which should address everything you mentioned. Tested locally on M5 Max MacBook Pro, as per before.

image

ddalcu:
thank you !

PR #310, opened 2026-08-30
fix(prefix-cache): keep bounded SSM checkpoints spread, not oldest-first

kartalbas:
Follow-up to #307, which is already merged. That PR bounded the merged checkpoint list and
fixed the unbounded growth, but it drops **oldest-first**, and running the merged behaviour
against a real session showed that is the wrong policy across turns.

## What goes wrong with oldest-first

Within one prefill it is right. Across turns it collapses every survivor onto the end of the
prompt, so a request that diverges earlier finds nothing at or below its match:

```
[hot-cache] hybrid miss (no checkpoint <= 16382 of 178509); cold prefill
<- 178509+63 tokens streamed [prefill: 430.5 tok/s, decode: 28.8 tok/s]
```

178509 / 430.5 = **415 s** of cold prefill. Two of those in one session, plus a 110 s and a
61 s one. That trades a memory failure for a latency failure.

It is also the expensive choice: a checkpoint costs roughly a constant plus a term linear in
its position, so oldest-first discards the cheap early ones and keeps the large late ones.
For 8 checkpoints over a 230k-token prompt:

| retention | cost | coverage |
|---|---:|---|
| 8 newest | ~7.0 GB | last ~16k tokens |
| 8 spread | **~4.3 GB** | whole prompt |

## The change

Thin the interior instead, always keeping the first and the newest: drop whichever
checkpoint sits between the closest pair of neighbours, i.e. the one whose removal widens
the coverage gap least. `n` is at most `ssm_checkpoint_max`, so the quadratic scan is
irrelevant. One file, no new plumbing.

## Tests

Red before green — the test now asserts the spread rather than a raw count. Two turns, cap 4,
eight checkpoints merged, expecting 100/300/500/800. With oldest-first restored:

```
error: 'prefix_cache.test.HotPrefixCache: replace path bounds SSM checkpoints and keeps them spread' failed:
       expected 100, found 500
Build Summary: 7/9 steps succeeded (1 failed); 1/2 tests passed (1 failed)
```

With the change:

```
Build Summary: 9/9 steps succeeded; 1759/1905 tests passed (146 skipped)
```

## Human testing

macOS 26.6.2, Mac17,6 (M5), 128 GiB. Same model and flags as #307
(`--mtp --prefix-cache-mem 12GB --prefix-cache-disk 20GB --ssm-checkpoint-max 8 --kv-quant turbo4`),
built with `zig build -Doptimize=ReleaseFast` and served to a coding agent. 39 requests at
~66k tokens average:

```
hybrid misses 0   evictions 0   rejections 0
residency 1588.42 -> 1930.58 MB over 14 turns, i.e. +1 to +6 MB per turn
```

Modelled 1841 MB against 1930 MB measured, 5% off. The same context under the merged
oldest-first policy would hold ~2796 MB, so this is ~31% less memory with full coverage —
and no cold prefills, where the merged policy produced four.

### DIFF
--- a/src/prefix_cache.zig
+++ b/src/prefix_cache.zig
@@ -739,15 +739,43 @@ pub const HotPrefixCache = struct {
                 self.allocator.free(old);
                 self.allocator.free(new);
                 // The merged list spans BOTH turns, so the per-prefill cap in
-                // generate.zig no longer bounds it. Re-apply it here, dropping
-                // oldest-first exactly as the capture path does: the newest
-                // positions sit closest to end-of-prompt, which is where warm
-                // multi-turn requests match.
+                // generate.zig no longer bounds it. Re-apply it here — but not
+                // oldest-first. Within one prefill that is fine; across turns it
+                // collapses the survivors onto the end of the prompt, and then a
+                // request that diverges early finds no checkpoint at or below its
+                // match and pays a FULL cold prefill:
+                //     [hot-cache] hybrid miss (no checkpoint <= 16382 of 178509)
+                // That one cost 415 s. Oldest-first is also the expensive choice:
+                // a checkpoint costs roughly a constant plus a term linear in its
+                // position, so it discards the cheap early ones and keeps the
+                // large late ones.
+                //
+                // Thin the interior instead, always keeping the first and the
+                // newest: drop whichever checkpoint sits between the closest pair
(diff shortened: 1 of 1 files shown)


ddalcu:
thanks again !  good catch

Issue #288, opened 2026-08-28
Chat history stores attachment bytes inline


lojza3d:
### Problem

# Chat history stores attachment bytes inline

`~/.mlx-serve/chat-history.json` carries images, video frames and audio samples inline as base64. On a perfectly ordinary conversation that is 97% of the file, and the file is rewritten in full on every save.

Measured on macOS 26.6.

## Measurement

One conversation, 11 messages: a generated image and an uploaded image, one question about each.

```
file:              1,049,739 B  (1.00 MB)
conversations:     1
messages:          11
inline base64:     1,020,352 B  (97.2% of the file)
text:              ~29 kB
```

The two attachments, decoded out of the JSON:

| message | kind | format | raw | base64 |
|---|---|---|---|---|
| 7 | generated | JPEG 1344x768 | 317,811 B | 423,748 B |
| 9 | uploaded | JPEG 1024x1024 | 447,452 B | 596,604 B |

Full resolution, no cap on either path. `ChatView.nsImageToJPEG` re-encodes an upload at `compressionFactor: 0.85` and never resizes, so a photo straight off a phone would land in the history at its full pixel count.

## What goes inline today

Three carriers on `ChatMessage`, all `Data` in the model and therefore base64 in the file:

| field | what it holds | worst case |
|---|---|---|
| `ChatImage.data` | uploads, generated images, and `browse` screenshots | one full-resolution JPEG per image |
| `ChatVideo.frames` | `[Data]`, one JPEG per sampled frame | `VideoPreprocessor.maxFrames = 16`, so 16 full-resolution JPEGs per uploaded video |
| `ChatAudio.pcm` | float32-LE 16 kHz mono | 64 kB per second, so a one-minute clip is 3.8 MB raw |

Video is the largest single case: an image costs one copy, a video costs sixteen.

## Why it compounds

`AppState.saveChatHistory()` encodes the **entire** `[ChatSession]` array and writes it on every save. So the megabyte above is not paid once, it is paid on every change to any conversation, including ones with no attachments at all.

## The part that looks like an oversight

Message 7 carries **both**:

```
7  assistant  images[] inline base64 = 423,748 B  |  media[] path -> generations/images/...-a-retro-crt-television-set-with-a-chunky.png
```

The generated PNG is already on disk in `generations/images/`, in its original lossless form. The history keeps a second, lossy JPEG copy of the same picture.

That is the case `app/CLAUDE.md` already decided the other way for tracks and clips:

[...]

PR #313, opened 2026-08-30
Store chat image attachments on disk, not in the history

lojza3d:
Closes the uploads half of #288. Generated media was PR #293.

## What changed

Uploaded images no longer ride `chat-history.json` as base64. They are written to
`~/.mlx-serve/attachments/` and the history carries a path, the same shape
`ChatMediaRef` already uses for generated media.

Measured on my own conversation set: `chat-history.json` went from 1 583 578 B to
80 KB, with four new image conversations added on top.

`AttachmentStore` is the one place that decides where an attachment goes, what it is
called, and when it may be deleted.

## Bytes are kept, not re-encoded

Every capture path now asks for the file's own bytes first. A dropped 4.2 MB PNG is
stored byte for byte (verified with `shasum` against the original); before this it
became a JPEG at `compressionFactor 0.85`.

A `loadObject(ofClass: NSImage.self)` decodes and throws the encoding away, so the
drop handler asks `loadItem` for the type the provider actually REGISTERED. A Finder
drag of a PNG registers `public.png`, not `public.image`, so a request for the
umbrella type fails.

Only a paste with no file behind it has to be encoded, and that goes to PNG rather
than JPEG: the pasteboard usually carries uncompressed TIFF, these are mostly
screenshots, and JPEG rings around text.

## Deleting

Files are written on send, not on attach: the preview row can drop an attachment
again, and a file per abandoned pick is how a folder fills with things no
conversation names.

Deleting conversations removes what they own minus what a surviving session still
names. `ChatFork` copies messages, so a fork and its source can point at the same
file, and deleting one must not take the picture out of the other. Paths outside the
root are never touched, since the path comes out of saved history and is data, not a
constant.

## Old histories

`ChatImage` decodes an old record without throwing: the `data` key is simply unknown
now, `path` is absent, and the transcript draws a row saying the attachment is no
longer on disk. A required field would throw, and `loadChatHistory`'s `?? []` turns
one unreadable image into an EMPTY history, every conversation rather than one
picture.

No migration, as agreed in the issue. Existing conversations lose their pictures on
first save. Worth a line in the release notes.

## One thing that came along

`MultimodalContent.build` drops an image with no bytes. Sending it anyway is an
`image_url` with an empty payload: the model is told there is a picture and handed
nothing. A consequence of the bytes leaving the history, not scope creep.

A `browse` screenshot deliberately gets no file. Its bytes now live only as long as
the app runs, which is all they ever did anything with: the transcript hides tool
messages (`ChatRows.rows` drops everything with a `toolCallId`) and the agent loop
sends images from the last USER message only, so a reopened conversation neither
draws them nor feeds them to the model. Writing them out would fill the folder with
files nobody looks at, kept until the conversation is deleted. Their base64 leaves
`chat-history.json` either way.

## Testing

`swift build` clean, full `swift test` green. 19 new tests in `AttachmentStoreTests`,
weighted to the things that are invisible until they bite: filename sanitizing, the
name a dropped file keeps, the fork-delete guard, root containment, and an old
history decoding.

Manually, on the built app:

- paste, Finder drop, paperclip: file lands in `attachments/` under the name it came
  with, model reads the image, `shasum` matches the original
- restart: images still there, history 80 KB
- delete a file by hand: the transcript draws the missing-attachment row, no crash
- delete a conversation: its files go, others stay
- fork, then delete the source: the fork's picture survives (tried across three forks
  off one source with images added afterwards)

## Known, not fixed here

A drag out of a browser page still does nothing. It did nothing before this change
either: Brave's provider registers `public.jpeg` with no loader block behind it, so
nothing loads by any route (`Cannot load representation of type public.jpeg` /
`No loader block available`). Worth its own issue if this is meant to work.


ddalcu:
Nice work, this is close to mergeable. One real regression I want fixed first:

Before this change every attach got re-encoded to JPEG, so a dropped HEIC (iPhone default), TIFF or BMP always worked. Now you store and send the file's own bytes, and for non-gemma models those raw bytes go straight to the server, which only decodes png/jpeg/gif/webp (stb_image + libwebp, no HEIC). So a dropped iPhone photo now silently fails vision on qwen and friends. `ext(for:)` also labels unknown bytes as png, which makes it worse.

The fix is small: in `payload(for:)` keep the original bytes only when they sniff as png/jpeg/gif/webp, and transcode everything else to PNG (you already have the encode path). Plus a test for it.

Rest looks good: single chokepoint, deleteSessions is the only removal path so the hook covers everything, fork guard, traversal guard, legacy histories decoding without nuking the whole file. The base64URL sniff and the empty-payload filter are good catches.

Minor notes, fine as is: decode still pulls all bytes into memory so RSS doesn't change, only the json file shrinks. A failed write loses the picture on next launch (you documented it). Message edits can orphan files until the chat is deleted.

Fix the HEIC case and I'll merge.

lojza3d:
Thanks @ddalcu, I'm on it. Two things I want your call on before I push.

## PNG vs JPEG for the transcode

You said transcode everything else to PNG. I'd like to make it **JPEG q0.95, with
PNG only when the bitmap has an alpha channel**, and I don't want to slip that past
you.

What lands in this branch is a photo, since HEIC is the iPhone default. A 12 Mpx
frame is a couple of MB as HEIC and tens of MB as PNG, because a lossless codec has
nothing to exploit in camera noise, and it would be preserving the artifacts of a
lossy source with perfect fidelity. q0.95 is far above what the vision tower can use
after its own smart-resize.

PNG stays where it is the right answer: a bitmap with alpha (JPEG has none and would
turn transparency black), and a paste with no file behind it, which is nearly always
a screenshot.

Straight PNG everywhere is one line if you prefer it.

The gate itself follows your shape: `sniffedExt` answers nil instead of guessing,
and the whitelist is exactly what `decodeRgbOwned` handles. WebP was missing from
the old sniff, so it goes in too. The stored extension will describe the stored
bytes rather than the incoming name, so a converted `IMG_1234.HEIC` lands as
`<uuid>_IMG_1234.jpg` - the sniff is the only thing that ever reads a stored file
back.

## The orphans

Slightly worse than "until the chat is deleted": `deleteMessage` drops the message
from the array, so the path goes with it, and `removablePaths` builds its list from
the messages a session still holds. By the time the chat is deleted, nothing names
that file any more. Orphan forever, not orphan until.

Only manual message deletion does it. Regenerate and edit-and-resend carry the
images to the new message with the same path, and a fork shares the file with its
source by design.

My preference is to fix it here rather than leave a loose end behind a change whose
whole point is that attachments are accounted for. A file nobody can name again is
not something a later PR is likely to come back for, and the fix is the same "minus
what a surviving session still names" logic scoped to one message, so it reuses
what's there.

The argument against is yours: it adds a second removal path, and one removal path
is the property you liked. Your call, and I'm happy either way.

lojza3d:
Pushed. Two places where I made a call rather than wait on a round trip. Both are one line to reverse.

## PNG vs JPEG

You said transcode everything else to PNG. It is **JPEG q0.85, PNG only when the bitmap has alpha**, and I would rather flag that than let it pass.

What lands here is a photo, since HEIC is the iPhone default, and a lossless copy of a 12 Mpx frame runs to tens of MB while preserving the artifacts of a lossy source with perfect fidelity. 0.85 is the quality this app re-encoded every attachment at before they moved to disk, so nothing gets worse than what ships today, it just applies to far fewer inputs now.

Measured on a real iPhone photo: 1.2 MB HEIC in, 2.4 MB at q0.95, 2.1 MB at q0.85. Any JPEG of a 12 Mpx frame will beat the HEIC it came from, HEVC intra being a generation ahead. For a working copy the tower is about to shrink to a fifth of its size that seemed like the right trade, but say the word for straight PNG or a lower quality.

The gate follows your shape: `sniffedExt` answers nil instead of guessing, whitelist is exactly `decodeRgbOwned` (WebP was missing, so it went in), and the stored extension describes the stored bytes rather than the incoming name, so a converted `IMG_1234.HEIC` lands as `.jpg`. One thing came out of testing: the bitmap now comes from the `CGImage`, not `tiffRepresentation`, which re-encodes and can hand back an alpha channel the picture never had - that would have sent every photo down the PNG branch.

## Orphans

Fixed here too, and it was worse than it looked. "Until the chat is deleted" is optimistic: `deleteMessage` drops the message, the path goes with it, and `removablePaths` builds its list from the messages a session still holds. By the time the chat is deleted nothing names that file. Orphan forever, not orphan until.

Same "minus what a surviving session still names" logic scoped to one message, so a fork that copied that message still spares the file. It does add the second removal path you liked not having, so it is pinned: a source scan asserts there are exactly two and that `truncateMessages` is not one of them. That part is load-bearing - regenerate and edit-and-resend truncate and then rebuild the turn from the same `ChatImage` values, paths included, so a removal there would delete a file about to be used again.

## Tests

Full suite green, 3165. New: each of the four formats passes through byte-identical, HEIC converts to JPEG, a transparent source converts to PNG, the extension follows the stored bytes, the sniff declines HEIC/TIFF/BMP/empty instead of calling them PNG, a message delete removes its own file but spares a fork's, plus the two source scans.

By hand: an iPhone HEIC lands as `.jpg` and the model describes it correctly; a dropped PNG is still byte-identical by shasum; deleting a message removes exactly its own file; deleting a forked message leaves the original alone; regenerate touches neither the transcript nor the folder.

PR #318, opened 2026-08-30
fix(vision): select media from the active turn

IridiumMaster:
## Summary

- Select the newest media-bearing user message in the active turn, so trailing user-role context injections do not hide an attachment.
- Use that same selected message for vision encoding, exact multimodal-token placement, Qwen M-RoPE construction, and LFM2 image segments.
- Preserve assistant-prefix continuation and assistant tool-call/result flows while treating ordinary assistant replies as durable turn boundaries.

## Problem

Agent clients can append project or system context as additional `user` messages after the human message. The vision path previously selected only the final user message. A request shaped like `user(image), user(context)` therefore decoded the image during JSON parsing but never encoded or inserted it into the prompt.

This was reproduced through DeepSeek Harness with Qwen3.8 Flash Next: the uploaded image was present in the request, but mlx-serve selected the trailing context message and generated a text-only prompt.

## Validation

- `zig build test -Dtest-filter=activeTurnMediaMessage`
- `zig build test -Dtest-filter=insertMultimodalTokens`
- `zig build test`
- `bash -n tests/test_vision_prefix_cache.sh`
- Live Qwen3.8 Flash Next regression: the Harness-shaped request processed one image, inserted 484 vision tokens at the owning user turn, built Qwen M-RoPE, and correctly read `Grey Fox Tr / Waterfall Dr`.
- Text-only continuation did not re-encode the historical image.
- Autoreview reported no actionable findings.

## Scope

This change deliberately selects new media only from the active turn. Reconstructing historical multimodal placeholders on a later text-only turn is a separate pre-existing multi-message concern and is not claimed here.


ddalcu:
thanks for your PR's.  In the future, if they are modifying same files / somewhat related, feel free to keep it in the same PR.  (as long as its not huge)

Would be easier for both of us, I had to checkout each one, test, solve conflicts, etc :)

PR #320, opened 2026-08-31
perf(vision): decode only active-turn media

IridiumMaster:
## Summary

- select the active multimodal wire message before parsing OpenAI or Anthropic requests
- decode image, video, and audio payloads only for that active message
- ignore OpenAI wire messages the parser will certainly skip, so the gate remains a conservative superset of parsing
- preserve historical media in the text prompt, including image-only user role boundaries
- cover long image histories, injected context, assistant continuations, and Anthropic tool chains

## Root cause

The server previously parsed every message before choosing which media belonged to the active turn. Parsing eagerly decoded every historical data URL, so a text-only continuation repeatedly paid base64, JPEG, resize, and patchification costs for media that would not be sent through the vision tower.

This change performs a metadata-only scan of the raw wire messages first. The existing prompt history remains intact, but heavyweight media decoding is limited to the selected active turn.

## Validation

- `./.zig-toolchain/zig build test`
- `./.zig-toolchain/zig build -Doptimize=ReleaseFast`
- `VISION_CACHE_MODEL='/Volumes/CorbanuDrive/Models/DeepSeek-Harness/Qwen3.8-Flash-Next-Uncensored-MLX-Serve-mixed-4-8bit' MLX_SERVE_BIN='./zig-out/bin/mlx-serve' bash tests/test_vision_prefix_cache.sh` — 31/31 assertions passed, including empty-assistant and trailing-empty-user parser gaps
- structured local review (`autoreview`) — no actionable findings after fixing its image-only-history finding
- signed ReleaseFast binary tested through DeepSeek Harness with Qwen 3.8 Flash Next Uncensored mixed 4/8-bit:
  - image turn decoded one 720×720 image and correctly read both street signs
  - text-only continuation returned successfully with `Decoded .* image` count unchanged (3762 → 3762)
  - continuation reused 7,168 prompt tokens and produced no Harness connection or browser error

## Documentation

- records the active-turn media invariant in `CLAUDE.md`
- documents why HTTP adapters must choose active media before decoding wire payloads


ddalcu:
@IridiumMaster on a roll ! Thank you and Nice work, this is close. One thing I want fixed before merge:

activeWireMediaIndex breaks on an empty assistant message, but the parse loop skips those, so activeTurnMediaMessage never sees it as a boundary. Send [user(image), assistant(content:""), user("hi")] and the old code re-encoded the image as active turn media. Now the wire gate calls it historical, the image never gets decoded, and the parsed selector finds nothing. Model answers blind with a 200. Worse, the rendered prompt has no assistant marker either (the empty message got skipped), so the prompt says the image is in-turn while the pixels are gone. Same rule as the MiniCPM ladder: the gate has to be a superset of the parser, never narrower. 

Fix should be small: treat an assistant the parse loop will skip (no text, no tool_calls, no reasoning) as invisible in the wire selector instead of a boundary, and have wireContinuationRequested skip messages the parse loop skips too (a trailing empty user flips continuation detection the same way). Plus a regression test for the empty-assistant shape.

The cases where the wire gate is wider than the parser (malformed tool_calls, text+tool_calls continuation) just waste a decode, output is identical, those are fine as is.

Everything else looks good. Anthropic side is unaffected since empty assistants get appended there.

IridiumMaster:
Fixed in 215af12 — the report is valid.

- OpenAI wire selection now ignores only messages its parse loop certainly skips (empty content, no media, no non-empty tool_calls, no reasoning), keeping the wire gate a conservative superset of parsing.
- `wireContinuationRequested` and `activeWireMediaIndex` use the same visibility rule; Anthropic behavior is unchanged.
- Added unit regressions for `[user(image), assistant(content:""), user("hi")]` and a trailing empty user after an assistant-prefix continuation.
- Added live HTTP assertions proving both shapes decode one image and read the sign text.

Validation: full `zig build test` passes; ReleaseFast builds; Qwen 3.8 Flash Next vision integration passes 31/31; autoreview reports no actionable findings. Rebased on current `main` before pushing.

ddalcu:
Thank you !  Awesome fix... I was experiencing this yesterday with Qwen Next a lot when it was "self validating" by taking screenshots in chrome all the time

PR #323, opened 2026-08-31
feat(qwen4_exp): YaRN rope scaling to 1M context

beamivalice:
Extend `Qwen3.8-Flash-Next` (`qwen4_exp`, `head_dim 256`, `partial_rotary_factor 0.25`, `rotary_dim 64`) from 262144 to 1048576 via HF/vLLM-compatible YaRN.

Matches `transformers` `modeling_rope_utils._compute_yarn_parameters` and vLLM `YaRNScalingRotaryEmbedding` (same `factor=4`, `original_max_position_embeddings=262144`, `rope_theta=10000000`, `mrope_interleaved=true`, `mrope_section=[11,11,10]`, `beta_fast=32`/`beta_slow=1`/`truncate=true`, `attention_factor=0.1*ln(f)+1`).

### Config
- `src/model.zig`: parse flat `text_config.rope_parameters` with `rope_type=yarn` for `qwen4_exp`/`qwen4_exp_text` (`parseYarnRopeParameters`), derive `factor` when omitted, honor `attention_factor`/`attn_factor`, expose `yarnPartial`/`yarnRotaryDim`/`contextCap` (`orig*factor`), wire `try parseYarnRopeParameters` in `qwen4_exp` arm.
- `src/model.zig` / `src/main.zig`: `--config-overrides` / `--hf-overrides` deep-merge (arena, vLLM `_apply_dict_overrides` semantics) applied in `parseConfigFromJson` to every load/discovery stub; validated at flag time.

### RoPE
- `src/mrope.zig`: `Yarn` (`correctionRange`/`invFreq`/`attentionFactor`/`contextLen`), `fillCosSin(...,mscale)` folds `cos/sin*mscale`.
- `src/transformer.zig`: single `mrope.Yarn` spectrum precomputed at load (`rope_freqs_yarn` denominator + `yarn_inv_freq` host + `yarn_mscale [head_dim]`), `yarnSpec`/`yarnMscale`/`yarnActive`, `computeYarnFreqs` delegates to `mrope.Yarn`, `mropeCosSinAt`, `ropeAtPositions`/`ropeAtFreqs`, `gatedFullAttnWith` (fused 128 `ms` + fused 256 `mscale` in angle rows + composed dynamic/scalar), QSA indexer `q_rope` + pooled keys, `qkAngleRowsFor` `mscale`, `yarnScaleQK`.

### Sizing
- `src/server.zig`: `autoContextFor`/`computeMaxSafeContext` and logs use `contextCap()` (`max_position_embeddings` or `orig*factor`), same derivation as vLLM `max_model_len`.

### Tests
- `src/mrope.zig`: hermetic YaRN goldens vs HF/vLLM (f64 1e-12, HF f32 1e-6), band/mscale/1M→262k and env-gated oracle `QWEN4_YARN_FIXTURES` (generated by `tests/dump_qwen4_yarn_fixtures.py` via real HF + vLLM transcription).
- `src/model.zig`: `QWEN4_SHIPPED`/`QWEN4_YARN` literals, mscale/beta/truncate, derived factor, error paths, `contextCap` clamp, `--config-overrides` merge vs replace.
- `src/transformer.zig`: device denominator vs host `inv_freq` reciprocity.

### Verification
- `zig fmt --check` / `zig build -Doptimize=ReleaseFast` / `zig build test` (1845 pass, 149 skip after transformer invariant fix).
- Real pack `Qwen3.8-Flash-Next-MLX-Serve-mixed-4-8bit` extended externally to `max_position_embeddings 1048576` `rope_type yarn`; boot shows `[rope] YaRN: factor 4.0000 (window 262144 -> 1048576), rotary 64 dims, ramp [14, 22], mscale 1.138629` and `context_length 1048576`; 275054-token prompt returns planted `ARCHIVE-QUASAR-9471` with HTTP 200 at 1045 tok/s prefill.


ddalcu:
Impressive work, the rope math itself checks out. I ran the merged tree locally: ReleaseFast build clean, full zig build test green, goldens match HF, and I verified the factor-1.0 path is bit-identical so unscaled checkpoints can't regress. Also checked contextCap against the real dsv4/laguna/gpt_oss configs, orig*factor equals declared max on all three, so no sizing shift there.

Problems:

1. QSA indexer mscale is inconsistent across arms. mropeCosSinAt folds the mscale into the shared tables, so on image turns the indexer q and the pooled block keys get scaled, while the scalar arms deliberately skip it. The comment justifying the skip ("one positive constant cannot change a top-k") doesn't hold for this geometry: indexer_head_dim is 128 with only 64 rotated, so score = ms²·A + B vs A + B and blocks do rerank. Worse, qsa_pooled is cached and rides the prefix cache, so a text turn followed by an image turn scores scaled and unscaled blocks against each other on one ladder. Pick one convention everywhere (check what the reference indexer actually does with the scaled cos/sin), and pin it with a test.
2. The SSD KV cache will restore rotations from the wrong spectrum. modelFingerprint is model dir + config.json size/mtime, and --config-overrides isn't in it. Flip the override between boots, which is the flag's whole A/B purpose, and a --prefix-cache-disk restore hands back K rows rotated under the other spectrum, silently. You already export getConfigOverrides() and nothing calls it, fold it into the fingerprint.
3. tests/dump_qwen4_yarn_fixtures.py is referenced three times but isn't in the PR. Without it the QWEN4_YARN_FIXTURES oracle can't be regenerated. Please add it.

Two smaller notes, your call whether to address here: vLLM's attn_factor is a multiplier on the computed mscale, not a replacement (HF's attention_factor is the one that replaces), so the two keys shouldn't be read the same way. And --config-overrides applies to every model this server loads and discovers, not just the one you're scaling; that's fine for keys only qwen4 reads, but a generic key like max_position_embeddings would hit everything. At minimum the help text should say so.

Also add a CHANGELOG line and the new flag to the CLI list in CLAUDE.md.

beamivalice:
The skip assumed a constant cannot change top-k, now it is fixed and tested for correctness.

---

Grok 4.6's summary.

Here only half the indexer dims rotate, so the scale *does* move which blocks win. Image turns already used the scaled tables; text and single-token decode did not, and those two got mixed in the cache. We now use the same scale the reference uses, fingerprint the override so an A/B boot cannot restore the wrong rotations, and treat HF vs vLLM keys as they actually mean.

- QSA: one effective mscale. M-RoPE already had it; scalar `mlx_fast_rope` / `ropeAtFreqs` now scale the rotated 64 of 128. No second multiply on `applyMrope`.
- `qsa_pooled` stays cached; both arms write the same convention.
- `modelFingerprint` hashes `--config-overrides`. Wipe `~/.mlx-serve/kv-cache` if you already served with the flag.
- HF `attention_factor` replaces; vLLM `attn_factor` multiplies. Help text: the JSON merges into every loaded/discovered `config.json`.
- `tests/dump_qwen4_yarn_fixtures.py` is in the tree. CHANGELOG + CLI list updated.

beamivalice:
I tested with 300k needle in a haystack - pass with 1122 tok/s prefill but 15 tok/s decode.

=== log tail ===
[rope] YaRN: factor 4.0000 (window 262144 -> 1048576), rotary 64 dims, ramp [14, 22], mscale 1.138629
[nax-sdpa] engaged: stock fused sdpa (force_fused) serves hd-256 causal prefill and verify blocks > 8 rows; msv_attn_p256 causal arm declined (MLX_SERVE_NAX_SDPA=0 restores)
Prefill chunk: 4096 tokens (memory-sized down from 8192; --prefill-chunk overrides)
Context size: 1048576 tokens (manual)
Model context length: 1048576 tokens
  prompt=300046 tokens, max_gen=32, ctx=1048576
[qsa] sparse attention engaged: kv=4096 blocks=1024 top-512 (budget 2048, ratio 4)
  <- 300046+11 tokens (268119ms) [prefill: 1122.4 tok/s, decode: 14.8 tok/s] [stop]

beamivalice:
Some numbers. My machine oom after 384 / kv 8.

Screenshot 2569-09-01 at 17 34 30

ddalcu:
Thanks, nice work. Verified the fix on the merged tree: mscale convention is consistent across both arms now (checked every rope call site for double-scaling, found none), fingerprint rolls with the override. Ran all the new yarn tests ran on GPU.

Also benched main vs this branch on the unscaled Flash-Next pack (M4 Max, 512/4k/8k, 3 runs per rung, counterbalanced boots): decode and prefill within 0.2% everywhere, so no regression on the default path.

LGTM, merging.

PR #325, opened 2026-08-31
feat: make repetition loop cutoff configurable

ViRb3:
I have a specific workflow that involves byte patterns. During the thinking process, models will correctly rehearse these patterns, and correctly repeat upwards of 20 '??' pairs. This seems to trip the repetition loop guard in mlx-serve.

This PR makes it configurable in the GUI and CLI, so people can adjust it for their use case. The default is untouched.


ViRb3:
Hi @ddalcu can we please upstream this so I can stop maintaining my own fork? Many thanks!

ddalcu:

- `src/generate.zig:8711` — `--loop-repetitions N` only raises the *short* tier, but the long tier (`exactCyclePeriod(tokens, 9, 64, 10)`) still convicts any ≤8-token cycle once its tail spans ≥ 90 tokens (a run of period-p tokens is also periodic at every multiple of p in 9..64), so the configured threshold is silently capped at roughly `90/p` repetitions.
  Failure: `--loop-repetitions 64` (a Settings preset) with a 1-token cycle: `[5,6,7] + 100×404` skips the short tier (span 6400 > len) but `p=9, reps=10` matches the last 90 tokens → cut as `.long_cycle` after ~90 repeats instead of 64… and worse, for an 8-token cycle the long tier (p=16 ×10 = 160 tokens) fires at 20 repeats, so presets 24/32/48/64 have no effect at all for the cycle lengths the flag is documented to govern. The new tests only exercise 20–32 reps of a 1-token cycle, which stays under the 90-token long-tier span. Either scale the long tier's `reps` off the configured value (e.g. skip the long tier for periods that are multiples of a short period, or require `p*reps >= max_period*exact_reps`), or document/clamp the reachable range.

ViRb3:
@ddalcu rebased and addressed comments.

ddalcu:
🤖 **MLXServe-PRBot approved** — reviewed at head `308a86ce`, no issues found.

The part that could have gone wrong is the new short-cycle exclusion in the long tier, and it is sound: the long tier's minimum period is `degenerate_loop_max_period + 1` (9), so `short_p` (1..8) can never equal `p` and the `cycle[short_p..] == cycle[0 .. p - short_p]` periodicity test can never degenerate into comparing two empty slices and disabling the tier wholesale. It is also inert at the default: for the exclusion to be reached the short tier must decline while the long tier matches, which needs at least `10 * p / short_p` (≥ 30) copies of the short cycle but fewer than 16 — impossible at `degenerate_loop_reps = 16`, so existing behaviour is unchanged and only a raised threshold relaxes the guard, which is the point.

The wiring holds end to end: `scheduler.zig:6739` is the only production caller of the guard, it goes through `loopStopDecision` → `loop_repetitions`, and the default `degenerateTail` wrapper is left with tests as its only callers, so there is no second threshold to drift. The Swift side mirrors the Zig default (16), emits `--loop-repetitions` only when it differs, adds the `SettingsReset` entry the tripwire test requires, decodes tolerantly, and carries an explainer; `main.zig` parses the flag and rejects `1` by name; `docs/reference.md:45`'s flag list is updated (the CLAUDE.md link to `docs/model-rules.md` is a pre-existing dangling reference — that file is absent on `main` too, so it is not this PR's to fix).

Note only: `build-test` is still pending on this head, so nothing here reflects a green suite.

ViRb3:
Addressed Copilot comments.

ddalcu:
🤖 **MLXServe-PRBot approved** — reviewed at head `0d1c3204`, no issues found.

The increment closes the plumbing class the repo names for CLI flags. `serve()` logs the resolved value straight from `scheduler_mod.loop_repetitions` — the same variable the guard reads — and `test_headless_spec_flags.sh` greps that line for the default 16, for `--loop-repetitions 64` and for the disabled arm, so a flag that parses but never reaches the scheduler now fails a test instead of shipping silently. The flag is in both lists (`CLAUDE.md` and `docs/cli.md`), with the `>= 2` constraint and the `0` escape hatch stated where a user will read them.

The detector analysis from the previous head is unchanged: the long tier's minimum period is 9, so the short-cycle exclusion can never compare two empty slices, and reaching it requires at least 30 copies of the short cycle while the short tier has seen fewer than 16 — impossible at the default, so existing behaviour is untouched.

ddalcu:
🤖 **MLXServe-PRBot approved** — reviewed at head `71645b9d`, no issues found.

Re-reviewed after the rebase onto current `main`; the PR's own patch is 199 lines across 13 files.

The plumbing has exactly one serving entry point. `scheduler.zig:6770` and `loopStopReason:5165` both route through `loopStopDecision`, which now forwards to the configurable form, and the only remaining callers of the fixed-threshold `degenerateTail` are tests — so the flag cannot be bypassed on a live request. `loop_repetitions` is written once in `main()` before any thread starts, so the process-wide global is not a race, and it is initialized from `generate_mod.degenerate_loop_reps` rather than a literal, so the Zig default cannot drift from the constant. On the Swift side the default is a literal but `ServerOptionsTests` pins it with the constant named in the assertion, the preset list makes `1` unselectable, and `toCLIArgs` suppression, `serverLaunchEquals` and the section reset each got coverage.

The interesting part is the new fall-through in `degenerateTailWithExactReps`, and it holds. `short_p` ranges 1..8 while the long tier's `p` starts at 9, so `cycle[short_p..]` and `cycle[0 .. p - short_p]` are always the same non-empty length — no slice can go out of bounds. More to the point, the `return null` is unreachable at the default: a tail that is a sub-period cycle long enough for the long tier (10 reps of a 9..64 period) has always already cleared 16 reps of the short cycle, so the short tier convicts first and the guard only bites once the threshold is raised above what the short cycle has reached. That is exactly its stated purpose, and skipping the fuzzy tier along with it is deliberate and documented rather than an oversight. The three new unit tests cover it from both directions — every period 1..8 against every preset, a 2048-rep threshold that the near-repeat tier would otherwise cap, and genuine 9..64 periods keeping their own ten-repeat bar.

`--loop-repetitions` is in `main.zig`'s match list and in the `CLAUDE.md` flag list, so it is not flag-eater material; `classifyUnparsedArg` is shape-based and needs no registration. Rejecting `1` by name is right — `exactCyclePeriod` with one repetition would convict any trailing token.

ddalcu:
@ViRb3 in latest main I essentially removed this all together, replaced with a smarter detection method. If it's still a problem we can re-open, rebase this and update it.

PR #328, opened 2026-08-31
ui: add remaining space to model browser view in my models

justinluque:
This pull request enhances the `MyModelsPane` view by displaying the amount of free disk space available to the user. The implementation includes tracking and formatting disk space, and updating the view when it appears.

**User Interface Improvements:**
- The available free disk space is now shown in the UI as a caption, making it easy for users to see how much storage is left.
- The disk space is refreshed when the view appears, ensuring the information is up to date.

**Follow-up fix:** the free-space label now reuses `MemoryInfo.format(_:)` (one-decimal "X.X GB", same as the rest of the pane) instead of a separate `ByteCountFormatter`, and reads the volume capacity via `try?` to match the existing pattern in `DownloadManager.swift`.

## Testing

- `cd app && swift build` — builds clean, verified with `bash app/build.sh`.
Before:

Before:

image

After:

image

### DIFF
--- a/app/Sources/MLXServe/Views/ModelBrowserMetrics.swift
+++ b/app/Sources/MLXServe/Views/ModelBrowserMetrics.swift
@@ -92,4 +92,14 @@ enum ModelBrowserMetrics {
     static var minDetailWidth: CGFloat {
         fixedWidth(for: .compact) + compactModelColumnMinWidth
     }
+
+    /// Formats a volume's available capacity for the My Models footer, or
+    /// `""` when the value couldn't be read (`resourceValues` failed). Split
+    /// out of `MyModelsPane` so the formatting is testable without a real
+    /// volume: the caller resolves `nil`/a byte count off whichever URL is
+    /// the actual download destination, this just renders it.
+    static func freeSpaceLabel(availableBytes: Int64?) -> String {
+        guard let availableBytes else { return "" }
+        return MemoryInfo.format(availableBytes)
+    }
 }
--- a/app/Sources/MLXServe/Views/ModelBrowserView.swift
+++ b/app/Sources/MLXServe/Views/ModelBrowserView.swift
@@ -714,6 +714,7 @@ private struct DiscoverPane: View {
 private struct MyModelsPane: View {
     @Binding var filter: String
     @EnvironmentObject var appState: AppState
+    @EnvironmentObject var downloads: DownloadManager
 
     @State private var freeDiskSpace: String = ""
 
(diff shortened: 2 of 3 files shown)


ddalcu:

- `app/Sources/MLXServe/Views/ModelBrowserView.swift:812` — Free space is read from the home volume (`NSHomeDirectory()`) instead of the configurable models download root, so the footer reports the wrong disk whenever the download folder is on another volume.
  Failure: user sets the download folder (Settings → chooseDownloadFolder) to an external 2 TB drive while the boot disk has 12 GB free — footer says "12.0 GB available" while `DownloadManager` correctly pre-checks against `destDir`'s volume and 50 GB downloads succeed (or the inverse). `MyModelsPane` can take `@EnvironmentObject var downloads: DownloadManager` like 8 sibling panes in this file and read `URL(fileURLWithPath: downloads.modelsDir)`.

- `app/Sources/MLXServe/Views/ModelBrowserView.swift:806` — `freeDiskSpace` is only computed in `.onAppear`, so it goes stale after a delete or a finished download while the pane stays mounted, while the adjacent "N models on disk" count updates live.
  Failure: footer shows "3 models on disk · 20.0 GB available"; user trashes a 15 GB model → `LocalModelRow.performDelete()` refreshes `localModels` so the left text becomes "2 models" but the right still says "20.0 GB" until the pane is torn down. Add `.onChange(of: appState.localModels.count) { updateDiskSpace() }` next to the existing `onAppear`.

- `app/Sources/MLXServe/Views/ModelBrowserView.swift:811` — Ships a user-visible feature with no test, against root CLAUDE.md "Testing — TDD is mandatory" ("UI/build scripts = factor a pure helper and test that").
  Failure: the diff touches only `ModelBrowserView.swift`, nothing under `app/Tests`. A pure `freeSpaceLabel(forVolumeAt:)`-style helper with a test would have caught the wrong-volume bug above and naturally fixes the refresh issue too.

justinluque:
Appreciate the in-depth reviews, will begin work on these !

justinluque:
Addressed all three:

- Free space now reads `DownloadManager.modelsDir` (the actual download destination) instead of the home volume.
- The footer refreshes on model-count changes and on `modelsDir` changes, not just `.onAppear`.
- Split out `ModelBrowserMetrics.freeSpaceLabel(availableBytes:)` as a pure, testable helper (`testFreeSpaceLabelIsEmptyWhenUnreadable`, `testFreeSpaceLabelFormatsBytesLikeMemoryInfo`).

ddalcu:
🤖 **MLXServe-PRBot approved** — reviewed at head `5d02677`, no issues found.

PR #333, opened 2026-09-01
fix(reasoning): defer JSON grammar until final answer

### LINKED ISSUE #331: JSON schema output is returned as reasoning when Qwen thinking is enabled
### Priority

medium (workaround exists)

### Area

Server / API (chat, completions, messages, responses, embeddings)

### mlx-serve version

26.8.10 / current main at `cd93a2b`

### Hardware + macOS

MacBook Pro, Apple M4 Max, 64 GB, macOS 26.6.2

### Model

`ddalcu/Qwen3.8-27B-MLX-Serve-6bit`

Default settings. Speculative decoding is not used for the constrained request.

### What happened vs what you expected

When thinking and JSON-schema output are enabled together, the final JSON is returned in `reasoning_content` and `content` is empty.

Example of the incorrect result:

```json
{
  "reasoning_content": "{\"answer\":\"17\"}",
  "content": ""
}
```

I expected real arithmetic reasoning in `reasoning_content` and the schema-valid final answer in `content`:

```json
{
  "reasoning_content": "17 - 6 is 11, then adding 3 groups of 2 gives 17.",
  "content": "{\"answer\":\"17\"}"
}
```

The same problem affects streaming: the JSON arrives through `delta.reasoning_content`, with no final JSON in `delta.content`.

The Qwen chat template ends the rendered prompt inside a bare `<think>` block. mlx-serve applies the JSON grammar to the first generated token. This prevents the model from generating `</think>`, because that token is not valid JSON.

The model therefore generates the only allowed output—the JSON object—while it is still inside the reasoning block. The response parser then treats that JSON as reasoning.

Turning thinking off avoids the problem, but it removes reasoning from requests that need both reasoning and structured output.

### Reproduction steps

1. Start mlx-serve:

```bash
./zig-out/bin/mlx-serve serve \
  --model /path/to/ddalcu/Qwen3.8-27B-MLX-Serve-6bit \
  --port 8080 \
  --log-level debug
```

2. Send this request:

```bash
curl http://127.0.0.1:8080/v1/chat/completions \
  -H 'Content-Type: application/json' \
  -d '{
    "model": "mlx-serve",
    "messages": [{
      "role": "user",
      "content": "A basket starts with 17 apples. Six are removed, then three groups of two apples are added. Think through the arithmetic briefly and give the final count."
    }],
    "reasoning_effort": "medium",
    "response_format": {
      "type": "json_schema",
      "json_schema": {
        "name": "answer",
        "strict": true,
        "schema": {
          "type": "object",
          "properties": {
            "answer": {"type": "string"}
          },
          "required": ["answer"],
          "additionalProperties": false
        }
      }
    },
    "temperature": 0,
    "max_completion_tokens": 2048,
    "stream": false
  }'
```

3. Observe that `content` is empty and the final JSON appears in `reasoning_content`.

4. Repeat with `"stream": true`. The JSON is emitted through `delta.reasoning_content` instead of `delta.content`.

Equivalent structured-output requests through Anthropic Messages and the Responses API use the same grammar setup.

### Server log tail

```text
POST /v1/chat/completions (... thinking=true ...)
[grammar] enforcing JSON schema (vocab=..., mask=...b)
```

There is no grammar error. The grammar is applied successfully, but it is applied too early—before the model has closed the reasoning block.

perretv:
Fixes #331

Qwen-style templates can start generation inside a bare `<think>` block. Applying the JSON grammar from token zero makes `</think>` unreachable, so JSON lands in reasoning and final content is empty. This change defers the grammar until the reasoning boundary on supported requests.

## Behavior and scope

- With thinking enabled, a rendered prompt ending in bare `<think>` and an atomic `</think>` token allow unconstrained reasoning, followed by JSON-constrained final-answer generation. The grammar activates after the close token, before the next sample.
- With thinking disabled, JSON constraints still apply from token zero.
- Unsupported prompt/tokenizer boundaries retain the thinking-off, token-zero fallback. Finite response-side reasoning budgets on Chat Completions and Anthropic Messages also retain that fallback, avoiding a parser cutoff before grammar activation. Existing Responses budget behavior is unchanged.
- A repetition loop, premature EOS, or terminal padding forces the configured close token when the remaining completion allowance can fit the close and at least one constrained answer token. Recovery uses the live KV/SSM state and counts and publishes the close token normally.

The hard completion limit still includes reasoning and final-answer tokens. If reasoning consumes the entire allowance, empty final content with `finish_reason: "length"` is expected. There is no answer-token reserve and the server does not exceed the limit to finish JSON. A limit reached during final-answer generation can likewise truncate JSON. Reasoning is never moved into final content to repair a truncated response.

This PR supports only the Qwen-style bare-tag boundary. Hy3 suffixed tags, Muse channels, LFM-style prompts, and other reasoning grammars retain the fallback; extending their generation boundaries belongs in follow-up PRs. Tools retain the existing grammar-skip policy, and constrained requests remain excluded from speculative decoding. Template argument parsing, reasoning-budget aliases, prompt instructions, and global streaming-budget behavior are unchanged.

## Validation

The focused live test checks final-channel JSON, finite-budget fallback, deferred activation, and streaming reconstruction. New Chat Completions regressions use unlimited reasoning with `max_tokens: 64` and assert empty final content, nonempty reasoning, `finish_reason: "length"`, and exactly 64 completion tokens in both streaming and non-streaming responses. They do not grade arithmetic or exact reasoning prose.

- `test_json_schema_thinking.sh`, Qwen3.5-0.8B: 9/9 passed.
- `zig build test`: passed.
- `zig build -Doptimize=ReleaseFast`: passed.
- `swift build` in `app/`: passed.
- Shell syntax and `git diff --check`: passed.

Qwen3.8-27B validation was attempted but the server could not load the model because available memory was below its preflight requirement. No current Qwen3.8 result is claimed.


ddalcu:
The idea is fine. The patch is not.

1. You replaced "always returns JSON" with "returns JSON if the model feels like closing its think block". Qwen3.5-0.8B loops in reasoning, loop-stop cuts it, and the client gets content: "". Main returns JSON for the exact same request. A schema request that can return nothing is not a schema request. When reasoning gets cut, force </think> and arm the grammar. No exceptions.
3. You ripped out the thinking-off gate and its scan test, then wrote in the PR body that other families are "follow-up work". No. Hy3, muse and LFM2.5-VL have no bare atomic </think>, so they are now back to the token-0 mask inside an open think block, which is the bug this whole thing was supposed to fix. You dont delete a safety net because you built a better one for a third of the cases. Keep the fallback wherever deferral isnt possible.
4. chat_template_kwargs parsing, a thinking_budget_tokens alias that silently turns thinking on, and a change to how streaming budgets behave for every request. None of that is "defer JSON grammar until final answer". One PR, one change. Split it.
5. Your tests assert what the model happens to say. test_reasoning_json_schema.sh checks the arithmetic answer and dies on set -e the first time a 4-bit checkpoint gets it wrong, and the Responses arms fail because the model drafted the JSON in its reasoning while content was completely correct. Tests pin invariants. They dont grade the model.

perretv:
Thanks for your feedback! I was not completely sure how to approach this problem at first, so your comments gave me a much clearer direction.

I revised the draft to keep the change narrow and preserve existing behavior for models that do not use the supported Qwen-style reasoning boundary. I still think that support for other reasoning formats should come in follow-up PRs.

1. **Forced recovery when reasoning does not close**
   The grammar is deferred only while generation is inside a supported bare `<think>` block.
   - If the model emits `</think>` normally, the server commits it and activates the JSON grammar for the next token.
   - If reasoning gets stuck in the existing repetition-loop guard, reaches premature EOS, or ends in repeated padding before the close token, the server commits the configured `</think>` token, activates the grammar, and continues with constrained output.
   - This uses the existing KV and SSM state. It does not rebuild the prompt or reset generation.
   - I removed the earlier fixed 64-token reserve, this was probably too ad-hoc. Recovery only happens when there is room for the close token and at least one constrained answer token. Otherwise, the normal `max_tokens` length-stop behavior remains unchanged.

2. **Unsupported reasoning formats keep the old safe behavior**
   The deferred path is available only when both conditions are true:
   - The rendered prompt ends in a bare `<think>`.
   - The tokenizer has an atomic `</think>` token.
   If either condition is not true, schema requests re-render with thinking disabled and apply the JSON grammar from token zero.
   - Hy3, Muse, LFM-style prompts, and other unsupported reasoning formats therefore keep the existing safe path.
   - Tools still keep the existing grammar-skip behavior.

3. **Removed unrelated changes**
   I removed the unrelated work around:
   - `chat_template_kwargs`
   - `thinking_budget_tokens`
   - implicit thinking enablement
   - global streaming-budget behavior
   I also restored the original JSON-only prompt instructions. The behavior now comes from the grammar phase and response routing, rather than extra prompt wording intended to influence reasoning.

4. **Tests now check server behavior, not model answers**
   The tests no longer require arithmetic answers, particular reasoning text, or the absence of JSON-like text in reasoning.
   They now check that:
   - JSON reaches the final content or output-text channel.
   - The result parses and satisfies the required schema shape.
   - Streaming reconstructs final JSON from final-answer events.
   - Deferred grammar activation is logged, whether the boundary is natural or forced.
   - Unsupported formats use the thinking-off fallback.
   - Tools keep the grammar-skip behavior.
   - Constrained requests do not use speculative decoding.
   - A loop before activation recovers through the forced boundary, while a loop after activation keeps the normal repetition-loop result.

ddalcu:

- `src/server.zig:7541` — The streaming reasoning-budget exhaustion path (`think_tokens >= reasoning_budget` → flush `think_buf`, `in_think_block = false`) is not coordinated with the deferred grammar: the generator keeps sampling unmasked until it sees a natural `</think>`, the answer reserve, EOS/pad, or a loop stop, so once the parser has left the think block every further reasoning token is emitted as `delta.content`. Same shape on the `/v1/messages` stream at `src/server.zig:12834`.
  Failure: `POST /v1/chat/completions` on Qwen3.5 with `response_format: {type: json_schema, …}`, thinking on, `stream: true`, `reasoning_effort: "low"` (→ `effortBudget` 512 since the template doesn't consume effort). Policy resolves to `.deferred`; after 512 think tokens the stream flushes reasoning and flips to content, then the remaining reasoning prose and the literal `</think>` text arrive as `content` chunks ahead of the masked JSON — a content payload that isn't valid JSON, i.e. the #331 class this PR is fixing. Pre-PR `schemaMasksThinking` forced thinking off so this combination couldn't occur. Fix options: pass the budget into the generator so `forceDeferredConstraintBoundary` fires at the budget (mirroring the reserve/EOS paths), or route schema requests with a finite `reasoning_budget` to `.fallback_thinking_off`. `tests/test_json_schema_thinking.sh` has no arm with `reasoning_effort`/`reasoning_budget_tokens` set, so this stays green.

ddalcu:
🤖 **MLXServe-PRBot approved** — reviewed at head `2059d79c`, no issues found.

ddalcu:
🤖 **MLXServe-PRBot approved** — reviewed at head `6028318`, no issues found.

ddalcu:
🤖 **MLXServe-PRBot approved** — reviewed at head `50fe9fd`, no issues found.

perretv:
@ddalcu Would you mind having another look? I think I've followed your guidance regarding the implementation.
Let me know what you think of it!

ddalcu:
Thank you!

PR #336, opened 2026-09-01
RFC(qwen4): evaluate PLE pread worker width as a decode improvement

davidtai:
## Evaluation requested from the maintainer

**This is an experimental RFC for the mlxserve author to evaluate as a possible decode improvement. It is not presented as a finished optimization or a recommended new default.**

The question is narrow: **on each supported Apple-silicon and storage configuration, can a different width for the existing PLE pread pool improve end-to-end Qwen4 decode compared with the current fixed width of 48 workers?**

Please benchmark the candidate widths against the unchanged 48-worker default, then decide whether to keep this override, redesign it, or reject it.

## What changed

- Adds QWEN4_PLE_PREFETCH_WORKERS as a construction-time override for the existing native Qwen4 PLE pread pool.
- Keeps 48 workers as the default, so ordinary behavior is unchanged.
- Bounds experimental values to 1 through 48 and falls back to 48 for missing or invalid input.
- Sizes wakeup accounting, worker stride, and shutdown from the installed pool width.
- Reads the setting once when the n-gram table constructs its pool, never inside the gather hot path.
- Adds hermetic tests for default, invalid, and clamped values.

## What mlxserve has already proven

These are measurements of the existing optimization, not results produced by this PR:

| Existing decode implementation | PLE gather latency | Meaning |
| --- | ---: | --- |
| Serial random page faults through the 32 GiB mmap | about 5.0 ms/token | Original bottleneck |
| Shipping pool: 48 persistent workers using pread | about 0.45 ms/token | About 4.55 ms/token, or 91%, less PLE gather latency |

That establishes that native pread is valuable for decode. It does **not** establish that changing the shipping worker count from 48 is faster.

## What the author still needs to evaluate

Run matched full-request comparisons with QWEN4_PLE_PREFETCH_WORKERS set to candidate widths such as 8, 16, 32, and 48. The acceptance decision should use:

1. End-to-end decode tokens per second and wall time.
2. PLE gather latency, preferably p50 and p95.
3. Unchanged generated output and Qwen4 state behavior.
4. No prefill regression.
5. Repeatability across the supported machine and SSD configurations.

No such current-pread worker-width sweep is claimed in this PR.

## Known regression risks

Previous mmap page-fault experiments showed that thread topology matters: 16 newly spawned fault threads measured about 0.70 ms/token, a parked 16-thread fault pool measured about 0.85 ms/token, and 48 mmap-fault threads measured about 1.0 to 1.4 ms/token. Those are different mechanisms from the current persistent pread pool, so they are warnings, not direct evidence for a winning pread width.

Applying the worker route to a 4096-row prefill chunk measured roughly 4% slower. The existing decode-width gate is therefore unchanged, and this PR cannot opt large prefill gathers into the worker path.

## How to verify

    ./.zig-toolchain/zig test src/qwen4_exp.zig -O ReleaseFast

The focused suite passes 14/14 locally. No full-model benchmark was performed for this RFC.

## Notable decisions and tradeoffs

- Default behavior and maximum thread allocation remain exactly 48.
- The override is construction-bound rather than a per-gather branch.
- The decode-only row limit is untouched.
- This deliberately does not add a public CLI setting or choose a new default before the maintainer evaluation.

### DIFF
--- a/src/qwen4_exp.zig
+++ b/src/qwen4_exp.zig
@@ -179,7 +179,7 @@ pub const NgramTable = struct {
         if (8 + hlen > size) return error.NgramTableTruncated;
         var t = try parse(map, map[8 .. 8 + hlen], 8 + hlen);
         t.fd = fd;
-        if (plePrefetchEnabled()) t.pool = PrefetchPool.create() catch null;
+        if (plePrefetchEnabled()) t.pool = PrefetchPool.create(plePrefetchWorkerCount()) catch null;
         return t;
     }
 
@@ -322,7 +322,6 @@ pub const NgramTable = struct {
         };
         return std.c.pread(self.fd, dst.ptr, dst.len, @intCast(off)) == @as(isize, @intCast(dst.len));
     }
-
 };
 
 /// Persistent gather workers. Every row's three regions are one SSD read on
@@ -332,9 +331,10 @@ pub const NgramTable = struct {
 /// instead and dequantize their rows in place. Workers wake on a generation
 /// bump and count themselves down; the caller spins (the job is ~100 us).
 const PrefetchPool = struct {
-    const N = 48;
+    const MAX_WORKERS = 48;
     const MAX_ROWS = 64;
     const ROW_BUF = 512;
+    n: usize,
     mu: std.Io.Mutex = .init,
     cv: std.Io.Condition = .init,
     gen: u64 = 0,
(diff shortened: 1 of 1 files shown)


davidtai:
Closing this draft because I conflated two different datasets: the 64-66 tok/s receipts were MTPLX running an adaptation labeled mlxserve-ngram-pread, while the 0.45 ms PLE-gather result was a separate native mlxserve measurement. The proposed worker-width RFC therefore had no matched mlxserve end-to-end evidence and should not have been submitted.

PR #339, opened 2026-09-02
Let the reader choose the chat column, and lay a turn out around it

lojza3d:
Builds on the Interface settings from `527e855` and gives the reader the one knob that pane was missing: **how wide a conversation actually reads**. Narrow, Medium or Wide, from Settings or from F1-F3, applied live to the open chat.

Around that, a pass over the things a turn is made of. Your own message gets the four actions a reply has always had on its face, as icons on hover. Attached photos wrap into a proper grid instead of being squeezed into one strip. The reasoning block gets out of the way when collapsed, pulses while the model is actually thinking, and tells you afterwards how long that took. Table columns stop starving each other, so a header is never rendered as "Life spa n". And a reply finally uses the whole column it was given.

Tool-call rows are part of it too: a card is built from the structured record the model actually sent instead of from the summary string the engine had assembled to be read back, and a background process gets a badge that can say which process it stops. And the markdown renderer learned three things it never knew: numbered lists, blockquotes, and that inline code should look like code.

No behaviour outside the transcript changes. One new setting, one new optional field on `ChatMessage`, and a lot of layout.

## Before and after

Same conversation, same window width, same model output.

**One reply with the markdown blocks in it, a tool call and two reasoning blocks, all collapsed.**

Before:

screenshot_3608

After:

screenshot_3612

**The same reply with the reasoning and the tool call opened.**

Before:

screenshot_3609

After:

screenshot_3613

**Two background shells running, card collapsed.**

Before:

screenshot_3610

After:

screenshot_3614

**The same card opened, one stop button per process.**

Before:

screenshot_3611

After:

screenshot_3615

## Where this started

`527e855` gave the transcript a real reading rhythm, and one part of it didn't sit right at wide window sizes: the column is 80% of the window, and prose inside it wrapped early at an absolute `tailIndent` of ~45em. So one column had three different edges - prose stopped mid-column, a table beside it ran the full 80%, and a user bubble did neither. Widen the window and the gap grows.

That turned out to be a thread worth pulling, and most of this PR is what came off the end of it.

## The column is a setting

Settings ▸ Interface ▸ Chat Column: **Narrow (840pt), Medium (1260pt), Wide (the window, at `contentWidthFraction`)**.

Fixed points rather than a fraction, because a reading measure is a property of text: resizing the window then spends the MARGINS and leaves the lines where they are. Only once the window is narrower than the setting does it take over and wrap them. Transcript, composer and the empty-state greeting all read the one number, as before.

The same three widths are on **F1-F3** under View ▸ Interface, and Compact mode is on **⌃C** there too - both write the same `@AppStorage` keys the Settings rows do, so they are two doors onto one setting rather than two states that can disagree (test-pinned). Control rather than Option: holding Option over a transcript puts `NSTextView` into rectangular-selection mode and the key equivalent never arrives.

`CommandGroup(after: .sidebar)`, not `CommandMenu("View")` - the latter builds a second View menu beside the system one.

## Table columns get room for their own words

Narrowing the column exposed something that was always there: `columnFractions` weighted purely by how much text a column holds, so a short column beside a column of sentences starved. "Lifespan" got about 3% of the table and rendered as "Life spa n", because a word cannot wrap.

Content weight is now `sqrt`-compressed - ten times the text does not need ten times the width, since prose wraps onto more lines - and a column never asks for less than its longest WORD. That floor is capped at 18 characters, or one snake_case identifier, URL or hash would claim the table and squeeze every readable column into a sliver; past the cap the monster is broken mid-word, which is the better of two bad outcomes.

The grid also gained hairline rules between rows, an outline and a filled header. `NSTextTable` has no frame of its own, so the outline is four sets of cell borders, and only `.maxY` is ever set between rows - a neighbour's `.minY` lands on the same line and draws it twice as thick. No corner radius: AppKit's text tables have none.

## A reply IS the column

The assistant's `Spacer(minLength: 60)` is gone. A second margin inside a width the reader already chose is a margin on a margin, and it was what held the reasoning block's chevron short of the edge. Generated media and the progress card carry their own fixed widths (400/420), so nothing was leaning on it.

Your own turn keeps its spacer - that one separates the two sides of the conversation rather than trimming an edge - plus a cap of its own (700/900). Its attachments deliberately do NOT inherit that cap: a bubble is a reading measure, a row of photos wants room before it wraps, and tying them together let the measure decide how many pictures fit on a line.

## Attachments wrap

A plain `HStack` never wraps, so six photos in a message became slivers, and the strip then spanned the whole column while the text above it sat at a different edge.

They now wrap into rows of ONE height, packed to the trailing edge (`AttachmentFlowLayout`), with each width following its own picture's ratio - so a row shares a baseline and a landscape shot simply takes twice the room of a portrait one. Clamped both ways: a panorama could otherwise never share a row, and a sliver would disappear. `.fill` rather than `.fit`, because the rounded corners clip the FRAME - a letterboxed picture keeps square corners inside rounded empty space.

## The reasoning block

Hand-built rather than a `DisclosureGroup`, which pins its chevron to the leading edge with no way to move it, and hit-tests only that chevron - which left the "Thinking" text a dead click target.

The whole strip is one button. Padding and fill arrive WITH the content, so collapsed it is a line of type rather than a container and the chevron sits at the column's own edge. The brain pulses (`.symbolEffect(.pulse)`) only while the reply is still empty - once text starts, the thinking is over.

And the label now reports how long it took: "Thinking took 6 minutes 7 seconds". That needed one new optional field on `ChatMessage` (`thinkingSeconds`), measured once at the first content delta from the message's own timestamp - so it includes the prefill, which is time the reader spent watching the brain pulse. Absent from every older history, which simply reads "Thinking".

## Actions under your own turn

Copy, edit, branch and delete were context-menu-only on a user turn, while a reply has carried the same actions on its face all along. They are now icons on hover, 150ms fade, right-aligned.

Not drawn at all in compact mode - an invisible row that still holds its height gives nothing back - so the context menu stays the route there.

## Tool-call rows

A card used to render `call.content`, which the engine had just built as `**name**(key: value, key: value)` with every value cut at 80 characters. So the card stripped the bold markers back off a string assembled purely to be read back, and the panel could never show more than that string had already thrown away.

`SerializedToolCall` carries the name and the arguments as JSON, and it rides the assistant message that MADE the call - one message before the summary - so `ChatRowBuilder` carries it across. Argument order is recovered from the raw JSON, because a dictionary has none and alphabetical puts a file body above the path it was written to.

The card is built like the reasoning block for the same reasons, and it shows what the call was about without being opened: `writeFile · notes.md · 1053 chars`, `searchFiles · TODO · 3 occurrences in 2 files`. Those per-tool headlines are a LIST rather than a rule, so an unlisted tool shows nothing extra, which is what every tool did before - a new tool degrades to the old behaviour instead of to a wrong guess. `browse` is the exception that carries its action as part of its name (`browse:click`), since its seven actions share nothing but a browser. An MCP tool's `server__tool` is shown as `server/tool`, split by the function dispatch itself uses.

A round can hold several calls, and they run in ORDER, so the summaries pair with the calls by index - the result messages carry no id. A multi-tool card names the first two and counts the rest.

**Background processes.** `processHandles` is collected across the whole round and pinned to one message, so nothing in the model says which call started which process. Each result text names its own (`… as bg1 (pid …)`), which is enough to put each stop button beside the call it belongs to. A handle no call claimed keeps its badge in the header; a collapsed multi-tool card shows a button-less `running`, because with several calls one button in the header could not say which process it stops.

That turned up a real bug: the registry starts numbering at `bg1` on every launch and knows only the name, so a card from an old session asked "is bg1 alive?", was told yes about a process started today, and offered to kill it. The last row to announce a handle owns it now.

## The markdown renderer learns three blocks

**Numbered lists.** The parser stripped a list's numbers and the renderer drew a bullet for every item, so every ordered list arrived unordered and a "see step 3" above it pointed at nothing visible. `Block.listItem` had no slot for a marker, so there was nothing else the renderer could draw. It carries one now.

The detection went with it: "first character is a digit and a `. ` appears somewhere in the line" made a list of `1 pes. A kocka spolu.` and threw away everything up to the stop, leaving `• A kocka spolu.` It is anchored and shaped now (`^[0-9]{1,9}[.)] `, CommonMark's nine-digit limit), and a paragraph ends where a numbered item begins - only bullets broke it before, so a list written without a blank line above it was swallowed by the sentence.

**Blockquotes** were not handled at all: `>` fell through to a paragraph, marker and all. They are a barred block now, with consecutive quoted lines merged into one - a model wraps a quoted paragraph at its own width and puts a marker on every line, so a block per line would draw a stack of bars. The bar is a one-cell `NSTextTable` with a leading border, since an attributed string has no "rule beside this paragraph" attribute and that machinery was already here for tables.

**Inline code** had no ground of its own. It is found by `inlinePresentationIntent` rather than by the font's `.monoSpace` trait, which `NSFont.monospacedSystemFont` does not advertise - a trait check calls SF Mono prose and would be a branch that never fires. The tint is drawn by the text view rather than set as `.backgroundColor`, because that attribute fills the whole line fragment and at 1.4 leading it climbed into the descenders of the line above.

Two spacing findings came with them. A text table honours neither the paragraph spacing around it nor a paragraph style's indents, so a heading sat flush on the header rule however much air its own style asked for; air around a table or a quote is a cell MARGIN. And `blockSpacer` was not only the air between blocks but the only NEWLINE in the string, which is worth knowing before anyone tightens list rhythm the way I first did.

## Generated media in the transcript

A generated picture was drawn with `.fit` inside a `maxWidth`/`maxHeight` frame, so the frame was larger than the picture and the rounded corners were clipping the empty space beside it while the photo itself stayed square. The frame is the picture's own box now, from its own ratio, giving up height when the width cap bites - attachments solve the same trap by FILLING their frame, which crops, and that is fine for a thumbnail in a row but not for the picture the model was asked to make.

Rows are leading-aligned. A `.frame(maxWidth:)` with no `alignment:` centres, which is why every generated image read as indented from an edge the rest of the transcript shares.

In compact the preview shrinks and the prompt caption drops to one line. The Reveal button stays beside the caption rather than moving onto the preview: an `AVPlayerView` puts AirPlay and volume in its top corners and the transport bar across the bottom, so a player has no free corner.

## The footer under a reply

One left-aligned strip: when it was written, what you can do to it, how fast it came. Pinned to the right they were stranded beside nothing, now that a reply takes the column.

The two facts are pills that open on hover, and the full value floats over the pointer rather than growing in place - a pill that widens pushes the buttons beside it sideways, and a button that moves as you approach it is a button you miss.

The actions stay visible in both densities here, unlike a user turn's row, because this one carries Regenerate and Continue and they have no other home on the face of the transcript.

The gap that ends a turn is keyed on the FOOTER rather than on the role. A turn is several rows - a reasoning-only message whose output was tool calls, the card, then the reply - so paying it per assistant row opened an unexplained hole under a bare thinking block.

Worth saying plainly, since it shapes what the buttons can mean: **there is no concept of a turn in the app**. Every message is its own row with its own footer, so a turn that emits prose, calls a tool and writes more prose carries two footers, and copy takes that one message rather than the answer as a whole. That is out of scope here, and I would rather leave it visible than paper over it with a second, quieter definition.

## Compact mode

Takes its density from the LEADING now, not from the bubble's own padding: pulling the padding in pressed the text against the bubble's edge, which reads as a rendering fault rather than as density. Between-turn spacing came down too, since every turn now carries its own trailing band (a reply's footer, a user turn's action row) and a wide gap on top of that reads as a hole.

Prose leads tighter in compact as well, which it did not before - only your own bubble did. One trap in there: everything baked into a rendered string rides the render cache's KEY, and the cache was keyed on theme and text size only, so the first version of this looked like ⌃C did nothing at all to a transcript already on screen.

## Tests

`swift build` clean, full suite green: 3284 tests, 0 failures, on the branch with `main` merged in. (Before that merge, `HFSearchIntegrationTests.testSafetensorsExpand_PopulatesSizeData` was failing against the live HF API on this branch; v26.9.1's repo-size fix is what settled it, and nothing here touches that path.)

Around 60 new tests. The later ones pin the parser through the rendered string rather than through its own internals: that a numbered list keeps its numbers, that a sentence starting with a digit survives whole, that each item starts its own line (the regression I made while tightening list spacing), that a quote drops its marker and merges, that inline code is set in the code face. Two of them are written the way they are because my first version was wrong - a list-vs-paragraph assertion on the rendered TEXT cannot tell the two apart, since blocks are joined with a single newline either way, so it reads the hanging indent instead; and the code-face assertion names the font, because `.monoSpace` is not reported for the system mono font and a trait check would have passed on nothing.

`ProcessCardControlsTests` covers the handle-reuse bug directly: a reused name belongs to the newest row, ownership is per handle rather than per card, and a claim on a dead handle adds nothing.

The earlier tests are on the arithmetic and on the things that fail silently: the reading width and its edges (unmeasured panel, window narrower than the setting), table column shares (the starved-column case, the capped monster, no column squeezed to nothing), attachment geometry (ratio, both clamps, an image with no size), the duration label, and that `thinkingSeconds` survives a save - `ChatMessage` decodes by hand, which is exactly where a new key gets forgotten on one side.

Two structural guards worth naming: a row rule is never drawn twice, and there are no vertical rules between columns. Both are regressions that are easy to reintroduce and hard to see.

`VoiceOrbInlineTests.testThinkingHeaderTogglesTheAccordion` moved with the implementation: it pinned `DisclosureGroup` and a tap gesture, and now pins what those were FOR - one button, a hit-testable shape, and the chevron pushed to the trailing edge.

## Not in this PR

A real notion of a turn, which is what the footer's scope wants (see above). Nested lists, which the renderer still does not parse - an indented item falls through to a paragraph, as it did before. Both are their own change.


ddalcu:
haha, I knew I was opening a can of PR's with that new "Interface Settings".

Will test this once it's out of draft and ready + time permits

lojza3d:
A word about the size, because I know what this looks like on the Files tab.

I did think about splitting it. The reason it is one PR is that these pieces share one surface and mostly one set of numbers: the column width, the leading, the spacing between turns, and the widths of the things inside a turn all read from `ChatMetrics`, and several of the fixes only became visible once an earlier one landed. The table starving its own columns showed up when the column got narrower. The generated image reading as indented showed up once the reply stopped being inset. The list rhythm showed up once the leading tightened. Shipping them separately would have meant a run of releases where the transcript is half converted and looks worse than either end state.

Nothing here reaches outside the transcript: one new setting, one new optional field on `ChatMessage`, and layout. No API, no server, no protocol.

That said, it is your review time, not mine. If you would rather have this in pieces, say the word and I will split it: the markdown renderer (numbered lists, quotes, inline code), the tool-call cards, and the column plus everything laid out around it are three clean cuts, and I am happy to rebase them into that shape.

ddalcu:

- `app/Sources/MLXServe/Views/ChatView.swift:4638` — `handleOwnership` is computed over ONE session's messages, but `ProcessRegistry` handles (`bg1`, `bg2`…) are a single app-wide counter and `isAlive(handle:)` / `kill(handle:)` are not session-scoped, so the "last announcer owns the handle" rule only dedupes within a chat — the stale-card problem this PR sets out to fix still occurs across chats.
  Failure scenario: last week chat B's tool card announced `bg1`; today (fresh launch, counter reset) chat A starts a dev server that gets `bg1`. Opening chat B, its card is the last owner of `bg1` *in B*, `isAlive("bg1")` is true → the card shows a live "running" pill and its ✕ kills chat A's process. Fix: gate `killableHandles` (line 4754) on the registry entry's `sessionId` matching the chat (e.g. `processRegistry.list(sessionId:)`) rather than the global `isAlive`.

lojza3d:
> 🤖 **MLXServe-PRBot** — code review
> 
> * `app/Sources/MLXServe/Views/ChatView.swift:4638` — `handleOwnership` is computed over ONE session's messages, but `ProcessRegistry` handles (`bg1`, `bg2`…) are a single app-wide counter and `isAlive(handle:)` / `kill(handle:)` are not session-scoped, so the "last announcer owns the handle" rule only dedupes within a chat — the stale-card problem this PR sets out to fix still occurs across chats.
>   Failure scenario: last week chat B's tool card announced `bg1`; today (fresh launch, counter reset) chat A starts a dev server that gets `bg1`. Opening chat B, its card is the last owner of `bg1` _in B_, `isAlive("bg1")` is true → the card shows a live "running" pill and its ✕ kills chat A's process. Fix: gate `killableHandles` (line 4754) on the registry entry's `sessionId` matching the chat (e.g. `processRegistry.list(sessionId:)`) rather than the global `isAlive`.

Good catch, and correct. Fixed in the next commit.

`handleOwnership` dedupes within one transcript, which is the half of the problem I could see from the card. The counter is app-wide, so the other half needed the registry: `ProcessRegistry.isAlive(handle:sessionId:)` answers alive AND started by this chat, `ToolCallRow` now carries the session, and `killableHandles` asks the scoped question. `ManagedProcess` already had `sessionId` on every creation path, so this is a read, not new plumbing. A nil session (a task run's transcript) keeps the global answer rather than losing its badge. Covered by three tests in `ProcessCardControlsTests` using `registerSandboxed`, which is the one entry point that registers without launching anything.

While in there: `AgentEngine.processToolOutput` gates `killProcess` on `registry.list(sessionId: nil)`, so the MODEL can kill another chat's process by naming its handle, which `bg1` makes easy. The doc comment above the function says "against the session-scoped registry". That predates this PR and is a behaviour change rather than a layout one, so I have left it alone. Happy to send it separately if you want it scoped.

ddalcu:
🤖 **MLXServe-PRBot approved** — reviewed at head `4deb9ca`, no issues found.

ddalcu:
Thanks @lojza3d  Just had to hop on this branch to update keybindings a bit and remove some extra verbosity.

PR #341, opened 2026-09-03
qwen4_exp: gather QSA decode KV, flatten long-context tok/s

beamivalice:
## Summary

Decode on Qwen 3.8 Flash Next collapsed **47.5 → 10.7 tok/s from 16k → 384k**. Past the QSA budget, S==1 built a dense bool mask and ran full-KV SDPA. The QSA branch also skipped `qkvAttnDecodeKernel`, so `--kv-quant 8` still dequantized every cache row every step.

`qsaDecodeGatherAttn` reuses the sorted top-k, gathers those ~2k rows from the quant triples (or dense), and runs SDPA over K'≪kv.

- Prefill, verify widths, MTP verify, below-budget dense: untouched
- Batched decode still expands per-slot selections to masks (stacked path has no per-slot gather)
- Kill switch: `MLX_SERVE_QSA_DECODE_GATHER=0`

## Locked serial ladder

Same binary, same flags, same prompts; only the env gate differs.

`--kv-quant 8 --kv-attn-mode auto --no-mtp --no-pld --no-drafter`, temp 0, 32 tokens.

| ctx | ON | OFF | speedup |
|---|---|---|---|
| 32k | 51.3 | 40.8 | 1.26× |
| 64k | 48.7 | 32.9 | 1.48× |
| 128k | 46.1 | 23.5 | 1.96× |
| 256k | 40.9 | 14.6 | 2.80× |

ON is ~flat (51→41); OFF collapses (41→15). Residual ON slope is the indexer (still O(kv/ratio)), not attention.

## Parity

- Unit: subset-SDPA vs masked-full-SDPA max diff **0.000000** (quant and dense); take-then-dequant **bit-exact**
- Live 32k logprobs: top-1 32/32, identical completions, max\|dlogprob\| 0.24 (inside the pre-existing prefill gather-vs-mask envelope)

Two footguns pinned in code/tests: `mlx_take_axis` needs unsigned indices; quantized kernels misread strided take views (`mlx_contiguous`, not `+0`).

## Tests

- `zig build test -Doptimize=ReleaseFast` — **1868 pass**, 151 skip, 0 fail


beamivalice:
Some wired gemma size issue in HF API failed, I didn't touch that.

ddalcu:
Merged, thanks. Landed on main as 9657847 (rebased over the deferred n-gram commit, code unchanged) so GitHub did not close this on its own.

Verified here on an M4 Max, Flash Next 4-bit, serial decode, default flags: 16k and 32k flat, 128k 39.9 -> 47.4 tok/s. Added the [qsa-decode-gather] engaged grep to tests/test_qwen4_exp.sh so the kernel cannot silently stop engaging. Ships in 26.9.1.

PR #343, opened 2026-09-03
feat: Qwen3-VL-Embedding support with chat-template-based embedding pipeline

yachen4ever:
## Summary

Adds `/v1/embeddings` support for the Qwen3-VL-Embedding family (`model_type: qwen3_vl` / `qwen3_vl_text`), with embeddings matching the Python reference (mlx-embeddings) at cosine >= 0.999.

### What's included

1. **Model discovery whitelist** (model_discovery.zig): `qwen3_vl` / `qwen3_vl_text` added to `supported_model_types`.

2. **Config parsing** (model.zig): new `qwen3_vl`/`qwen3_vl_text` branch — mapped to the dense qwen3 trunk (QK-norm, silu, no GDN/attn_output_gate, `language_model.model` weight prefix), plus `parseQwenVisionFields` for ViT/M-RoPE config, with a `rope_parameters orelse rope_scaling` key fallback (Qwen3-VL-Embedding carries M-RoPE in `rope_scaling`).

3. **Chat-template embedding path** (server.zig `handleEmbeddings`): when `useChatTemplateEmbeddings()` is true and a chat template exists, embeddings render via `formatChat` (`[{role:"user", content:text}]` — the Jinja template auto-inserts `default_system_message`, matching mlx-embeddings' `format_embedding_input`) then manually append `bos_token_id`, emulating the tokenizer's TemplateProcessing post-processor that Python's `apply_chat_template(tokenize=True)` applies automatically.

4. **Family gating** (model.zig): the chat path is gated on a new `original_model_type` field — the verbatim config.json `model_type`, since family branches remap `model_type` to the trunk (qwen3_vl -> qwen3). Only the qwen3_vl embedding family enters the chat path, mirroring mlx-embeddings' own model_type-based dispatch. Checkpoints that merely ship a template stay on the raw path:
   - text-only Qwen3-Embedding (qwen3) embeds RAW in the reference (no Processor) despite carrying a ChatML template in tokenizer_config.json
   - WeMM-Embedding (qwen3_5) ships a GENERATIVE chat_template.jinja whose assistant prompt + think block must never render into an embedding request

### Verification

- **Qwen3-VL-Embedding-2B-4bit** (mlx-community): Zig vs Python `embed()` cosine **0.999421 / 0.999233 / 0.999235** (3 test texts, threshold 0.98)
- **Token-sequence parity**: Zig renders the same 22-token sequence as Python's full chat pipeline (system prompt + user text + generation prompt + post-processor EOS)
- **WeMM-Embedding-2B** (qwen3_5 trunk): raw path preserved (2 tokens for a 2-token text; an earlier iteration that keyed on template availability rendered 13)
- **Qwen3-Embedding-0.6B**: raw path unchanged (reference also embeds raw)
- **Regression**: `zig build test` 1865 passed / 0 failed / 149 skipped

### Notes for a follow-up PR

Multimodal (image) embeddings will ride on this baseline: content-parts request parsing, vision tower integration into the embedding forward, M-RoPE alignment, and pooling semantics for image tokens.

### Known limitation

The post-processor emulation appends `config.bos_token_id`. For Qwen3-VL-Embedding this is exactly the token the reference post-processor appends (151643 `<|endoftext|>`). Models whose post-processor appends a different special token (e.g. WeMM's `<embedding>`) would need a generic TemplateProcessing reader in tokenizer.json — out of scope here.


ddalcu:

- `src/server.zig:4577` — Use-after-free + double free in the raw-tokenization fallback: `defer allocator.free(raw_ids)` now covers the whole `blk2` block, but the final `else raw_ids` arm returns `raw_ids` itself, so the slice is freed on block exit, stored into `seqs`, read by the embedding batch, and freed again by the `seqs` cleanup `defer` at `server.zig:4515`. The pre-PR code only had the `defer` inside the two `wrapEncoderIds` arms for exactly this reason.
  Failure: POST /v1/embeddings against any model that is neither `use_bidirectional_attention` nor last-token pooled (e.g. a decoder with `pooling_mode: mean`, or any generative checkpoint — `effectivePooling()` defaults to `.mean`) embeds freed memory and double-frees on return (crash under GPA / silent garbage in release).

- `src/model.zig:1662` — `config.original_model_type = try allocator.dupe(u8, model_type)` adds the first owned heap string to `ModelConfig` that nothing frees: `ModelConfig` has no `deinit`, and none of the ~67 unit tests that call `parseConfigFromJson(testing.allocator, …)` (e.g. `model.zig:4463`, `4529`, `4619`, plus `deepseek_v4.zig`, `scheduler.zig`, `mage_flow.zig` callers) release it.
  Failure: `zig build test` — `std.testing.allocator` reports a leak at the end of every one of those tests, turning the previously green model-config suite red; in production every `/v1/load-model` swap leaks one string.

- `src/server.zig:4535` — No test for the new behaviour (CLAUDE.md: TDD mandatory, cross-arch = cover every touched arch). The PR adds a new `qwen3_vl` config branch, `useChatTemplateEmbeddings()`, the `rope_scaling` fallback in `parseQwenVisionFields` (which now also applies to the shared qwen3_5/qwen4_exp callers), and a new embedding token path, with zero unit tests or `tests/*.sh` coverage.
  Failure: none of the three regressions above is caught by the suite; a `parseConfigFromJson` test for a `qwen3_vl` config and a token-sequence test for the chat-template embedding path would have exposed both memory bugs.

ddalcu:

Re-reviewed at `cebf606`: the raw-id double free and the `original_model_type` leak from the previous pass are fixed. One finding remains.

- `src/model.zig:876` — the PR still adds zero tests (`git diff main...HEAD | grep '^+test "'` → 0): the new `qwen3_vl` config branch, `useChatTemplateEmbeddings()`, the `original_model_type_buf` capture, the `rope_scaling` fallback in `parseQwenVisionFields`, and the chat-template embedding path in `handleEmbeddings` (`server.zig:4539`) have no unit test or `tests/*.sh` coverage, which the repo CLAUDE.md makes mandatory (TDD; cross-arch changes cover every touched arch).
  Failure: the two memory bugs fixed in `cebf606` would both have been caught by a `parseConfigFromJson` test for a `qwen3_vl` config (testing.allocator flags the leak) plus a token-sequence test of the embedding path; without them the next regression in this branch ships unnoticed.

yachen4ever:
Thanks for the careful review — all findings verified and fixed in `cebf606` (pushed).

1. **UAF/double-free (`server.zig:4577`)** — confirmed. Reproduced against a mean-pooling generative checkpoint: first `/v1/embeddings` request returns 200, the second panics with "second free" and takes the server down. Fix restores the `defer` to the two `wrapEncoderIds` arms; the `else raw_ids` arm now documents that ownership moves to `seqs`.

2. **`original_model_type` dupe leak (`model.zig:1662`)** — confirmed. Measured against a pristine `upstream/main` worktree baseline: upstream suite reports 0 leaks, this branch reported 53 (46 pointing at the dupe). Replaced with a by-value `[64]u8` buffer + length; `ModelConfig` stays free of allocator-owned fields, and oversized model types simply fail the family match.

Suite after fix: 1865 passed, 0 leak reports; same-weights text parity vs `mlx-lm` unchanged (worst cosine 0.999743).

3. **Test coverage** — fair. Happy to add a `qwen3_vl` `parseConfigFromJson` case and a token-sequence test for the chat-template embedding path if useful.

yachen4ever:
Both follow-up test items from the review discussion are now on `main` (ddd7b6e):

1. **qwen3_vl `parseConfigFromJson` test** (model.zig) — verifies the family remap (`qwen3_vl` → `qwen3`, `qwen3_vl_text` kept in-family, `qwen3_5` excluded), the `useChatTemplateEmbeddings` gating flag, original-type preservation, vision-tower field capture (with `qv_num_pos_emb=2304` standing in for the multimodal RoPE width), and interleaved M-RoPE section parsing. The deepstack assertions from the `feat/mm-embeddings` counterpart are intentionally absent — that field only exists on the feature branch.

2. **Chat-template embedding token-sequence test** (server.zig) — the rendering block in `handleEmbeddings` is extracted into a `renderEmbeddingInputIds` helper, tested hermetically against an empty byte-level BPE tokenizer: the exact reference sequence including the EOS append, plus the four fallback paths (family gate off / no chat config / empty template / no-bos ownership transfer). The extraction is behavior-preserving.

Full suite: 1867 passed / 149 skipped / 0 failed, leak-checked.

ddalcu:
Re-reviewed at `ddd7b6e6`: the remaining test-coverage finding is fixed — `parseConfigFromJson` test for a `qwen3_vl` config (family remap, `original_model_type_buf` capture for `qwen3_vl`/`qwen3_vl_text` vs `qwen3_5`, vision geometry, `rope_scaling` M-RoPE fallback; by-value buffer so testing.allocator would flag any dupe), and the chat-template embedding path extracted to `renderEmbeddingInputIds` with a hermetic-tokenizer test pinning the exact rendered-ids + trailing `bos_token_id` sequence plus the gate-off / no-config / empty-template / no-bos arms. Verified `Tokenizer.deinit` frees the dupe'd vocab/special-token keys the test inserts (no leak, no double free), and the extraction is a verbatim hoist of the previous inline arm.

🤖 **MLXServe-PRBot approved** — reviewed at head `ddd7b6e6`, no issues found.

ddalcu:
🤖 **MLXServe-PRBot approved** — reviewed at head `b8a2c0e`, no issues found.

ddalcu:
🤖 **MLXServe-PRBot approved** — reviewed at head `c1a84e77`, no issues found.

yachen4ever:
Superseded by #351, which contains this PR's chat-template embedding pipeline and continues with the DeepStack vision path, the review fixes, and tests — all rebased onto current main. Closing this to keep the review in one place.

PR #344, opened 2026-09-03
fix(qwen4): stop cloning QSA history into every SSM checkpoint

beamivalice:
## Summary

Prefix-cache snapshots cloned Qwen 3.8 Flash Next's **growing QSA indexer history** (`aux_state` / `qsa_pooled`) at every SSM stride. 32 copies of a ~273k-row buffer is ~30 GB on top of ~70 GB weights.

That is not a RAM-ceiling OOM. A 400k prefill died at **77% RAM** with:

```
MLX error: [reshape] Cannot reshape array of size 524288 into shape (1,31744,4,128)
```

after a 2 GB hot-cache trim dropped the only snap that still held QSA history. Auto-context still advertised 700k–1M.

## Fix

- Stride captures keep GDN/PLE only (O(1) per layer).
- `attachQsaHistoryToLatest` stores **one** copy on the latest snap; restore slices it to `cp.pos`.
- Byte-budget trim that would drop that snap first slices history onto the last kept checkpoint.
- `qsaMaskFromQk` returns `error.QsaHistoryGap` if `keys` is shorter than `kv` (no process abort).
- Sizer + prefill guard bill live aux **once** (`qsaHistoryBytesPerToken`).

Sits on #341 (decode gather). Unique commit is the checkpoint leak.

## Verify (M5 Max 128 GB, `--kv-quant 8`, mixed 4/8 pack)

- `zig build test -Doptimize=ReleaseFast` — **1870 pass**, 151 skip
- **600k needle-in-haystack:** recovered `ARCHIVE-ID-ZEPHYR-9F3A` (824 prefill tok/s)
- **1M prefill:** 509 tok/s, decode 18.7, 92.4 GB active, no crash
- After 4 min cool, short ladder n=80:

| ctx | prefill | decode |
|---|---|---|
| 2k | 1064 | 62.2 |
| 32k | 1513 | 56.0 |
| 64k | 1348 | 52.7 |
| 128k | 1182 | 48.8 |


ddalcu:
Merged, thanks. Landed on main as 290b84c (your commit, rebased) plus a follow-up ce5b072 with three things the review turned up:

- the cancel handoff attached the live history to an earlier stride snap unsliced, so a restore installed rows the cache did not hold; attach now slices to the snap's pos and the gap guard is != instead of <
- the prefix-extend merge kept the old latest snap's full copy next to the new one, one copy per turn; keepOnlyLatestQsaHistory runs on the merged list
- a snap with no history restored aux-less and every turn on that prefix then failed with QsaHistoryGap; on this arch that is now a cache miss on both tiers

Verified here with a 350k prefill, cache on, then an extend turn that restored 350,738 tokens through the sliced history and answered in 2 s. Ships in 26.9.1.

Issue #330, opened 2026-08-31
Hot prefix cache commits nothing in a long agent session — #326's oversized-entry decline is a cliff, not a cap


d-b:
**Build:** `26.9.1-dev`, tree at `3afb77d` (#326) — the `skipped oversized entry` line exists only after that commit.
**Config:** Qwen3.8 Flash Next (`qwen4_exp`), `ctx=262144`, `--prefix-cache-mem 4GB`, SSD tier off, KV dense bf16 (no `--kv-quant`), PLD on.
**Follows** #324 (the unbounded growth this behaviour replaced — still open) and #306. **Caused by** #326 (merged as `3afb77d`), whose byte-budget decline sits on top of `4c06847`'s load-time clamp. Filed separately rather than as a #324 comment because #326 fixed the reported leak correctly — this is the policy it replaced it with.

## Summary

#326 made `--prefix-cache-mem` a real hard cap by **declining any commit whose snapshot exceeds the byte budget** (`prefix_cache.zig:764`). The accounting half of that PR is correct and #324's leak is genuinely closed — but the policy half has only one move available: when a conversation's KV no longer fits, the cache stops committing **entirely**.

In a long agent session that is the normal steady state, not an edge case. The entry crosses the budget once, at some mid-session prompt length, and from that turn onward every request cold-prefills the whole conversation. Nothing is ever evicted, nothing is ever reused, and the byte cap is "enforced" by holding zero bytes.

## Excerpt — `mlx-serve-2026-08-31T19-29-54Z.log`

Session prompt text is redacted; every other line is verbatim.

```log
  <- 102992+530 tokens streamed [prefill: 690.6 tok/s, decode: 35.9 tok/s] [stop]
  pld=enabled (draft_len=5, key_len=3)
POST /v1/chat/completions (45 msgs, max_tokens=32000, temp=1.00, top_p=0.95, top_k=20, stream=true, thinking=true, sys=123104b, user=225b, tools=46763b, tool_msgs=26)
  > "…"  [session text redacted]
  prompt=103761 tokens, max_gen=32000, ctx=262144
  spec-gate: ngram-score=0.179 (threshold=0.010)
  pld=enabled (streaming, draft_len=5, key_len=3)
  [hot-cache] skipped oversized entry (109199 tokens, 6121.58 MB > 4096.00 MB budget)
  <- 103761+5438 tokens streamed [prefill: 691.1 tok/s, decode: 35.7 tok/s] [tool_calls]
[...]

PR #349, opened 2026-09-03
metrics: expose live prefix-cache reuse during prefill

Sinojen:
Two gauges behind `--metrics` (no extra work when it is off):

| gauge | meaning |
|---|---|
| `prefill_prompt_tokens` | untrimmed prompt length of the running prefill |
| `prefill_target_tokens` | tokens actually forwarded, after the hot prefix cache trims the reused head |

They answer two questions that have no answer today.

### Progress

`prefill_tokens_live` reports "12288 tokens done" with no denominator. `live / target` is a true percentage — and because the target is published **after** the trim, a 34K prompt that restores 33K shows the 1K it really computes instead of a bar jumping to 97%.

### Live prefix-cache reuse

`prompt - target` is what the cache restored **for this request, while it happens**. The `prefix_cache_*_total` counters only give lifetime sums, so they cannot answer *"what did the request I am watching reuse?"* — which is exactly the question when validating a prefix-cache change. Given how much of 26.9.1 is prefix-cache work (#307, #310, #312, #330, #326), that seemed worth making observable.

## Measured

M4 Max, `Qwen3.5-4B-MLX-4bit`, 65K ctx, `--prefix-cache-mem 2GB --prefix-cache-entries 16`. Same prompt twice, the second with ~3K appended:

```
cold   prompt=24319  target=24319  reused=0      ( 0%)   22.4 s
warm   prompt=27399  target= 3111  reused=24288  (88%)    3.7 s
```

`27399 - 24288 = 3111`. The reused count matches the cold prompt to within the chat-template boundary, and the 6x speedup matches computing 11% of the tokens.

## Two placement details, both deliberate

- `prefill_prompt_tokens` is published at the top of the MLX prefill, **below** the ds4/llama/diffusion returns. Those paths never move `inflight_prefill_tokens` either; publishing at `runPrefill` entry would leave `prompt > 0` next to `target = 0` and `live = 0` for the whole of such a prefill, which reads as "nothing left to compute" — or divides by zero. All three gauges stay 0 together there, exactly as before this change.
- It is still published **before** the hot-cache restore, which can take seconds on a long prompt. A panel that shows `0/0` for that whole window looks stalled.

## Known limit, stated in the help strings

Like the existing `prefill_tokens_live`, both gauges describe a *single* in-flight prefill: with several slots prefilling at once the last writer wins, so the pair is well-defined while `requests_prefilling <= 1`. Fixing that properly means per-slot counters and would change the existing gauge’s semantics, so I left it out of this PR rather than widen the scope.

## Checks

- `zig build -Doptimize=ReleaseFast` — clean
- `zig build test` — 1873 passed, 151 skipped, 0 failed
- 3 files, +69 lines, no deletions, no behaviour change when `--metrics` is off


ddalcu:

- `src/metrics.zig:103` — Two new gauges (`prefill_target_tokens`, `prefill_prompt_tokens`) ship without any test, while `CLAUDE.md` ("TDD is mandatory": feature = unit test that fails without it + integration script if HTTP-observable) and the neighbouring gauges (`test "prefill progress is exposed live…"`, `tests/test_metrics.sh:236-284`) set the pattern.
  Failure: a later refactor that drops the `inflight_prefill_target_tokens.store(prefill_tokens.len)` line (or moves it back above the hot-cache trim, reintroducing the 97%-bar case the comment warns about) passes `zig build test` and `tests/test_metrics.sh` unchanged.

Otherwise the wiring checks out: the prompt gauge is published after the ds4/llama/diffusion early-returns so the three gauges stay 0 together on those paths, the target is stored from the post-trim `prefill_tokens` slice, and the shared `defer` clears all three on every exit.

Sinojen:
Fixed in b06dc68 — you were right, the rule is explicit and I skipped it.

**Unit test** — `prefill target/prompt pair exposes live prefix-cache reuse` (`src/metrics.zig`), next to `prefill progress is exposed live`. Pins both names and values through `renderPrometheus` *and* `renderJson`, the `target <= prompt` contract, `prompt - target` as the reuse, and that all three prefill gauges read 0 together at rest.

Red on revert, for the right reason: I removed only the two `writeGauge` calls and the two JSON fields, keeping the struct fields so it could not fail on a compile error:

```
error: 'metrics.test.prefill target/prompt pair exposes live prefix-cache reuse' failed:
           if (!ok) return error.TestUnexpectedResult;
```

**Phase 6 of `tests/test_metrics.sh`** — this is the part that actually answers your failure scenario. The unit test cannot catch a moved `inflight_prefill_target_tokens.store(prefill_tokens.len)`: above the trim it still renders a well-formed gauge and still passes `zig build test`. So Phase 6 samples the pair mid-flight and compares a cold prompt against the same head with a fresh tail, and fails when the warm target is not far below the warm prompt — the 97%-bar case, caught live.

Writing it turned up three ways the phase could have passed while proving nothing, all now handled explicitly:

* **Keep only post-trim samples (`target > 0`).** The sampler ticks every 2 s and the prompt is published before the trim, so an early tick honestly reads `prompt=N target=0`. Latching the largest-prompt sample keeps exactly that one, reports `target=0` for every request, and passes the warm assertions for the wrong reason. `target > 0` is a sound post-trim signal because a full match still re-forwards the last token (`prefix_cache.zig`, `final_len = effective_matched - 1`), so the trimmed length is never 0.
* **Unique tail per warm request, shared head.** Re-sending an identical prompt trims to one token and finishes between two sampler ticks. A fresh tail keeps the warm prefill thousands of tokens long so it spans a tick; up to three attempts, and the head stays hot so every attempt is equivalent.
* **Cold asserts `target >= 95% of prompt`, not `==`.** Earlier phases leave entries sharing this prompt's chat-template header, and the RAM tier has no minimum match length, so a few header tokens can legitimately be trimmed off a genuinely cold prompt. My first draft asserted equality and would have been flaky for a reason that has nothing to do with what the phase is testing.

Also documented the pair in `docs/reference.md` with its `Guards:` line, like every other gauge in that section, and extended the `tests/test_metrics.sh` row in `tests/CLAUDE.md`.

**Verification, and one gap I want to be straight about.** `zig build test` — 9/9 steps succeeded, 0 failed; plus the red-on-revert run above. `tests/test_metrics.sh` I have **not** executed against this commit: the checkout I wrote it in has the Zig toolchain but no model weights, so Phase 6 is reviewed and shell-checked (`bash -n`, and the sampling latch exercised against simulated tick sequences including the missed-window case) but not yet run end to end. I'll run it on the box with the model and report the output here — happy to hold the merge for that.

ddalcu:
🤖 **MLXServe-PRBot approved** — reviewed at head `b06dc68`, no issues found.

Re-reviewed at `b06dc68`: the missing-test finding is addressed — `metrics.zig` gains a render test for the `prefill_prompt_tokens`/`prefill_target_tokens` pair (Prometheus + JSON emission, `target <= prompt`, reuse = `prompt - target`), and `tests/test_metrics.sh` Phase 6 pins the post-trim placement live (cold prompt → warm same-head/fresh-tail, keeping only `target > 0` samples). Docs updated in `docs/reference.md` and `tests/CLAUDE.md`.

Sinojen:
Ran it properly, and closed the one gap the first reply didn't cover. Two things.

## 1. `tests/test_metrics.sh` — run, and Phase 6 proven non-vacuous

Full script against Qwen3.5-4B-4bit on an M4 Max (the model the PR's numbers came from): **57/57 passed**.

```
── Phase 6: prefill target is post-trim (live prefix-cache reuse) ──
  PASS cold prefill: both gauges published mid-prefill (prompt=16512 target=16512)
  PASS cold prefill: nothing substantial reused, target >= 95% of prompt
  PASS warm prefill: post-trim sample caught (prompt=22518 target=6037)
  PASS warm prefill: prompt grew (=22518 > cold 16512)
  PASS warm prefill: target is POST-trim, well below prompt (6037 < half of 22518)
  PASS warm prefill: reuse is visible and matches the cold prompt (reused=16481 ~ 16512)
  PASS both gauges back to 0 at rest (0 0)
```

`22518 - 6037 = 16481` against a cold prompt of `16512` — 31 tokens apart, which is the chat-template boundary the commit message predicted.

A passing integration test proves nothing on its own, so I built a **sabotaged binary** with exactly your failure scenario — `inflight_prefill_target_tokens.store(...)` moved to entry, above the hot-cache trim — into a separate `--prefix` so `zig-out/bin/mlx-serve` was never touched, and ran the same Phase 6 against it:

| | correct build | target stored at entry |
|---|---|---|
| warm prompt | 22518 | 22518 |
| warm target | **6037** | **22518** |
| visible reuse | 16481 | **0** |
| Phase 6 | 7/7 pass | 5 pass / **2 fail** |

Identical prompts, identical token counts, one moved line. The two assertions that fail are exactly `target is POST-trim, well below prompt` and `reuse is visible`. Your scenario is caught.

## 2. The gap you didn't flag but would have next: nothing in-tree read it

`src/html/metrics.js` and the macOS tray both poll `/metrics.json`, and neither read the denominator this branch added. The commit argued its own placement entirely in panel terms — "a bar jumping to 97%", "a panel showing 0/0 looks stalled" — while there was no bar. That's a fair thing to be called out for, so 194e2e1 makes the argument true:

```
before   prefilling · 6.0K tok
after    prefilling · 2.0K / 6.0K (34%) · 16.5K reused
```

`target == 0` renders the bare phase word, never `0%` — the size is not *known* yet (the cache restore runs before the trim, and a ds4/llama/diffusion prefill never moves any of the three), which is a different state from "0% done". A feed with no pair at all falls back to the old `· N tok` instead of dividing by zero. The percentage is clamped to 100, since the chunk loop stops a token — plus up to 30 for the SSM snapshot backoff — short of the target.

The logic is a pure `prefillPhaseLabel(g)`, so it is tested hermetically in `tests/metrics_panel_test.mjs` (Phase 0, no model, no GPU): seven cases over the real numbers above. `a pre-trim target would hide the reuse and inflate the bar` is the panel-side statement of what Phase 6 pins server-side — same live count, same prompt, only the target moves, and the reuse clause vanishes while the bar reads 9% instead of 34%. Red on revert: pointing the denominator at `prefill_prompt_tokens` fails 3 of the 7.

## Everything green

- `zig build test` — 0 failed (re-run after the panel change; `metrics.js` is `@embedFile`'d, so `server.zig` rebuilt)
- `tests/test_metrics.sh` — 57/57
- `node tests/metrics_panel_test.mjs` — ALL PASS

One thing I deliberately did **not** do: add a `## Rules` bullet to `CLAUDE.md`. That file is at 99,998 bytes against its own stated 100k ceiling, so there is no room without deleting something of yours. The rule and the full story live in the `docs/reference.md` observability section instead, with the `Guards:` line — say the word if you would rather have a line in `CLAUDE.md` and I will make room wherever you point me.

Sinojen:
Addendum — the panel claim above rested on hermetic tests of the pure `prefillPhaseLabel`. Confirmed live as well.

Rebuilt with `zig build -Doptimize=ReleaseFast` (`metrics.js` is `@embedFile`'d, so this is the JS that actually ships), served the index page from that binary, and evaluated the embedded function in a real browser against live `/metrics.json` during a warm prefill — same shared-head / fresh-tail request shape Phase 6 uses:

```
prefilling · 4.1K / 6.0K (68%) · 16.5K reused
```

Consistent with Phase 6's `target=6037`, `reused=16481`: roughly two thirds of the post-trim work done, with the restored head shown next to it rather than hidden inside a prompt-sized denominator.

One methodological note, so the claim is not read as more than it is: I drove `prefillPhaseLabel` against a live feed from inside the loaded page rather than reading the rendered tile. The panel's own 1 s poll loop is timer-throttled in a background tab and does not repaint during the few seconds a warm prefill lasts. So this verifies the shipped logic against live gauges; the DOM write itself is covered by the existing Phase 2 index-page assertions.

Full suite re-run against that same freshly built binary, since the earlier 57/57 predated the panel commit:

- `tests/test_metrics.sh` — 57/57
- `zig build test` — 0 failed
- `node tests/metrics_panel_test.mjs` — ALL PASS

No further changes; b06dc68 and 194e2e1 are the whole of it.

ddalcu:
🤖 **MLXServe-PRBot approved** — reviewed at head `194e2e1`, no issues found.

Re-reviewed at `194e2e1`: the delta adds the panel-side `prefillPhaseLabel` (pure, exported via `__mlxPanel`, `target == 0` renders the bare phase word, pct clamped to 0..100) with seven node tests; `tests/metrics_panel_test.mjs` passes locally (ALL PASS).

Sinojen:
Addendum: the panel change is now confirmed live, not just by the hermetic test.

The previous comment backed `prefillPhaseLabel` with `tests/metrics_panel_test.mjs` only. Since that is a pure function tested in node, it says nothing about the code that actually ships inside the binary, so I rebuilt and checked it end to end.

Fresh `zig build -Doptimize=ReleaseFast`, server started on its own port with Qwen3.5-4B-4bit, prefix cache primed with a head, then a warm request with a fresh tail. Sampling `/metrics.json` from the loaded page and rendering it through the panel function embedded in that binary:

```
prefilling · 4.1K / 6.0K (68%) · 16.5K reused
```

4096 of 6037 tokens forwarded, 16.5K restored by the hot cache — the same `target=6037`, `reused≈16481` that Phase 6 asserts on the server side. The arithmetic holds in the shipped panel, not only in the test.

One methodological note for accuracy: I drove `prefillPhaseLabel` against a live feed from inside the page rather than reading the rendered tile, because the panel's own 1 s poll loop is timer-throttled in a background tab and does not repaint during the few seconds a warm prefill lasts. So this verifies the shipped logic against live gauges; the DOM write itself is covered by the existing Phase 2 index-page assertions.

Re-ran everything against the newly built binary, since the earlier `57/57` was measured on the pre-panel build:

- `tests/test_metrics.sh` — 57/57, Phase 6 reproducing `cold 16512/16512`, `warm 22518/6037`, `reused 16481`
- `zig build test` — 0 failed
- `node tests/metrics_panel_test.mjs` — ALL PASS

No further changes; b06dc68 and 194e2e1 are the whole of it.

PR #350, opened 2026-09-03
perf(qwen4): cheaper MTP rounds — deferred PLE gather, EV seed, coarse-head draft rerank

beamivalice:
## Summary

Speculative decoding (in-checkpoint MTP head) on Qwen3.8-Flash-Next (`qwen4_exp`) gets cheaper per round and reaches full speed on the first request. Measured on an Apple M5 Max 128 GB with the `mixed-4-8bit` pack, greedy, 400-token generations, counterbalanced boots with a serial drift meter (≤ 0.7% drift in every run):

| | before | after |
|---|---|---|
| code, auto depth | 106.6 tok/s | 127.1 (×1.19) |
| prose, auto depth | 70.2 | 76.8 (×1.09) |
| code, forced depth 5 | 110 | 128 |
| llmprobe decode (median of 2 boots) | 89 | 106 |

### What changed

- **Deferred PLE gather in the MTP verify build.** The trunk's layer-1 n-gram lookup is a host gather on the draft ids, which are still lazy when the verify graph is built; the eager gather parked the 47-layer build behind the whole draft chain. The verify build now uses the same deferred leaf as pipelined decode (`ple_defer`), with the PLE rollback-capture claim moved to build time (otherwise partial-accept rollback silently lost PLE state). Verify lap 6.0 → 2.0 ms, byte-identical output.
- **The MTP head projects only the row its caller consumes** (`Qwen4MtpProject`): no mixer + 248k-vocab lm_head over discarded rows on the merged history step; nothing at all on prompt-prefill appends.
- **Cross-request EV seed on the qwen4 head.** The depth controller inherits the previous request's acceptance surface instead of re-warming for ~15 rounds. Declined under `MLX_SERVE_MTP_FORCE_DEPTH` at both publish and consume.
- **Draft rerank for qwen4 drafts.** A draft step was bytes: the full 8-bit 675 MB lm_head read per draft. Drafts now shortlist on a 3-bit coarse copy (265 MB, built at first draft, never written to disk) and re-score the top 32 rows exactly — the sidecar's existing rerank machinery, shared as free functions, fed the MIXER output (never the pre-mixer stream). Round 41.8 → 38.4 ms at depth 5, per-index acceptance identical. `MLX_SERVE_MTP_DRAFT_RERANK=0` restores.
- **Fixes:** a double free of the head's stream on an lm_head error path (latent); padding rows past the tokenizer's last id (243 on this model) were samplable and decoded to nothing — now suppressed, logprobs unchanged.

### Measured and not shipped (stories in `docs/gotchas/engine-mlx.md`)

Per-step draft dispatch (neutral: the draft phase is lm_head bytes, not dispatch latency); 8-bit head experts pack (−1.7% code, acceptance within noise); a measured serial-vs-MTP arm-probe controller (−3.8% code); planning tokens from per-request acceptance instead of the round-cost table's `tok` column (novel −4..10%). The last two record a real open item: on predictable text the EV base never leaves width 1 and throughput rides the two-chunk extension horizon, which flaps between ~111 and ~154 tok/s on near-identical prompts.

## Test plan

- `zig build test -Doptimize=ReleaseFast`: 9/9 steps, 1883 passed, 152 skipped, 0 failed (rebased on current main).
- New guards: deferred-PLE verify-width parity (S=4/7) + scan pin; head last-row projection fixture test; `qwen4MtpForward` single-owner scan; EV-seed round-trip; rerank exactness/fallback/suppression (bars on exact logits, ties acquitted by margin); `fillSuppressMask` / `definedVocabSize`.
- Byte identity at forced depth across 4 boots (72 responses) for the round-pipelining commits; `test_mtp_equivalence.sh`-class acquittal for the rerank (drafts steer acceptance, the trunk verify decides bytes).
- Not run here: `benchmarks.md` column (rule: M4 Max only); `app` tests (no Swift changes in this diff).


beamivalice:
Context scaling  decode · first-token latency · speculation vs prompt size
      ~513   81.8 tok/s   579 ms      1.92 tok/step · 1.23× ceiling · 2.91 tok/step at ceiling
     ~4.3k   96.1 tok/s   2.6s        3.31 tok/step · 1.24× ceiling · 4.8 tok/step at ceiling
     ~8.2k   83.9 tok/s   5.6s        2.74 tok/step · 1.48× ceiling · 6.4 tok/step at ceiling
    ~16.3k   78.5 tok/s   11.0s       2.78 tok/step · 1.52× ceiling · 6 tok/step at ceiling
    ~32.9k   58.8 tok/s   23.5s       1.94 tok/step · 1.04× ceiling · 2 tok/step at ceiling
    ~65.5k   54.8 tok/s   48.6s       3 tok/step · 1.41× ceiling · 4.47 tok/step at ceiling

beamivalice:
Re the `qwen4MtpForward` ownership finding (`h` leaking on a failed draft step): checked, and the stream does stay owned on every error path — but rather than argue it, 48662a7 proves it with a test.

**Why the paths are covered.** `h` is created by the `mlx_reshape` at the top of the function and the very next line, at *function* scope, is `errdefer _ = mlx.mlx_array_free(h);` — Zig evaluates that at unwind time against the current value of `h`, so a failure in `hcRead`, the `.last_row` slice, or `lmHeadProject` frees whatever `h` holds at that moment. The slice path frees its predecessor before rebinding (`_ = mlx.mlx_array_free(h); h = h_last;`), and its own `errdefer` on `h_last` is block-scoped with no fallible statement between the rebind and the block end, so there is no window where two frees can fire. `hcWrite` frees its input only after its last `try` (12168), so a failed write leaves the caller's handle for that same errdefer. The errdefer the PR removed was the *second* one after the mixer read; with both live, an `lmHeadProject` failure freed `h` twice — that was the double free, and removing the later one is the fix.

**What 48662a7 adds** (`src/mlx.zig`, `src/transformer.zig`):
- a test-only fault injector at `mlx.check` (compiled under `builtin.is_test` only; zero cost in release) that makes the k-th checked op report failure;
- `qwen4MtpForward: every error path of the MTP head forward releases its stream (fault-injected, QWEN4_TEST_MODEL)` — sweeps every k across the function for all four `Qwen4MtpProject` variants at seq_len > 1 (so the slice path runs), asserting active memory returns to baseline after each injected failure; a double free would abort the binary, which is the other half of the proof;
- a model-free sweep of the same ownership shape with a negative control (the shape minus the errdefer trips the leak oracle), so the real test is not vacuous where the fixture is absent;
- the existing scan strengthened: the errdefer is the line after `h`'s creation, every `h =` rebind is either `hcWrite` or preceded by a free of the predecessor, and `hcWrite` frees its input past its last `try`.

Full suite on the branch: 9/9 steps, 1884 passed, 0 failed. One honest limit, documented in the tests: MLX laziness means a handle leaked before anything downstream is evaluated holds no buffer, so the memory oracle is necessary-not-sufficient for the first few ops; the scan covers those unconditionally.

ddalcu:

- `src/mtp.zig:911` — `rerankSelect` derives the coarse head's bit width from the environment (`rerankCoarseBits()` → `getenv` + parse) on every draft step, while `buildRerankCoarse(s, target, bits)` packed the weights at whatever width its caller passed. The width is a property of the built head and should travel with it (e.g. a `bits` field next to the `QLinear` / `rerank_rows`), not be re-derived per step.
  Failure: the new tests call `buildRerankCoarse(s, &fx.xfm, 3)` then `rerankSelect(...)`; running `zig build test` with `MLX_SERVE_MTP_DRAFT_HEAD_BITS=2` or `4` (the knob the doc comment invites for A/B) dispatches `mlx_quantized_matmul` at the wrong width over 3-bit-packed rows and the shortlist test fails/produces garbage. Same double read in `maybeBuildDraftRerank` (`mtp.zig:709/719`). Storing the width also removes a per-token `environ` scan on both arms.

- `src/transformer.zig:13290` — The qwen4 coarse head is built lazily inside the first request's `mtpChainBuild` (`generate.zig:4688` → `canRerankDrafts()` → `qwen4DraftRerankReady`): a full `requantizeRows` of the 248320×2560 lm_head plus a synchronous `mlx_eval` of ~240 MB runs on the inference thread in the middle of the first speculative round, and that eval also drains everything already queued on the stream. The sidecar arm pays this at load (`maybeBuildDraftRerank`).
  Failure: first request after `/v1/load-model` on Qwen3.8-Flash-Next takes the whole requantize+eval as first-token latency. The stated reason ("the load path runs before `lm_head_w` is assigned") only holds inside `loadQwen4Mtp`; `Transformer.init` has `lm_head_w/lm_head_s` as locals before the struct literal, and `main.zig` already has a post-init hook (`installSuppressMask`) where `qwen4DraftRerankReady()` could run once at startup.

- `src/transformer.zig:5433` — The doc comment says "transformer.zig cannot import mtp.zig", but line 8 of the same file is `const mtp_mod = @import("mtp.zig");` (this PR adds it). `ev_seed_accept: ?[mtp_mod.MAX_DEPTH]f32` works directly, which deletes `MTP_EV_SEED_DEPTH`, the comptime assert in `generate.zig:356`, and the misleading comment.
  Failure: the next reader adds more mirrored constants + asserts instead of using the import.

beamivalice:
All three addressed, one commit each on this branch:

- **Coarse-head bit width travels with the built head** — `ba16883`. `buildRerankCoarse` now returns a `RerankCoarse { q, bits, group_size, rows }` and `rerankSelect` reads the width from it; the env is read exactly once, at build, on both the sidecar and qwen4 arms (the parallel `rerank_rows` fields went away with it). Test: build at 3 bits with the env set to 2 → the shortlist is still exact vs the full argmax. Before the fix that combination dispatched `mlx_quantized_matmul` at 2 bits over 3-bit-packed rows and aborted the process.
- **Coarse head built at load, not on the first draft** — `5cd10fa`. Both load paths converge on `scheduler.doLoadOnInferenceThread` (boot load and `/v1/load`), so one hook there builds it once, gated exactly as the lazy build was; the lazy path survives only as a one-shot fallback for a path that did not run the eager one. The `Transformer.init` locals were not needed. Logged with bits, MB and ms. Tests: a scan that both load paths reach the builder, and `canRerankDrafts()` true before any draft with the same mlx handle afterwards (no rebuild).
- **`MTP_EV_SEED_DEPTH` gone** — `9d6d4a5`. `ev_seed_accept: ?[mtp_mod.MAX_DEPTH]f32`; the comptime assert and the stale comment are deleted; a scan pins that the mirror does not come back.

Full suite on the branch: 9/9 steps, 1888 passed, 0 failed.

ddalcu:
🤖 **MLXServe-PRBot approved** — reviewed at head `9d6d4a5`, no issues found.

PR #351, opened 2026-09-04
feat: multimodal (vision) embeddings for Qwen3-VL-Embedding

yachen4ever:
## Summary

Adds vision (multimodal) embedding support to `/v1/embeddings`, enabling Qwen3-VL-Embedding-2B to produce joint text+image embeddings. Also fixes two pre-existing bugs in the dense M-RoPE path that affected image chat quality.

This PR builds on #343 (PR 1: text-only Qwen3-VL-Embedding support). The commits are on a separate branch (`feat/mm-embeddings`) to keep `main` clean for PR 1.

### 1. Vision embedding pipeline (826 lines across 6 files)

- **`server.zig`** (+491): `handleEmbeddings` now accepts OpenAI-compatible content parts (`input: [{type: "image_url", image_url: {url: ...}}, {type: "text", text: ...}]`). Two-phase orchestration: (1) `encodeVisionDeepstack` produces visual features via the ViT tower with DeepStack merger outputs, (2) token sequence + visual features are combined for the embedding forward pass.
- **`scheduler.zig`** (+116): `encodeVisionDeepstack` orchestrates the vision tower forward pass and DeepStack layer capture (layers 5/11/17 per config).
- **`qwen_vision.zig`** (+156): `forwardWithDeepstack` extends `QwenVision.forward` to capture intermediate hidden states at configurable layer indices and run them through the DeepStack merger (post-shuffle PatchMerger → LayerNorm → add to decoder layers 0/1/2). The merger weights (`deepstack_merger_list.{0,1,2}`) are conditionally loaded during init.
- **`generate.zig`** (+37): `computeEmbeddingsVision` entry point for batch=1 vision embedding forward pass.
- **`model.zig`** (+17): `qv_deepstack` config field parsing for `deepstack_visual_indexes` and `deepstack_merger_list`.
- **`vision.zig`** (+10): helper functions for vision embedding pipeline.

### 2. Bug fix: DeepStack scatter row-index missing cumsum (`transformer.zig`)

**The bug**: The DeepStack scatter block had a comment saying `src_idx = cumsum(mask) - 1`, but the implementation used `mask - 1` instead — missing the `cumsum` call entirely. This meant every image token picked the **same row** (stream[0]) from the DeepStack output instead of the row at its own position. 

**Symptoms**: cosine similarity vs Python reference was 0.82 (should be ≥0.98). The error had a distinctive fingerprint: all image rows shared the same fixed error vector (cos 0.9745 with each other), while text rows were perfect. Row 15 happened to be correct by coincidence (its src_idx resolved to 0). Rows 211+ (text) were perfect because the mask zeroed them out.

**Fix**: `mlx_cumsum(ds_mask, axis=0, inclusive)` → subtract 1 → `remainder(n_rows)`. Text positions get negative indices that wrap back, but the mask multiplies them to zero, preserving the original design.

**Forensic methodology** (documented for future contributors):
1. Distribution comparison first: `mean/std/min/max` matched perfectly but `cos ≈ 0` → layout/misalignment issue, not numerical
2. Transpose-view dump trap: MLX lazy-transposed output materializes as token-major (L,H,hd) in memory, but the dump header wrote the logical shape (H,L,hd) → `v_t`/`sdpa_out` appeared to have `cos ≈ 0.06` when re-read with correct layout they were 0.995/0.998
3. Error fingerprint analysis: `cos(err, ds0[0] - ds0[r])` per-row → mean 0.9916, max 1.0000 → pinned row-selection bug

### 3. Bug fix: M-RoPE dtype hardcoding (`transformer.zig`)

`applyMrope` in the full-rotation branch (when `head_dim == rope_dims`) hardcoded `astype(.bfloat16)`, which silently downgraded fp16 checkpoint q/k tensors. Fixed to preserve the input array's dtype. `forwardMoeWith`'s `beginMropeChunk` was also updated to pass `h`'s dtype (matching `forwardStandardWith`'s existing behavior).

**Note**: `buildMropeCosSin` (pure-text QSA path) still defaults to bf16. No model has triggered an issue there, so it's left unchanged.

### 4. PR 1 commits (included)

This branch also contains the PR 1 commits (`48ed537`, `b011593`) which add Qwen3-VL-Embedding config parsing, chat-template-based embedding pipeline, and the `useChatTemplateEmbeddings()` guard. See PR #343 for details.

### Multimodal parity (vision embeddings)
- **Model**: `mlx-community/Qwen3-VL-Embedding-2B-4bit`
- **Reference**: `Blaizzy/mlx-embeddings` running `model.embed()` with image input (448×448 test image → 196 image tokens)
- **Cosine similarity**: **0.998207** (threshold: ≥0.98)

### Text-only regression
- **Model**: `mlx-community/Qwen3-Embedding-0.6B-4bit-DWQ`
- **Reference**: `mlx_lm` same-weights, last-token pooling, 8 test sentences
- **Worst cosine**: **0.999743** (threshold: ≥0.999)

### Test suite
- `zig build test --summary all`: **1873/2024 passed** (151 skipped, 0 failed)
- 53 pre-existing leak reports — confirmed identical to baseline via `diff` (not introduced by this PR)

## Design decisions

1. **Image embedding batch=1**: M-RoPE position tables and visual row splicing have complex batch semantics. Fixed batch=1 for vision, text retains existing batch support.
2. **DeepStack merger is config-driven**: Layer indices come from `deepstack_visual_indexes` in config (not hardcoded). Falls back gracefully when absent.
3. **Dense path M-RoPE fix**: The `forwardStandardWith` M-RoPE fix benefits both embedding and chat paths — existing image chat was also affected by the missing M-RoPE in the dense path.

## Notes for follow-up

- WeMM-Embedding-2B (`model_type: qwen3_5`, `deepstack_visual_indexes: []`): config loads and **text-only** embeddings work correctly (the PR 1 `useChatTemplateEmbeddings()` guard keeps it off the generation chat template). **Multimodal embedding requests return 400** — the mm path is gated on the `qwen3_vl` family (`useChatTemplateEmbeddings()`), mirroring the reference implementation's routing. Supporting WeMM vision embeddings would need separate work (it uses a different embedding contract — `<embedding>` sentinel tokens, no DeepStack).
- Batch vs single equivalence: a pre-existing `cos=0.9967` difference for repeated long sentences in batch mode is confirmed via `git stash` baseline (not introduced by this PR).
- Reranker support (yes/no logit → sigmoid) is out of scope; can be a follow-up PR.

## Model weights

- `mlx-community/Qwen3-VL-Embedding-2B-4bit` (third-party 4bit conversion, not official)
- Official Qwen/Qwen3-VL-Embedding-2B is PyTorch format; MLX conversion instructions can be added to docs if needed.


ddalcu:

This branch contains #343's commits, so it inherits both memory bugs reported there; listing them first because they are shipped by this PR too.

- `src/server.zig:4635` — (inherited from #343) Use-after-free + double free in the text fallback: `defer allocator.free(raw_ids)` now spans the whole `blk2` block but the final `else raw_ids` arm returns the same slice, which is then stored into `seqs` and freed again by the `seqs` cleanup `defer`.
  Failure: `/v1/embeddings` with text input on any model that is neither bidirectional nor last-token pooled (`effectivePooling()` defaults to `.mean`) reads freed memory and double-frees on return.

- `src/model.zig:1692` — (inherited from #343) `config.original_model_type = try allocator.dupe(...)` is an owned string that no code path frees; `ModelConfig` has no `deinit`.
  Failure: every `parseConfigFromJson(testing.allocator, …)` test (~67 of them) now reports a leak under `std.testing.allocator`, failing `zig build test`; production leaks one string per model load.

- `src/server.zig:5015` (and the `deepstack_alloc` dupe failure a few lines below) — On the `computeQwenMrope` error path the tower is `allocator.destroy(t)`ed "without deinit" because "arrays move out below" — but they have not moved yet at that point, so `tower.merged` and every `deepstack[i]` mlx array leak. The dupe-failure branch additionally leaks the heap `md.pos` slice.
  Failure: a large image whose M-RoPE computation fails (or an OOM assembling the request) leaks the whole merged vision embedding plus three DeepStack streams of GPU memory per request; under load this is a steady MLX-memory leak with no log signal beyond the 500.

- `src/scheduler.zig:4361` — In `runEmbedRequest`, the `resetCache` failure return happens before the `defer` that frees `vision_emb` / `mrope_pos_alloc` / `deepstack_alloc`, and `computeEmbeddings` returning early on enqueue failure (`scheduler.zig:1991`) never runs the request at all — in both cases the caller has already handed ownership over and does not free them either.
  Failure: same leak class as above on the inference-thread error paths.

- `src/server.zig:4748` — ~1,270 added lines (new `qwen3_vl` config branch, DeepStack config parsing, `ParsedEmbeddingInput`, `computeEmbeddingsVision`, `forwardEmbeddingVision`, DeepStack injection, `qsaHistoryBytesPerToken` in the memory sizer) with a single new assertion in an existing test; CLAUDE.md makes TDD mandatory and requires an integration script for HTTP-observable features. `ParsedEmbeddingInput.parse` and the `<|image_pad|>` expansion are pure and trivially unit-testable; a `qwen3_vl` `parseConfigFromJson` test would have caught the leak above.
  Failure: none of the four defects above is detected by `zig build test` or `tests/*.sh`.

Minor: `if (md.pos != null and mrope_pos_alloc == null)` can never be true (`@constCast` cannot fail) — dead error branch.

ddalcu:

Re-reviewed at `e2551f7`: the raw-id double free, the `original_model_type` leak, the `computeQwenMrope` / `deepstack_alloc` error-path leaks and the dead `mrope_pos_alloc == null` branch are fixed. Two findings remain, one of them introduced by the fix commit.

- `src/server.zig:5078` — the new cleanup in the `computeEmbeddings` catch frees `vision_emb` / `mrope_pos_alloc` / `deepstack_alloc` on EVERY error, but `error.EmbedFailed` is also what `computeEmbeddings` returns after the inference thread already ran the request and freed those same buffers itself: `runEmbedRequest` (`scheduler.zig:4361`) now frees them in the `resetCache` failure arm, and its `defer` at `scheduler.zig:4387` frees them when `computeEmbeddingsVision` fails at `:4395`. Only the enqueue failure (`embed_queue.append` at `scheduler.zig:1989`) leaves the buffers unowned.
  Failure: any inference-thread failure on a multimodal embedding (a `computeEmbeddingsVision` error, or `resetCache` failing) → `finishEmbedRequest` wakes the conn thread, which `mlx_array_free`s the merged vision embedding, the DeepStack streams and the position slice a second time (racing the inference thread's own `defer`) — double free / use-after-free instead of the previous leak. Distinguish the enqueue error (e.g. check `req.done == false`, or free only for the `append` error) so exactly one owner frees.

- `src/server.zig:4748` — still zero tests added (`git diff main...HEAD | grep '^+test "'` → 0) for ~1,300 new lines: the `qwen3_vl` config branch, DeepStack config parsing, `ParsedEmbeddingInput.parse`, `<|image_pad|>` expansion, `computeEmbeddingsVision` / `forwardEmbeddingVision`, and the ownership handoff above; CLAUDE.md makes TDD mandatory and requires an integration script for HTTP-observable features.
  Failure: the double free above and the four bugs fixed in `e2551f7` are all invisible to `zig build test` / `tests/*.sh`; a `parseConfigFromJson` test for a `qwen3_vl` config plus a failure-injected `runEmbedRequest` ownership test would catch them.

ddalcu:

Re-reviewed at `fa4a9ec`: the double free is fixed — `handleMultimodalEmbedding` now frees `vision_emb`/`mrope_pos_alloc`/`deepstack_alloc` only when `computeEmbeddings` fails before the inference thread saw the request (`err != error.EmbedFailed`), which matches the scheduler's ownership (`runEmbedRequest` frees on the `resetCache` arm and via its mm `defer`; `computeEmbeddings` only returns `EmbedFailed` after `done` is set). Remaining:

- `src/server.zig:4840` — no tests accompany the new multimodal `/v1/embeddings` pipeline (`handleMultimodalEmbedding`, `computeEmbeddingsVision`, `forwardWithDeepstack`, the DeepStack cumsum scatter fix, the M-RoPE dtype fix); CLAUDE.md's TDD rule requires a failing unit test per feature and a red→green regression test per bug fix, and `git diff main...` adds zero `test` blocks and nothing under `tests/`.
  Failure: the `mask - 1` → `cumsum(mask) - 1` scatter regression (cos 0.82) and the ownership hand-off that took three rounds to get right can both silently regress — nothing in `zig build test` exercises them.

yachen4ever:
Thanks — good catch on the cleanup commit. All findings verified; fixed in `fa4a9ec` (pushed).

1. **Double free in the `computeEmbeddings` catch (`server.zig:5078`)** — confirmed, and it was introduced by `e2551f7`: freeing on every error double-freed the inference-thread failure paths (the `resetCache` arm and the mm `defer` both run before/while `finishEmbedRequest` wakes us) and raced the inference thread's own cleanup. The fix narrows conn-side ownership to exactly one case: `err != error.EmbedFailed` (enqueue failure, request never seen by the scheduler). On `EmbedFailed` the buffers are now exclusively the inference thread's — single owner, no race.

2. **Test coverage** — fair point; all five bugs so far were caught by manual verification (crash repro + leak accounting against a pristine `upstream/main` worktree) rather than the suite. Happy to add the failure-injected `runEmbedRequest` ownership test you sketched, plus `ParsedEmbeddingInput.parse`, the `<|image_pad|>` expansion, and a `qwen3_vl` `parseConfigFromJson` case — say the word and I'll push a follow-up.

After `fa4a9ec`: 1873 passed / 0 leak reports, mm parity 0.998207 unchanged, text parity 0.999743 unchanged.

yachen4ever:
Tests are in (3de0e08) — one per review-flagged surface:
- qwen3_vl `parseConfigFromJson` family remap + DeepStack index capture (model.zig)
- `ParsedEmbeddingInput.parse` parts / chat wrapper / rejection (server.zig)
- `expandImagePadTokens` pad-run expansion (server.zig — extracted from `handleMultimodalEmbedding` as a pure function)
- `computeEmbeddings` ownership contract (scheduler.zig): a FailingAllocator-driven enqueue failure surfaces as a non-EmbedFailed error with the request untouched; a done+error_name request maps to `error.EmbedFailed`

Full suite: 1878 passed / 151 skipped / 0 failed, leak detector clean.

ddalcu:

Re-reviewed at `3de0e08`: the new hermetic tests cover the surfaces flagged earlier — `parseConfigFromJson` for `qwen3_vl` (family remap, `original_model_type_buf`, DeepStack indexes, M-RoPE sections), `expandImagePadTokens` (extracted from the handler; pad expansion, bos, `NoImagePad`), `ParsedEmbeddingInput.parse` (parts, chat wrapper, rejections), and the `computeEmbeddings` ownership contract (enqueue OOM ≠ `EmbedFailed`, `done + error_name` → `EmbedFailed`). Field names, defaults and signatures used by the new tests check out against the tree. Remaining:

- `src/transformer.zig:11276` — the two bug fixes this PR ships in the forward path (the DeepStack scatter `mask - 1` → `cumsum(mask) - 1` row index, and `applyMrope` preserving the input dtype instead of hardcoding `.bfloat16`) still have no regression test; `git diff main...` adds zero `test` blocks to `transformer.zig`, while CLAUDE.md requires "Bug fix = regression test red→fix→green, red-on-revert".
  Failure: reverting the `mlx_cumsum` line makes every image token read DeepStack row 0 again (the cos-0.82 symptom in the PR body) and `zig build test` stays green — a tiny synthetic mask/stream pair asserting that row `r` of the scatter equals stream row `r` (and that an fp16 `q` comes back fp16 from `applyMrope`) is enough to pin both.

yachen4ever:
Done in f29a583 — both forward-path fixes now have red-on-revert regression tests:
- `deepstackScatter` (extracted from the forward loop into a file-level fn, alongside `spliceVisionRows`): a synthetic mask `[0,1,1,0,1]` + constant-row stream asserts the r-th image token reads row r (10/11/12) and text positions stay zero. Reverting to `mask-1` makes it fail.
- `applyMropeCore` (the body of `applyMrope`, which only carried the stream): an identity rotation (cos=1/sin=0) must preserve fp16 values AND dtype — the dtype expect trips on the old hardcoded `.bfloat16` — plus a cos=0/sin=1 case pinning `rotate_half` exactly.

Both verified red-on-revert per CLAUDE.md. Full suite: 1880+20 passed / 151 skipped / 0 failed, leak detector clean (fresh cache).

ddalcu:
Re-reviewed at `f29a583a`: the remaining finding is fixed — the DeepStack scatter is extracted to `Transformer.deepstackScatter` with a regression test that fails under the old `mask-1` indexing (distinct stream rows 10/11/12 at image positions 1/2/4), and `applyMropeCore` gets an fp16 identity-rotation + rotate_half test whose dtype expect is red under the old hardcoded `.bfloat16`. Bindings used by the tests (`mlx_array_data_float16`, `mlx_default_gpu_stream_new`) exist; the refactor is a pure hoist of the previous inline body (only `self.s` → `s`).

🤖 **MLXServe-PRBot approved** — reviewed at head `f29a583a`, no issues found.

ddalcu:
Re-reviewed at `8208b272`: the delta is a merge of `main` (#350) — the one manual resolution in `src/scheduler.zig` correctly keeps both appended tests — plus the new `tests/test_embeddings_mm.sh`. One finding in the new script.

- `tests/test_embeddings_mm.sh:87` — the script runs under `set -e` (line 24), so when the embedded python block ends with `sys.exit(1)` bash aborts at that command; `PY_RC=$?` (line 143) and the `cat /tmp/test_embeddings_mm.out` failure branch are never reached, and the per-check `PASS`/`FAIL` lines — which were redirected into the file — are never printed.
  Failure: any real regression on the vision path (e.g. check 4 "text+image differs from text alone" failing because the image is silently dropped again) makes the script exit 1 with no output beyond the `===` header: the server is killed by the trap and the reason is lost. The sibling `tests/test_embeddings.sh` avoids this by having its python blocks always exit 0 and parsing the output with `check`; here either drop `set -e` before the block (`set +e` / `python3 … || PY_RC=$?`) or follow that pattern. (`check()` at line 46 is defined but unused, which is the symptom of the same divergence.)

ddalcu:
Re-reviewed at `194eb67c`: the `set -e` finding is fixed — the python block now always exits 0 and emits `TAG 0|1` lines (plus a `DIAG` line on exception), and `mm_check` parses each tag through `check()`, falling back to the DIAG line and server-log tail when a tag is missing, so a failing vision-path check is reported instead of aborting the script.

🤖 **MLXServe-PRBot approved** — reviewed at head `194eb67c`, no issues found.

ddalcu:
🤖 **MLXServe-PRBot approved** — reviewed at head `c1a174c`, no issues found.

ddalcu:
🤖 **MLXServe-PRBot approved** — reviewed at head `42ca9561`, no issues found.

yachen4ever:
Rebased onto current `upstream/main` (3a1ff57) — 10 commits, no merge commits left. Conflict resolutions, all favoring current upstream semantics:

- **embeddings encode loop** (`server.zig`): upstream's empty-input guard + the chat-template path now compose, with the fallback arms keeping upstream's per-arm `raw_ids` ownership (the block-level defer this PR originally shipped would have freed the escaping slice in the plain-pooling arm — the same finding round-2 review made; the ownership comment now lives where upstream's fuller last-token rationale is kept).
- **chat-template gate**: kept this PR's family gate (`useChatTemplateEmbeddings()`), which upstream's later embedding work made load-bearing.
- **eval cadence**: upstream's `evalCadencePoint` (which also evals SSM entries on hybrid trunks) replaces this PR's plain `mlx_array_eval`; the `MLX_SERVE_MM_DEBUG_LAYERS` dump hook is preserved on top.
- **M-RoPE dtype**: upstream's `actDtype()` covers this PR's dtype-preservation fix generally (prism packs serve in their own numerics); the rationale comments are merged.
- **runEmbedRequest**: the multimodal branch now sits AFTER upstream's unconditional `resetCache()` (e28722d), so vision embeddings get the same per-request KV reset.
- **tests**: the two ownership-contract tests were adapted to the current scheduler shell (config/tokenizer moved onto `LoadedModel`, new `mtp_head_kv_quant` / `inflight_prefill_expected` counters).

Verified on the rebased tree: `zig build test` 2603 pass / 0 fail; `tests/test_embeddings_mm.sh` against `Qwen3-VL-Embedding-2B-4bit` — all five checks pass (image-only, text+image, self-consistency cos=1, image-contributes, pure-text regression).

ddalcu:
🤖 **MLXServe-PRBot approved** — reviewed at head `95a0e94a`, no issues found.

Delta from the last head is a rebase plus the `raw_ids` ownership made explicit: each `wrapEncoderIds` arm frees its input after wrapping and the pass-through arm hands the slice to `seqs`, so neither path double-frees or leaks. The stray unindented `.vision_encoder` line from the previous rebase is fixed.

PR #352, opened 2026-09-04
perf(qwen4): verify widths gather the QSA-selected rows — long-context MTP with 8-bit KV

beamivalice:
Stacked on #350 (the first 14 commits are that PR; this one adds the last 5). Merge #350 first and this rebases to five commits.

## Summary

At long context, an MTP verify forward on Qwen3.8-Flash-Next (`qwen4_exp`) read the **whole** KV cache in every full-attention layer: verify widths (S = 2..15 rows) built a dense `[S, kv]` sparse-attention mask over `denseView`, and with `--kv-quant 8` that also dequantized the entire stored range into a fresh bf16 copy per layer per forward. Decode (S = 1) and prefill (S ≥ 16) already gathered only the selected blocks; verify widths were the gap.

This adds the S = 2..15 arm: the union of the rows' top-512 block selections (fixed-size, deduplicated, host-sync-free) plus the covering tail, gathered rows only (for an affine cache: packed rows + their scales/biases, dequantized as a slab), a per-row `[S, R]` membership mask, and the existing split masked SDPA over R instead of kv. Engages past 16k tokens (`MLX_SERVE_QSA_VERIFY_GATHER_MIN_KV`), kill switch `MLX_SERVE_QSA_VERIFY_GATHER=0`.

## Measured (M5 Max 128 GB, mixed-4-8bit pack, 62.7k-token prompt, forced depth 5, cold boots, counterbalanced, cache off)

| | mask arm (before) | gather arm |
|---|---|---|
| verify forward, `--kv-quant 8` | 56.9 ms/round | 43.8 (−23%) |
| MTP tok/s, `--kv-quant 8` | 40.4 | 47.4 (+17%) |
| serial decode (S=1, untouched) | 54.2 | 54.3 |
| short-context control (kv ≈ 60) | 36.9 ms/round | 36.9 |
| resident bytes | 76.47 GB | 76.49 GB |

Against bf16 KV the arm closes ~90% of the kv8 verify penalty (control-normalized eval 1.54 → 1.19 vs 1.15 dense), so 8-bit KV keeps its 13 GB saving at 1M context without paying for it per round. MTP at depth 5 still trails serial at 62.7k (47 vs 54); this makes it much cheaper, not winning, and it stays opt-in.

**Equivalence.** Greedy MTP vs serial on the same prompt diverges at near ties (top-2 gaps 0.000 and 0.125 nats) in **both** arms, at the identical token on one nonce and earlier in the control arm on the other, so the fork is the known MTP class, not this arm. Rows gathered at 62.7k: 12,296 of 62,760 at S=6.

## Also in this PR

- Diagnostics: QSA arms log one-shot **decline reasons** and per-(arm, width-bucket) engagement with kv, plus a per-request `[qsa-arms]` tally (`qsa=verify_gather calls=549 mask=26 decode=55 verify=286 prefill=182` observed live) — the old one-shot meters could not tell a 62k verify call from a 16k decode one.
- Fix: that tally was wired only to the legacy generate() path; the serve path (`scheduler.finishSlot`) emits `[spec-stats]` too. A scan pins the seam.

## Tests

- `zig build test -Doptimize=ReleaseFast` on the merged tree: 9/9 steps, 1890 passed, 153 skipped, 0 failed.
- Hermetic: the union/mask planner vs a brute-force dense `[S, kv]` mask (7 geometries incl. ragged tails; strict ascending rows so no key can be summed twice); MLX parity vs the dense-mask arm on random dense **and** affine-8 `KVCache`s (S ∈ {2,5,7,15} × kv ∈ {4096, 20000}, cos ≥ 0.99999, max abs diff ≤ 4.9e-4); scan pins for the selection arm, dispatch order, and the prefill gather's new q_len floor (it had none and would have served S=1 declines with a full-cache read).
- Two harness traps recorded in `docs/gotchas/engine-mlx.md`: fixed nonces + a persistent disk tier warm-start later boots; `du -sk` is not a content fingerprint.


ddalcu:

Reviewed the delta on top of #350 (`94c03da..494cb1e`); #350's own findings were posted there.

- `src/transformer.zig:7836` — `qsa_arms: QsaArmCounts` lives on the shared `Transformer`, but `[qsa-arms]` is documented and tested as a per-REQUEST meter ("what ran on THIS request at THIS context length"), and `qsaArmsTake()` resets it on whichever slot finishes first. qwen4 text slots batch-decode on the same `Transformer` (`forwardMoeBatchedDecode`, mask arm), and a cold prefill interleaves with other slots' decode ticks, so the tally a finishing request prints is the union of every concurrent slot's arm notes since the last finish — and the next request to finish gets a near-empty tally.
  Failure: two concurrent chats, one at 40k context (verify gather engaging) and one short (mask arm): the short request's `[qsa-arms] qsa=verify_gather calls=…` claims the gather served it, while the long one later logs `qsa=mask` from the handful of layers that ran after the reset — exactly the misattribution the meter was added to rule out. Move the tally onto the `Slot`/`Generator` (or key it by `ctx`) so the seam that logs it owns the count.

beamivalice:
Addressed in `015cb33` (branch rebased onto the updated #350; the delta is now six commits). `QsaArmCounts` lives on `ForwardCtx`, which is the request's own state (`Generator.ctx`; the batched tick hands each slot's ctx to the forward), so every increment in the full-attention layer goes through `ctx` and `logQsaArms` reads the request's own tally. The `Transformer` field and `qsaArmsTake` are deleted. Batched decode, which previously counted nothing because its branch returned before the arm section, now bills its mask arm once to every participating slot's ctx. Tests: two ctx owners with interleaved gather/mask calls never mix and a take on one leaves the other intact; a comptime/scan pin that no `Transformer` field named `qsa_arms` remains and every increment reads `ctx.qsa_arms`. One documented scope limit: the qwen4 MTP head's own QSA layer is not counted (it builds its own ctx); the tally is trunk-scoped.

Full suite on the rebased branch: 9/9 steps, 1896 passed, 0 failed.

ddalcu:
🤖 **MLXServe-PRBot approved** — reviewed at head `015cb33`, no issues found.

Issue #353, opened 2026-09-04
Cold prefill past ~400k tokens is admitted by the memory preflight and dies on an uncatchable Metal OOM (buffer growth copy and SSM checkpoints unbilled)


beamivalice:
**Observed** (Apple M5 Max 128 GB, Qwen3.8-Flash-Next `mixed-4-8bit`, main at fa960f1, flags `--ctx-size 1048576 --kv-quant 8 --mtp --mtp-depth 5 --prefix-cache-entries 4 --prefix-cache-mem 24GB`): a cold prefill of **458,832 tokens was admitted by the memory preflight and then killed the process** mid-prefill with an uncatchable Metal OOM (`Command buffer execution failed: Insufficient Memory (kIOGPUCommandBufferCallbackErrorOutOfMemory) at transforms.cpp:15`), no 503, connection dropped. The previous request (393k tokens, cold) had succeeded at `peak_bytes` 103.0 GB with 26.6 GB reported available; the hot cache held one 6.5 GB entry.

**What the guard bills vs. what the prefill touches.** `server.checkAttentionMemory` → `prefillMemoryNeeded + qsaMaskBytes + seq × qsaHistoryBytesPerToken` compared against `currentGpuMemoryCeiling − active_mem`. At 458k it billed ~18.1 GB against 25.4 GB available and admitted. The QSA prefill workspace is already billed (and row-chunked to a 256 MB score sheet), the resident hot-cache entry is correctly inside `active_mem`, and the allocator cache is cleared per chunk. Two terms are not billed:

1. **Buffer growth coexistence.** Neither the KV cache (`KVCache.growQuantBuf`: `mlx_zeros(new_cap)` + `slice_update(old)`) nor the QSA indexer history (`mlx_concatenate_axis(entry.aux_state, k_raw)` per chunk per full-attention layer) grows in place; the old and new buffers coexist inside the chunk's lazy graph until its eval. `PREFILL_RUNTIME_FLOOR_BYTES` documents this ("old + new buffer coexist across a grow") and prices it at a flat 512 MB, calibrated on prompts of a few thousand tokens. At 458k on this arch (12 attention layers × 1088 B/token at 8-bit + 3840 B/token of history) it is **~7.75 GB**.
[...]

PR #354, opened 2026-09-04
fix(prefix-cache): a commit from a restored prefix inherits the donor's checkpoints; a hybrid miss is never silent (related: #353)

beamivalice:
Independent of #350/#352 (based on current main, touches only `src/prefix_cache.zig` and one new helper in `src/transformer.zig`). Found by a 64k-step context ladder on Qwen3.8-Flash-Next; the OOM found in the same run is filed separately as #353.

## Summary

A long hybrid-model session could suddenly re-prefill its whole prompt from scratch (560 s at 393k tokens on an M5 Max) with **nothing in the log**.

**Cause.** Checkpoint inheritance in `commitWithMediaState` only ran on the *replace* path, i.e. when the new prompt is a strict prefix-extension of an existing entry. Two requests that share a prompt but generate different tails (any pair of turns, or a speculative and a serial request on the same prompt) commit `P ++ genA` and `P ++ genB`; neither is a prefix of the other, so the second entry was appended fresh and carried no restore points. Its own always-on snapshot is placed by the *local* prefill length (`ssmSnapshotBackoff` on a ~30-token tail after a full restore), so it lands at the prompt end, unreachable for any later extension. Once the budget evicted the first entry, `findBestRestorableMatch` skipped the survivor and the lookup returned null silently.

**Fix.**
- A fresh commit inherits the best key-compatible donor's checkpoints below the shared prefix, refcount-shared (never copied) and bounded by the hot-cache byte budget, highest positions first. The disk tier persists what the RAM entry holds, so it inherits transitively.
- A miss that had a raw token match is logged: `[hot-cache] hybrid miss (no checkpoint <= N of M in K entries)`.
- The replace path's checkpoint merge is factored into one helper, whose error path no longer leaks the un-moved tails.

**Pre-existing**: `git log fa960f1..015cb33 -- src/prefix_cache.zig src/kv_disk_cache.zig` is empty; the new ladder test fails identically on current main (`expected 3, found 1`).

## Tests

- `HotPrefixCache: a commit from a restored prefix inherits the donor's checkpoints` — the ladder end to end (entry A with checkpoints, divergent-tail entry B, evict A, the next rung restores at A's checkpoint). Red before the fix.
- `an inherited checkpoint SHARES the donor's buffers and is budget-bounded` — restore after the donor is freed pins the share.
- `a hybrid miss with a raw token match names itself` + a scan that the no-match arm consults it.
- Full suite: 9/9 steps, 1877 passed, 151 skipped, 0 failed.

One deliberate accounting choice: `current_kv_bytes` is per entry and cannot see sharing, so inherited bytes are billed again until the donor is evicted; a hybrid at the budget edge evicts one entry sooner than before, which is the safe direction for a hard cap.


ddalcu:

- `src/prefix_cache.zig:1050` — the inheritance budget is `max_kv_bytes - new_bytes`, i.e. headroom of an *empty* cache; it ignores `current_kv_bytes`, so the new-entry path (line ~1166 `while (current_kv_bytes + new_bytes > max_kv_bytes) evictOneLru`) can now evict resident entries — including the donor itself — purely to make room for the double-billed shared clones.
  Failure: `max_kv_bytes=100MB`, donor A resident at 60MB (40MB KV + 20MB checkpoints), commit B (30MB own, `replace_idx == null`, shares A's prompt). Budget = 70MB → all 20MB of A's checkpoints are cloned → `new_bytes = 50MB` → `60 + 50 > 100` → LRU eviction drops A (its KV snapshot is real GPU memory, and it was a full-match hit for prompt A). Before this PR `60 + 30 = 90` fit and both stayed resident. The replace path explicitly guards against this ("evict OTHER entries first, then shed checkpoints", test at line 2853); the new-entry path has no equivalent, and none of the four new tests exercise inheritance under a non-zero `max_kv_bytes` through `commitWithMediaState`. Bounding by `max_kv_bytes -| (current_kv_bytes + new_bytes)` (or shedding inherited checkpoints before evicting) keeps the "costs no GPU memory" claim in the doc comment true in practice.

PR #355, opened 2026-09-04
Separate starting the server from loading a model (#214)

### LINKED ISSUE #214: "Auto-start on launch" also auto-LOADS the selected model; please split it into two settings (load defaulting OFF)

justinluque:
Builds on @Fe2-O3's draft #215, with his two commits intact and unedited — he wrote the split and flagged it honestly as unbuilt and untested (no Zig toolchain, 2.6 GB free). This is that branch compiled, run and tested on an M-series Mac, plus three more places the same conflation lived and a handful of follow-ups. #215 can close with this.

Closes #214.

## The bug

"Auto-start on launch" is a sentence about a server. What it did was `server.start(modelPath:)`, and `--model` is an eager, blocking load, so a setting whose only promise was "be running when I log in" read the whole selected checkpoint off disk at every login — tens of gigabytes, before anything was asked of it, with nothing in the UI saying so.

The server has always supported the other half (`runHeadlessServe` → `no_initial_load = true`), and the app already hot-loads on the first chat turn (`ServerManager.ensureDefaultChatModel`). The app just never asked for it.

## The rule

`--model` is now passed by exactly one thing: an auto-start you configured. Every other start is headless plus an on-demand load.

That matters beyond login time. A `--model` entry is the registry's *default*, so it comes back on the next request after you eject it — which is why a tray-started server could not be made to sit empty, while an auto-started one could.

## The change

- **Two decisions, two settings.** `StartupModelChoice.launch` is the gate as one pure function (`.doNothing` / `.headless` / `.load(path:)`). Auto-start alone comes up headless; **Settings ▸ Server ▸ "Load a model at start"** buys the eager load.
- **A rule is not one of the items it chooses between.** "Last model used" is a mode in its own stored key, not a magic path in the model dropdown. Under it the dropdown is disabled and shows what the rule resolves to today.
- **One resolution.** The dropdown's readout and the launch gate both call `StartupModelChoice.resolved`, so the control cannot name a model the gate would decline. A model that is gone starts headless — never `--model <gone>`, never a substitute nobody picked.
- **"Last used" means confirmed, not requested** (`recordLoaded` from `transitionToRunning`, the `setDefault` branch of `loadModel`, and `ensureDefaultChatModel`; absolute paths only).
- **Every Start button takes one path** (`AppState.startServer(loadingSelection:)`): headless, then a hot-load through `ensureDefaultChatModel`. Chat always loads — you are looking at a dead composer and about to type; the tray asks the setting. The load is kicked there rather than left to the first turn so the press gets a spinner: `loadingModelPath` is what the pill spins on, and the tray button reports the same fact, since a loading model outranks "running" — the server is up and still cannot answer. Neither button is disabled any more; a Mac with only media models or only LAN peers wants a server with nothing in it.
- **LAN duty was the quiet one.** LAN sharing starts a server at launch even with auto-start off, and that start passed `selectedModelPath`. Launch-time LAN duty now starts what the plan asked for (empty = headless); `ensureServerForLan()` from elsewhere still loads the selection.
- `selectedModelPath` and its stored key are untouched: its `didSet` hot-switches a running server, and editing a *startup* preference must not swap the model out from under a live conversation.

## The default is OFF, deliberately

Including for upgrading users, with no migration. OFF costs one deferred load at the first message — the same load, moved to the moment you asked for something. ON costs it at every login, including the days you never open a chat, and holds the memory throughout.

Turning it back on restores a model going resident with the server, at login and from the tray, minus the trap: it loads on demand, so an eject sticks. Nothing restores `--model` from a button, and nothing should — the default returning after an eject was never a feature.

Happy to flip the default if you disagree; it is one line plus the test that pins it.

## Tests

`cd app && swift test` — 3224 tests, 0 unexpected failures from this change. 27 cases in `StartupModelChoiceTests` cover the gate, resolution, uninstalled/pinned/empty-library paths, the recording rule, pin seeding, LAN duty and the tray, plus three source scans (the launch-time LAN call site, and both halves of the button-start path) — each verified red on revert. `ServerControlButtonPresentationTests` pins the button's wording for a start that loads nothing and for a load in flight.

Pre-existing on `main`, unrelated: 8 `LaTeXFontsTests` / `LaTeXRenderingTests` failures (SwiftPM resource-bundle layout; `fix/latex-font-bundle-layout` is in flight).

No Zig sources are touched.

Also run by hand on an M-series Mac: login with the setting off and on, pinning, uninstalling a pinned model, LAN sharing with auto-start off, both Start buttons, and ejecting a resident model.

## Known, not addressed

The Server section's "Reset to Defaults" resets `ServerOptions` fields only, so the three new app-level settings are exempt. The `.interface` precedent (app-level prefs, no section Reset) doesn't apply to a section that has one. Happy to plumb them in.

## Screenshots

Settings ▸ Server, the three new rows:

image


ddalcu:
🤖 **MLXServe-PRBot approved** — reviewed at head `d35744f0`, no issues found.

ddalcu:
🤖 **MLXServe-PRBot approved** — reviewed at head `7c6792a`, no issues found.

ddalcu:

Re-reviewed at head `0746eb4e` (rebased onto current main; the new work since `7c6792a5` is `212f098`, `a48a3d6`, `0746eb4`). The `waitUntilRunning` fix is correct and the exhaustive switch is safe: `ServerStatus` has exactly the four cases, and every one of the five call sites reaches the wait with the server already `.starting` or `.running` — `startServer` and `ensureRunning` start it inline, `refreshLanModels` and `TerminalPane` go through `ensureServerForLan`, and `useModelStartAction` routes `.stopped`/`.error` to `.startExplicitly`. So the new `.stopped` throw can only fire on a real Stop, which is the bug. `toggleHeadless` genuinely has no callers left, and the rest of `a48a3d6` is comment compression.

- `app/Tests/MLXCoreTests/StartupModelChoiceTests.swift:251` — dropping `testEveryButtonStartIsHeadlessThenAnOnDemandLoad`, `testTheChatStartButtonDoesNotLaunchWithAModel` and `testTheLaunchTimeLanStartGoesThroughTheStartupPlan` leaves nothing pinning that any start path actually goes headless; the 24 surviving tests all exercise the pure `StartupModelChoice` functions and no test constructs an `AppState` or observes a `ServerManager`.
  Failure scenario: change `AppState.startServer` at `:705` back to `server.start(modelPath: path, options: serverOptions)` — the exact shape of #214, a button that reads tens of gigabytes before anything is asked of it — and the suite stays green. `CLAUDE.md` asks a bug fix for a regression test that is red on revert, and this PR's headline fix now has none. The anti-source-scan rule is the right reason to delete these, but the rule's other half is "test the behaviour", and the seam is already there: `ServerManager.currentModelPath` is `@Published` (`ServerManager.swift:62`), `startHeadless` clears it at `:175` and `start` sets it to the resolved model, both before `launch()` reaches `resolveBinaryPath` — so a test can call `appState.startServer(loadingSelection: true)` and assert `server.currentModelPath == ""` with no process spawned, exactly as `ServerManagerStopTests` already drives a real `ServerManager`.

justauserid:
Is this the issue that solves error popup every time I try to start the MLX Core app?

The error screen comes with this heading `mlx-serve failed to start` and then a text like this:


> Not enough free memory to load the model. Using the most right now: <an app> 1.1 GB · ‎<an app> 501 MB · <an app> 314 MB — quitting these frees about 1.9 GB. You can also turn on Settings ▸ Skip memory preflight, or pick a smaller model.

> Exit code 1. Full server log below — select & copy, or use the Copy Log button.

And then a whole lot of log lines below that.

Even though all I wanted to do was just start the MLX Core app and then decide what to do with it, if anything.

I suspect it tries to load the last used model or some model (and I can indeed see that the model it tried to load needed more free memory).

PS. If that's a different behaviour and it's needed, I can log a separate issue.

PR #358, opened 2026-09-04
docs(integrations): local-orchestrator + remote-specialist OpenCode plugin pattern

bahaehmimdi:
## Summary
- Adds a subsection to the OpenCode integration docs for a pattern not currently covered: a small local model as the primary orchestrator, delegating coding/reasoning to mlx-serve through a real OpenCode plugin tool.
- Includes a minimal, working plugin example (`remote_code` tool) that fetches from mlx-serve directly and writes the result to disk.

## Why not the alternatives already in the docs
The existing OpenCode section (via `OPENCODE_CONFIG_CONTENT`) covers pointing OpenCode's *primary* model directly at mlx-serve — this is a different use case: keeping a small/fast model as the primary and delegating specific tasks out. Two other approaches were tried and didn't hold up:
- Cross-provider subagent delegation (primary on one provider, subagent on another) didn't reliably connect in testing.
- A hand-rolled `tool_calls` relay looped indefinitely because it didn't forward the model's own prior tool call + result back to itself on the follow-up turn.

The plugin approach avoids both: the tool's `execute()` is deterministic code, and OpenCode's own tool-calling machinery (not a guessed schema) handles invocation.

## Test plan
- [x] Verified `remote_code` tool invoked correctly by a local 7B model (`qwen2.5:7b-instruct` via Ollama) across multiple runs
- [x] Verified generated file content is correct and runs
- [x] Verified graceful behavior (no loop/crash) when the remote call fails

### DIFF
--- a/docs/integrations.md
+++ b/docs/integrations.md
@@ -122,6 +122,67 @@ export OPENCODE_CONFIG_CONTENT='{"$schema": "https://opencode.ai/config.json", "
 opencode --model mlx/MODEL_ID
 ```
 
+#### Local orchestrator + remote specialist (plugin pattern)
+
+If your day-to-day model is small and fast (a laptop-friendly local model via
+Ollama or LM Studio) but you want it to delegate coding or reasoning to a
+bigger model on mlx-serve, wiring mlx-serve up as a second *provider* and
+hoping the orchestrator's subagent tooling picks the right one across
+providers is not reliable in practice — cross-provider subagent calls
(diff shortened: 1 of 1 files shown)


ddalcu:
🤖 **MLXServe-PRBot approved** — reviewed at head `3ac04be0`, no issues found.

ddalcu:
sorry, this one is more of a blog post or X post rather than a doc update. But good to know :)

PR #363, opened 2026-09-05
perf(qwen4): long-context bundle — split-K QSA, admission, SSD-first cache

beamivalice:
# perf(qwen4): long-context bundle — split-K QSA, admission, SSD-first cache

Branch `longctx-mega-all`, base `a93e2c0`. Head, commit count and suite counts are in the last section. Seven groups ship together because they change the same memory estimator. **Blast radius is qwen4_exp**: every behaviour change that another arch can reach is gated behind one predicate, `ModelConfig.longCtxGated`, and the other archs are characterization-pinned byte for byte to `a93e2c0`, with one exception: the width-trial period re-read (ae0574e) reaches every layout, applied by the maintainer after measuring the 27B on an M4 Max. The ledger of every such hunk is in `docs/gotchas/server-http.md` ("PR #363 blast-radius ledger"). Genuine bug fixes stay for every arch.

## What is in it

1. **Long-context QSA and the split-K kernel.** QSA plumbing built once per forward. A fused split-K sparse-attention kernel serves the verify widths: a verify forward at 62k is 13% faster at S=6 and 8% faster at S=3. It engages from about 32k tokens at S ≥ 2, so an 8k cell cannot see it. A fused top-512 select applies the tie rule of `torch.topk` exactly. Also `--max-mtp-ctx`, the adaptive serial switch (qwen4 module-owned heads only), MTP head persistence with its cache-step fix, and **#353**: the server survives a Metal working-set OOM with a named 503 instead of `exit(-1)`.
2. **Admission.** The chunk sizer no longer reads the cache ask, so the composed function is monotone. The prefill width is a per-request decision, re-asked after the eviction pass, and the re-asked width is the width that runs. A warm turn is not billed for the KV prefix the restore hands over, and only when the restore checks the entry out. One `[admission]` line on every decision. Auto boots advertise the session the cache floor billed.
3. **SSD-first prefix cache** (qwen4_exp, needs `--prefix-cache-disk`). RAM holds the model plus one session; the disk is the capacity tier. `--prefix-cache-mem` is the idle-entry allowance, a hard cap; `0` keeps nothing idle. An entry leaves RAM only after a durable copy is on disk and the index agrees. Chunks are shared across turns by hard link, landed chunks only. Write-through never runs inside a sub-chunk warm turn. One batched eval per chunk file.
4. **Checkpoint retention** (#330 follow-up, qwen4_exp). Span-preserving thinning at every site, the trim bills survivors, and the trim logs its inputs.
5. **Per-chunk adaptive prefill width** (qwen4_exp). A long prefill re-prices its width at each chunk boundary: down at once under pressure, up one rung after two clean probes.
6. **Restore by move.** On a full-prefix hit the cache hands its buffers to the request, so the first append writes in place instead of copying the whole prefix. Warm TTFT at 768k is 0.83 s.
7. **Round-cost table layout per arch.** The bucket grid and store version belong to the arch. A sidecar-MTP model (the 27B) reads and writes exactly the file 26.9.1 wrote. Its EV planner picks widths as `a93e2c0` did. Its width-trial SCHEDULE differs since ae0574e: the period is re-read every round on every layout, so a narrow start gets re-measured (first trial at round 20 instead of 136 on the M4 Max trace). The cause of the narrow start on main is still open; ebb914e acquits the EV seed. The DFlash/DSpark width chooser inherits the re-read behind the opt-in `MLX_SERVE_DFLASH_CHOOSER=1`, unmeasured. qwen4 warm-starts from the old file. This fixes the 27B regression an external test found on the draft.

## Tests

`zig build test` on the head, run-step cache cleared, exit captured without a pipe: counts in the last section, 0 failed. `cd app && swift build` clean. `zig build -Doptimize=ReleaseFast` exit 0. Four audit passes with adversarial bars ran on this tree; every blocker they raised is closed on the head. An external review at d87a43b listed eleven items; all are closed with tests, listed in the gotchas.

## Correctness bars

- **Split-K kernel.** Parity at two bf16 ulps; split-count invariance; a per-element certification on a live S=6 verify forward at kv 62,754 found 0 row classes worse than the union gather by more than 2 ulp. The greedy first-divergence bar does not discriminate at this length: a bisect over `a93e2c0` and its three ancestors shows the same 0.375-nat flips on every root, so the class predates this PR.
- **#353 and the latch.** A 374k prompt with no `max_tokens` returns 200 where the tree without the fix died 503. A latched failure never ends a request 200; streaming shares the non-streaming error mapping.
- **Admission.** A 786k warm turn admits at about 12 GB against 19 GB free; the same request billed 26 GB and was refused before the fix.
- **27B.** The EV planner functions diff to zero lines against `a93e2c0`; a characterization test replays a recorded cost table and acceptance stream and asserts the `a93e2c0` width decisions. The trial schedule is the one exception above, measured by the maintainer: 4k 75.1 vs 70.5 to 76.2 tok/s, 32k 53.1 vs 55.0, inside the per-boot swing.
- **SSD-first.** Every test injects its free-space probe; the inverse test keeps the RAM entry when the disk declines. A paused writer can no longer hang the suite.

## Judge (M5 Max 128 GB, deployed flags, fixed 120 s idles)

Flags: `--kv-quant 8 --mtp --prefix-cache-entries 4 --prefix-cache-mem 10GB --prefix-cache-disk 100GB`, no `--prefill-chunk`. Rows 4k to 64k are a same-box two-boot control (root_all2, ctx 1M, half-spread error bars); 128k and 256k ran on root_all4 (ctx 1M); 384k to 768k ran on root_all5 (ctx 786432). A context flag moves reservations only, not kernels, so decode compares across rows; TTFT, cache and peak are context-conditioned.

| rung | prefill new tok/s | TTFT | serial tok/s | auto-MTP tok/s | cached | hot cache GB | peak GB |
|---|---|---|---|---|---|---|---|
| 4k | 1436 ±0.2% | 0.20 s | 58.0 ±1.2% | 79.4 ±1.1% | 4071/4102 | 0.4 | 72 |
| 32k | 1325 ±1.3% | 0.52 s | 57.4 ±0.7% | 61.5 ±1.2% | 32396/32427 | 1.8 | 75 |
| 64k | 1304 ±0.4% | 0.93 s | 55.0 ±2.0% | 61.5 ±4.9% | 65634/65665 | 3.2 | 77 |
| 128k | 1292 | 0.26 s | 53.9 | 51.9 | 131176/131207 | 6.1 | 78 |
| 256k | 1112 | 0.38 s | 50.3 | 48.2 | 262068/262099 | 11.9 | 86 |
| 384k | 1162 | 0.50 s | 47.1 | 42.2 | 392935/392966 | 24.0 | 94 |
| 512k | 894 | 0.66 s | 43.9 | 44.6 | 523726/523757 | 20.1 | 104 |
| 768k | 782 | 0.83 s | 38.9 | 38.2 | 784715/784746 | 28.3 | 104 |

Prefill is the rate on new tokens; TTFT is the warm-turn time to first token on a cached prefix; 128k and 384k are chain starts (cold). The pre-#353 base died at 458k, so there is no base column above 256k. Against the recorded base at 4k to 256k: serial within noise or better, auto-MTP +10 to +18% at 4k and 8k, prefill +9 to +34%. The 768k rung was refused by the warm bill on root_all4 and serves on root_all5.

## Known limits

- **1M at kv8 on a 128 GB box** is refused by margin (33 GB billed against 32 GB free); the deployed shape advertises 786432 so harnesses compact where the box serves.
- **The legacy disk tier** keeps its per-turn rewrite; chunk sharing is SSD-first only.
- **ANE gate ordering** in `doLoadOnInferenceThread` is unchanged (opt-in path, not exercisable on this box).
- **Auto-MTP below serial from 512k** in this run: the adaptive switch takes over only after the first rounds of each request. A per-bucket memory across requests is the follow-up.
- **CLAUDE.md** is at its byte cap; splitting `## Rules` into `docs/rules/*.md` is next release's job.

## Levers and kill switches

| lever | default | effect |
|---|---|---|
| `--max-mtp-ctx <n>` | 0 | MTP off past `n` prompt tokens |
| `MLX_SERVE_QSA_ATTN_KERNEL` / `_SELECT_KERNEL` | on | `=0` restores the gather / the composed select |
| `MLX_SERVE_MTP_ADAPTIVE_SERIAL` | on | `=0` disables the serial switch |
| `MLX_SERVE_MTP_HEAD_PERSIST` | on | `=0` restores per-request head reset |
| `MLX_SERVE_MLX_ERROR_LATCH` | on | `=0` restores the fatal `exit(-1)` |
| `MLX_SERVE_KV_RESERVE` | on (qwen4_exp) | `=0` restores grow-on-demand |
| `MLX_SERVE_PREFILL_CHUNK_PER_REQUEST` / `_ADAPTIVE` | on (qwen4_exp) | `=0` boot-time width / no per-chunk change |
| `MLX_SERVE_PREFIX_SSD_FIRST` | on (qwen4_exp + disk) | `=0` RAM-first cache |
| `MLX_SERVE_SSD_CHUNK_SHARE` / `MLX_SERVE_RESTORE_MOVE` / `MLX_SERVE_QSA_HISTORY_SHARE` | on | `=0` write every chunk / copy on restore / two history copies |
| `--prefill-chunk` | auto | an explicit width outranks both choosers |

## Head

Head: `1a1100e` (299 commits on `a93e2c0`). Suite: 9/9 steps, 2232 passed, 153 skipped, 0 failed. Audit addendum 5: qwen4_exp bytes unchanged since the judged head, so the table above is the head's. Guards on `2727ff8` (staged as root_all6): 12 of 12 engine rows pass — adaptive switch 11/11, MTP head persistence 8/8, error recovery 8/8 including a streaming fault, batched equivalence 11/11, the four memory-guard rows (50k, 374k with no `max_tokens`, 512k cold, an injected Metal fault with the server alive after), both auto-context rows, and the hot prefix-cache script. The pad-waste cap fires at 1.97x waste on a live 1k plus 64k pair and the long slot falls to serial. Two warm-turn rows on resident 128k and 512k entries restore 131,166 and 523,716 tokens and are admitted, which is the regression the previous head had. Every remaining non-pass in the run traced to the bench harness, not the tree; those harness fixes are recorded beside the ledger.


beamivalice:
Thank you for the review. It found real defects. This comment gives the state of each finding and the commits that address it. All commits are on `longctx-mega-all`. The head at the time of this comment is `78f0c08` (271 commits on `a93e2c0`).

**27B and every sidecar-MTP arch.** We both found the same root cause. The round-cost table's store version rides the file name. Every existing table went cold on the next boot, and the planner ran the prior with its extension valve open. The fix is in f2cf5fd and d87a43b. The bucket grid and the store version now belong to the arch. A sidecar arch reads and writes exactly the file that 26.9.1 wrote, and it plans byte for byte as `a93e2c0`. A characterization test pins this. Its expected values come from the `a93e2c0` planner. qwen4 keeps its own layout and warm-starts from the old file. The audit diffed the planner functions against `a93e2c0` and found zero changed lines.

**Blast radius.** This PR now follows one rule. Every behaviour change that a non-qwen4 arch can reach is gated behind one predicate, `ModelConfig.longCtxGated`. The other archs are characterization-pinned byte for byte to `a93e2c0`, with qwen3_5 and llama configs. The five items you listed are gated: the KV capacity reservation past 32k, the pad-waste cap's kv-length rule on the other hybrids, span-preserving checkpoint thinning, the admission bill's new per-request terms, and the chunk sizer's two changes with the context bar. A second sweep of the full diff found seven more hunks and gated them. A ledger lists every hunk, its class and its gate site. It is in `docs/gotchas/server-http.md`, section "PR #363 blast-radius ledger". Genuine bug fixes stay for every arch: the double frees, the MLX error latch, and the error mapping.

**Bugs.**
- `spillIdleEntries` evicted every non-MRU entry. Fixed. Spilling and evicting are two decisions. Eviction is bounded by the resolved idle allowance as a hard cap, oldest first. `0` keeps nothing idle.
- Eviction on a silent skip. Fixed. The append path returns `persisted`, `partial` or `skipped`. Only `persisted` can evict, and the index must hold the full prefix.
- `store_declined` before `refreshDiskBudget`. Fixed. The refresh runs first.
- `ssdFirstEnabled()` without a disk tier. Fixed. `ssdFirstActive(config, has_disk)` is the one predicate at the budget site and at the arming site. Disk off is byte for byte the RAM arm.
- S1. The SSD-first budget arm now runs the floor reserve through `clampReserveWidth` and reads the static ceiling.
- The free-space test read the host volume. The probe is injectable through `DiskTier.space_probe`. Every SSD-first test arms it. The inverse test exists: with 10 GiB free, the store declines and the RAM entry stays.
- `postEvictionPrefillChunk` took the wider width. Fixed. The re-asked width is the width that runs, in both directions.
- Streaming error names. Fixed. One mapping serves streaming and non-streaming on all four surfaces. After the headers, an SSE error event carries the same type and message.
- The decode-time latch. Fixed. A latched MLX failure ends the request with the mapped error, never with a 200.
- A drain blocked the inference thread. The fix round introduced it and you found it. A non-blocking pending check replaces it.

**Dead code.** We removed two constants, one changelog entry for an intra-cycle regression, and one stale citation. A reachability count over the 295 declarations that the PR adds found one unreferenced symbol. The rest is reachable, tested code. The size complaint stands. The dead-code complaint does not.

**Measurements.** The final judge ran on the audited head at the deployed shape. The shape uses `--ctx-size 786432`, because a harness compacts at about 90% of the advertised context and a 128 GB box refuses 1M by margin. Rows 4k to 64k are a same-box control at fixed 120 s idles. Rows 128k to 768k are on the judged head, and a later audit confirms that no qwen4 byte changed after it. The table is in the PR body. The split-K verify kernel engages from about 32k tokens at S >= 2. An 8k end-to-end cell cannot see the 13% and 8% per-forward numbers. The body says so.

**Known limits in the body.** 1M at kv8 is refused by margin on this box. The legacy disk tier keeps its per-turn rewrite, because chunk sharing is SSD-first only. The ANE gate ordering is unchanged, since that path is opt-in.

Suite on the head: 9 of 9 steps, 2173 passed, 153 skipped, 0 failed. The judge table and the guards summary are in the PR body, sections "Judge" and "Head".

ddalcu:

- `src/prefix_cache.zig:2085` — The durability bar in `spillIdleEntries` only detects a background write failure that lands *within the same pass* (`d.writeErrors() != errs_before`); a chunk write that fails after the pass has already skipped the entry as `entryWritesPending` is never seen again, because nothing in the writer or the index records which entry the failed blob belonged to.
  Failure scenario: pass N stages entry E's chunks + `meta.json` and skips it (writes pending). The writer then fails `c000003.safetensors` (ENOSPC/EIO → `write_errors += 1`, blob dropped) but, being FIFO, still lands `meta.json`, and the in-memory `IndexEntry` (appended in `writeMeta` before any byte reached disk) keeps a non-zero `chunk_bytes[3]`. Pass N+1 samples `errs_before` *after* that error; `appendCommitWithSpec` hits the superseded branch at `src/kv_disk_cache.zig:1156` and returns `.persisted` without touching the disk; `fullPrefixEntryId` passes on the stale sizes; `entryWritesPending` is false; the error delta is 0 → `spill_durable = true` and `evictAt` drops the only good copy. The next turn restores from an entry missing a chunk — the "lost session" the S3 comment says this check exists to prevent. A fix needs the failure to poison the entry (e.g. the writer records failed paths / the tier invalidates the `e<id>` index entry on a failed blob, or `fullPrefixEntryId` stats each chunk file before the RAM copy is discarded).

beamivalice:
Thank you. The scenario is real, and it is fixed on this branch in 51e40ef (merged as d63f485).

The durability bar had four claims, and three of them were in-memory. A write that failed between two passes had no owner: neither the writer nor the index recorded which entry the dropped blob belonged to, and the error delta only spanned the current pass. Two changes close it.

1. **A failure poisons its entry.** The writer records the path of each failed blob, beside the count. `DiskTier.harvestWriteFailures` maps the path to its `e<id>` and marks that entry poisoned: its `chunk_bytes` are zeroed, `bestMatch` skips it (a restore is a miss), `fullPrefixEntryId` and `holdsFullPrefix` never report it complete, the supersede and extend scans and `chunkShareDonor` ignore it, the `complete on disk` marker never prints for it, and a commit on it returns `partial`, not `persisted`. The directory is reclaimed on the next commit path that may block, never on the spill's non-blocking path.

2. **The spill stats every chunk before it evicts.** `DiskTier.entryWholeOnDisk` compares each chunk file the index names against its recorded size. `spillIdleEntries` harvests failures at the top of the pass, before it samples the error counter, and runs the stat check after the pending and error bars, before any `evictAt`.

Tests, all red first, drive the writer through a failure-injection hook in both phases: your exact sequence (chunk 3 fails after the pass, `meta.json` lands, the next pass keeps the entry, a restore from it misses, and the tier rebuilds it under a new id), the same-pass failure, a truncated chunk under a clean writer, and scan pins that both bars sit in front of `evictAt` and that the marker sits behind the harvest. Suite on the fix branch: 9 of 9 steps, 2180 passed, 0 failed. The merged head runs the same suite once the box is free of a timed A/B, and I will note the result here.

The rule is recorded in `docs/gotchas/server-http.md`: an error counter cannot see a failure that lands between two passes, so the session it lost needs a name.

beamivalice:
Suite on the merged head d63f485, run-step cache cleared: 9 of 9 steps, 2182 passed, 153 skipped, 0 failed. Swift build clean.

ddalcu:
🤖 **MLXServe-PRBot approved** — reviewed at head `1495ed6`, no issues found.

beamivalice:
Two fp16 follow-ups are on the head at 54cbc1c. Both are qwen4_exp only.

**Decode at 16k with `--kv-quant off` was 17% below main.** The new fused sparse-attention entry point ran before the verify gather. Its dense-KV arm was the prefill kernel, called at verify width. That kernel has 8 threadgroups at S=4 and no split-K. On 8-bit KV the quantized arm runs, so kv8 did not show it. The dense arm now declines (`qsaSparseAttnServes`), and the verify gather from `a93e2c0` serves dense KV again. A hermetic dispatch test and a class scan pin it (c02f97f).

**A checkpoint save ran as a synchronous file write inside the prefill chunk loop.** `DiskTier.writeSsmFile` called `mlx_save_safetensors` on the inference thread. It now stages through the background writer like a KV chunk, `meta.json` still lands last, and `MLX_SERVE_SSD_WRITE_THROUGH=0` is the kill switch (dc05942). Measured on the inference thread, the whole mid-prefill write-through is about 5% of the fp16 prefill gap, so this commit does not claim that gap.

A/B for the first fix, fp16, chunk 2048, `llmprobe --bench-only --runs 1`, one boot per arm in the order fix, fix with `MLX_SERVE_QSA_ATTN_KERNEL=0`, base. Decode tok/s:

| rung | fix | fix, kernel off | base 1495ed6 | base vs fix |
|---|---|---|---|---|
| 512 | 98.0 | 97.9 | 92.0 | -6.1% |
| 4k | 104.4 | 96.4 | 91.9 | -12.0% |
| 8k | 97.9 | 97.6 | 88.3 | -9.8% |
| 16k | 93.1 | 87.5 | 79.5 | -14.6% |

The dispatch change is confirmed by the logs. Base ran the fused kernel's dense arm at S=5, kv 16386, and never the verify gather. Both fixed arms ran the verify gather and never the fused kernel. The 16k fix cell matches main's earlier fp16 cell (93.1 against 93.6). The base cell matches the earlier PR cell (79.5 against 77.7).

The size of the win is not clean in this run. The two fixed arms are dispatch-identical on fp16 and still differ by 6% at 16k, which is the noise floor of one boot per arm. Base is also slower at 512 to 8k, where the fix cannot act. A counterbalanced recheck against main, four boots in the order main, PR, PR, main, is running now. Its numbers follow here.

Suite on 54cbc1c: 9 of 9 steps, 2190 passed, 153 skipped, 0 failed.

ddalcu:
🤖 **MLXServe-PRBot approved** — reviewed at head `54cbc1c`, no issues found.

beamivalice:
The counterbalanced recheck was not run. In its place, one fp16 ladder on the head at 54cbc1c: `--kv-quant off`, chunk 2048, `llmprobe --bench-only --runs 1`, rungs 4k to 64k, one boot.

| rung | decode tok/s | prefill tok/s | first token | spec tok/step |
|---|---|---|---|---|
| 4k | 101.8 | 1486 | 2.8 s | 4.27 |
| 8k | 86.5 | 1301 | 6.3 s | 3.15 |
| 16k | 85.0 | 1233 | 13.2 s | 3.00 |
| 32k | 85.4 | 1211 | 27.1 s | 3.49 |
| 64k | 94.3 | 1175 | 55.7 s | 4.47 |

The dispatch fix holds at every rung. The fused kernel's dense arm fired 0 times and the verify gather served, at 32k and 64k as well as at 16k.

Two limits on the numbers. The engine had served 13 requests from an aborted probe before this one, so these are warm-engine cells, and the 8k and 16k decode cells sit below the cold-boot fixed arm in the table above (97.9 and 93.1). The rise from 16k to 64k follows the speculation width the planner picked (3.00 to 4.47 tokens per step), not attention cost. No main boot ran beside it, so this table supports no comparative claim.

beamivalice:
Thank you for `pr363-m4-fixes`. Your commit c1b3a65 is on the head as written, under your authorship, with two commits of ours on top (ebac25f, 2e096a7). The state of each item:

**A, the width-trial schedule.** The mechanism is confirmed. It reaches every sidecar-MTP head and the DFlash width chooser, and the 27B characterization test pins which width the planner picks, not which round, so the change was not covered by it. It is gated on the qwen4 round-cost layout: qwen4 re-reads the period every round, a legacy-layout schedule keeps the a93e2c0 date byte for byte. Two tests pin both arms. The every-arch version is owed to main as its own PR, since the stall is main's bug too.

**C, the PLE prefill prefetch pool.** Reproduced on the M5, larger than on your box. Cold prefill with the pool on, off-vs-default:

| rung | fp16 | kv8 |
|---|---|---|
| 4k | +8.5% | +7.1% |
| 8k | +7.8% | +5.9% |
| 16k | +5.3% | +6.5% |
| 64k | | -0.7% |
| 128k | | -0.1% |
| 256k | | +2.1% |

The pool never measurably pays up to 256k. The only recorded win is the 374k ladder that justified the default, where the serial gather rose from 67.7 to 267.9 ms per 1000 tokens between kv 24k and 355k. So instead of an opt-in, the pool is gated on KV length: the serial walk below `PREFILL_PREFETCH_MIN_KV` (262144), the pool above it. `QWEN4_PLE_PREFETCH_PREFILL=0` forces serial, `=1` forces the pool at every length, `QWEN4_PLE_PREFETCH_PREFILL_MIN_KV` moves the threshold. A short prompt and a long one are pinned to their arms with byte-identical output.

**D, the adaptive floor at 32k.** Taken as written. Both readers sit behind the qwen4 head predicate, and the env override stays.

**CLAUDE.md.** Your trim dropped two names, `billedPrefillChunk` and the round-cost `Layout` mapping. Both are restored, and the file is under the cap.

Suite on the head 2e096a7: 9 of 9 steps, 2194 passed, 153 skipped, 0 failed. Swift build clean.

ddalcu:
🤖 **MLXServe-PRBot approved** — reviewed at head `2e096a7`, no issues found.

ddalcu:
🤖 **MLXServe-PRBot approved** — reviewed at head `ae0574e`, no issues found.

ddalcu:
🤖 **MLXServe-PRBot approved** — reviewed at head `ebb914e`, no issues found.

ddalcu:
🤖 **MLXServe-PRBot approved** — reviewed at head `d799037`, no issues found.

beamivalice:
Two more fixes are on the head at 58ff92f (f89ee72 and 0b8bdc0). Both were found while the deployed head served an agent session at 140k to 273k tokens.

**A Metal shared-event leak, one per decode tick, on qwen4_exp.** After about 80k generated tokens in one process, every request failed with `[Event::Event] Failed to create Metal shared event`. The process held 65,526 of the 65,535 shared events the system allows. Memory was not the cause: 21.9 GB was free. MLX 0.32.2 mints one `MTLSharedEvent` per eval and, in async mode, stamps it on every array of the tape. The event is released only when that array is consumed, read on the host, or freed. `transformer.ropeAngleRows` builds the cos and sin rows for the fused decode QK-norm and RoPE kernel and returns a reshape view of an intermediate array. The view keeps its own reference, so the helper still owns the intermediate, and it freed it only on the error path. Every successful call left one live array. The table is cached per position and rebuilt once per decode tick, so the leak is exactly one array and one event per tick. Only qwen4_exp takes that arm: it is the one shipped arch with head dim 256 and a YaRN attention factor. qwen3_5 takes the early return above it. Main a93e2c0 leaks at the same rate, so this is not from this PR. The fix releases the intermediate on every path and adds the missing release of the view on the error path. The guard is a settled-residency balance across the fault sweep, red at 256 bytes per call before the fix. Live meter on the same workload: 801 events per 1000 generated tokens, linear, before; 62 and flat over 18k tokens after. On a shipped build, `MLX_SERVE_QK_NORM_ROPE_FUSED=0` in the environment removes the leak at the fused kernel's decode cost.

**`--max-tokens` is the serve-mode default for requests that omit the field.** Opencode omits `max_tokens`. The server filled in a sentinel that `clampMaxTokens` resolves to the remaining window, so an unasked-for reply could run to the window. The launch flag, which only fed the offline prompt mode, now sets the default for omitted or non-positive values on all four JSON surfaces through one helper. A request that sends the field keeps it. With no flag, behaviour is byte-identical, pinned by a test. The `/v1/messages` parse no longer casts a negative value into an unsigned integer. The headless flag guard covers the flag.

Suite on the head 58ff92f: 9 of 9 steps, 2203 passed, 153 skipped, 0 failed. Swift build clean. The headless flag guard passes 10 of 10. Live meter on the rebuilt deployment: 447 events after 13,500 generated tokens, flat, where the old build would hold about 10,800.

ddalcu:
🤖 **MLXServe-PRBot approved** — reviewed at head `58ff92f`, no issues found.

ddalcu:
🤖 **MLXServe-PRBot approved** — reviewed at head `9dd536a`, no issues found.

beamivalice:
This round is on the head at 1a1100e (9dd536a to 1a1100e, eight commits). An external audit of the qwen4_exp serving path found eight defects, and the deployed head found two more in live agent sessions. All are fixed with a failing test first.

**Load-time validation, F1 to F4 (qwen4_exp only).** `ngram_size` and `heads_per_ngram` were parsed with no bound and fed fixed arrays, guarded by one debug assert that every shipped binary compiles out. The indexer fields default to zero and the forward divided by the ratio. The n-gram table header was half validated: unchecked JSON unwraps, one region bound, `bits` of 32 accepted. `ple_layer_ids` was read at index zero only, with no check that the layer exists. Each is now a named load error that crosses the inference thread by name, and the PLE placement is asserted after construction. Nine error names, all pinned in the registry mapping. The shipped pack still loads.

**A refused warm turn destroyed its own session, D1 to D3 (qwen4_exp, SSD-first).** At 364k a warm turn restored 364k tokens by checking the cached entry out, admission refused the 3.7k-token append, and slot cleanup dropped the entry. The checkout had released the entry's handles at restore time, so the cleanup saw an empty record. A checkout is now a promise: the handles are donated only right before the generator starts, below every refusal, and a refusal hands the entry back whole. The refusal message printed the evictable bytes while the decision used the reclaimable bytes; one accessor serves both. The bill itself charged the entry's own SSM checkpoints as if the whole prompt were prefilled; it now bills the forwarded span. For the incident turn that is 13.7 GB down to 11.5 GB against 13.4 GB available, so it serves. A flat 8192-token reservation headroom is recorded as a design limit: every second or third warm turn at 364k pays one KV grow.

**Low findings, F5 to F8.** A restore that declined after bumping its protection marker shielded an entry that delivered nothing (F5); the marker is cleared on every zero-token arm, and the lien check now uses the deliverable share. The MTP head load line said the spec wiring was pending (F6). Three environment reads ran on every full-attention layer (F7). The 30 GB n-gram table warm was silent while it ran (F8); it now logs start and progress, says when it is disabled, and reports on `/props` and the metrics gauge.

**One regression, caught at deploy.** The F4 placement check fired on the MTP head's build, which copies the trunk config with the PLE layer set to -1 on purpose, so the real pack refused to load. The check now applies only to a config that asks for a PLE, the invariant stays two-sided, and a test builds the head through the same layer loop with -1. The suite had no test on that path before.

**Launcher.** `mlx-serve launch claude` exports `CLAUDE_CODE_MAX_CONTEXT_TOKENS` from the advertised context; Claude Code 2.1.260 otherwise assumes 200k for an unknown model. The Swift twin matches.

**Budget line.** The request log printed the server's own omitted-budget sentinel as if the client had asked for a billion tokens. It now prints `auto`, or the launch default, and the squeeze warning no longer fires on the sentinel.

Suite on the head 1a1100e: 9 of 9 steps, 2232 passed, 153 skipped, 0 failed. Swift build clean. The deployed server runs this build.

ddalcu:
🤖 **MLXServe-PRBot approved** — reviewed at head `1a1100e`, no issues found.

PR #367, opened 2026-09-06
fix(cache): retain pre-media prefixes across image changes

IridiumMaster:
Follow-up to #354: retain safe pre-media checkpoints across image-key changes/donor eviction, and allow image requests to reuse text-only SSD prefixes.
Long tool histories also thinned away the nearest media anchor; capture before media (including its removable opening marker), retain that anchor and the latest state, and keep existing count/byte caps.
Different media keys never reuse past the earliest known boundary; unknown boundaries remain cold. No new flags, dependencies, disk format or historical-media semantics.

- Apple M2 Ultra, 192 GiB, macOS 26.6.2; Qwen3.8-Flash-Next-Uncensored mixed 4/8-bit, MTP on, KV quantization off; tested base a93e2c0.
- Red→green behavioral regressions; final `zig build test`: 1,906 passed, 153 skipped, zero failures; ReleaseFast build passed. Swift: 3,195 tests, 24 skipped, zero failures (no Swift changes).
- `tests/test_vision_cache_retention.sh`: one-entry RAM baseline a93e2c0 failed on image 3; patched six-image run passed. The 89,890-token interleaved test fails on 15d1fda but passes here: fourth-image reuse 58,368→60,795 (full safe match). Restart/SSD test restores 3,508/4,016 tokens with `[disk-cache] restored` engagement and correct image content.
- DeepSeek Harness: 7 turns, 50 model calls, 3 generations, 2 reference-edit attempts, 11 image-read attempts, peak 131,617 tokens. Long image switches restored all 53,416 / 78,395 / 105,055 safe matching tokens; no large checkpoint fallback/crash/timeout observed. These are agent-driven cache checks, not separate human sign-off or a throughput benchmark; the edit workflow failed its recolors and one visual claim needed correction.

Draft pending integration and revalidation against the newly merged long-context bundle #363 (862bddf), which conflicts in generate.zig/prefix_cache.zig. Current patch is validated/deployed on the named prior base, not on that new bundle.


ddalcu:
🤖 **MLXServe-PRBot approved** — reviewed at head `15d1fda`, no issues found.

IridiumMaster:
Superseded by #373, a draft alternative based directly on current main rather than stacked on this conflicting branch. The checkpoint-only approach still replayed long image-bearing suffixes; the replacement retains ordered Qwen media history and compatible RAM state, with explicit limitations and real-Mac validation in its description. Closing this proposal so there is only one active implementation to review. The branch remains available for comparison; further related changes will stay in #373.

PR #370, opened 2026-09-06
perf(qwen4): long-context follow-ups: MTP/QSA rollback without copies, warm-append admission, error-path handle ownership

beamivalice:
# perf(qwen4): long-context follow-ups: MTP/QSA rollback without copies, warm-append admission, error-path handle ownership

Three topics found while serving the merged head at 140k to 450k, one commit each, minimal diff.

**Admission.** A warm append was billed as a cold prefill on about 28 of every 29 turns, because the credit was all-or-nothing on whether the reservation exceeded the buffer capacity. The credit is now the rows the append holds plus one layer of grow coexistence. The refused 448k turn bills 7.0 GB instead of 13.5 GB. When `iogpu.wired_limit_mb` is raised above the default, it floors the GPU ceiling at the limit minus 8 GiB, the margin under the deepest peak this box has served.

**Error-path handle ownership.** A KV buffer grow freed the old buffer before a fallible zero-fill: a double free. Ten spec-decode fill loops freed uninitialised slots on a mid-loop failure. Class A and B sites leaked handles on unwinds that #353 made reachable. One owner per handle throughout; two class scans over both files.

**MTP/QSA.** A partial-accept rollback re-slices the QSA capacity buffers instead of freeing and re-copying them. The score bank appends columns instead of rebuilding. The MTP head's QSA history takes the trunk's reservation. The pooled-rope cache is keyed on position and table and survives slot teardown through an atomic stale mark that only the inference thread consumes. A `partial_rounds` field on the stats line counts the rollbacks at all four call sites.

Measured against main at 862bddf, one boot each, incremental 4k to 384k, auto-MTP, restored tokens and peak memory identical on every rung:

| rung | prefill tok/s, main → branch | auto-MTP decode tok/s | round ms |
|---|---|---|---|
| 4k | 1503 → 1546 (+3%) | 67.8 → 68.6 (+1%) | 37.0 → 37.7 |
| 8k | 1697 → 1749 (+3%) | 71.3 → 68.1 (-4%) | 44.0 → 33.7 (-24%) |
| 16k | 1508 → 1670 (+11%) | 65.1 → 64.3 (-1%) | 37.9 → 24.8 (-35%) |
| 32k | 1331 → 1559 (+17%) | 67.8 → 70.5 (+4%) | 37.2 → 25.1 (-32%) |
| 64k | 1183 → 1445 (+22%) | 59.2 → 73.4 (+24%) | 42.5 → 35.6 (-16%) |
[...]

PR #371, opened 2026-09-06
fix(engine): error-path handle ownership on the qwen4 decode and prefill paths

beamivalice:
# fix(engine): error-path handle ownership on the qwen4 decode and prefill paths

Thirty-eight sites where an mlx handle was created and a later fallible call could unwind without freeing it, plus one live double free. They became reachable when #353 made MLX failures catchable: before it, every `try` below a handle was dead code.

**What was wrong.** A KV buffer grow freed the old buffer before a fallible zero-fill, so a failure there freed it twice. Ten spec-decode fill loops allocated a slice of handles, put a `defer` over the whole slice, and ran a fallible call inside the loop, so a mid-loop failure freed uninitialised slots. Twenty-three class A sites held a handle under an `errdefer` that ended before the handle's owner took it, or under none. Ten class B sites extracted from a vector into a struct and could unwind after a partial extract. The three C-callback sites were checked and are not leaks: the generated mlx-c wrapper leaves the destination null on failure.

**The fixes.** One owner per handle: `errdefer` until ownership transfers, then `defer`, or an explicit surrender before the consuming call. Fill loops track how many slots are built and free only those. The grow builds first, then swaps.

**Guards.** Two class scans now police both `transformer.zig` and `generate.zig`: no scope arms an `errdefer` and a `defer` over the same handle, and a bare free of a handle whose `errdefer` is still armed surrenders it first. A third scan pins every fill loop to a counter or an infallible pre-fill. Fault-injection tests drive the helpers through every faulted op with a settled-residency oracle, red before each fix; the double free reproduced as a segfault.

**Provenance.** The sites were found by an external model's audit and applied by the same model in its own worktree, then verified line by line by an Opus review that corrected five hunks and added twelve sites, including the eight fill loops. Comments in source are the maintainers' wording.

Suite on the head: 9 of 9 steps, 2175 passed, 153 skipped, 0 failed. Swift build clean.


ddalcu:
🤖 **MLXServe-PRBot approved** — reviewed at head `240045b`, no issues found.

beamivalice:
Folded into #370, which becomes the single follow-up PR with a minimal diff. This one is closed.

PR #372, opened 2026-09-06
perf(qwen4): MTP/QSA history rollback without copies

beamivalice:
# perf(qwen4): MTP/QSA history rollback without copies

Three changes to the QSA indexer state on the qwen4_exp MTP path, from an audit of the serving path, hardened through five audit and fix rounds.

**Q1. A partial accept re-sliced, not re-copied.** A partial-accept rollback freed the QSA raw-key buffer and the pooled bank and re-copied them from the reserve on the next forward: about six prefix-sized passes per layer per round. The audit's shape microbench on this box put those copies at 8 ms per round at 8k and 45 ms at 786k, up to 45% of a round; the end-to-end sweep below does not isolate this term from the other two. The capacity buffers already hold every kept row after the verify forward, so the rollback now re-slices the row count, the way the KV truncate does, and drops only the score bank. The head's truncate got the same treatment. A test pins the same buffer handle before and after a rollback, the pooled bank and the score bank after it, and the next append's bytes.

**Q2. The score bank appends columns.** The bank was rebuilt from scratch every four decode tokens on every full-attention layer, 0.15 to 0.29 ms each at 32k to 383k. It now appends the new blocks' columns. A test compares the incremental bank with a from-scratch build byte for byte and requires the incremental path to have run.

**Q5. The head's history reserved, and a pooled-rope cache that survives slot teardown.** The MTP head's QSA history now takes the trunk's reservation. The pooled cos and sin cache was keyed on the ForwardCtx pointer and rebuilt every draft step; it is keyed on block position and table now. The audit rounds moved the correctness of that cache from a borrowed key, to a reset at request end, to a reset on every slot end, to a mark set by the slot and consumed by the inference thread, to an atomic mark. Connection threads write one atomic word and nothing else; the inference thread frees, at its drain or on the next lookup. Round five found nothing left.

Byte identity is the bar for every change and every test compares bytes. No comment or doc text was written by the external agent.

**Measured against the baseline on main (862bddf)**, one boot each, 4k to 384k incremental, auto-MTP:

| rung | prefill tok/s, main → branch | auto-MTP decode tok/s | round ms |
|---|---|---|---|
| 4k | 1503 → 1546 (+3%) | 67.8 → 68.6 (+1%) | 37.0 → 37.7 |
| 8k | 1697 → 1749 (+3%) | 71.3 → 68.1 (-4%) | 44.0 → 33.7 (-24%) |
| 16k | 1508 → 1670 (+11%) | 65.1 → 64.3 (-1%) | 37.9 → 24.8 (-35%) |
| 32k | 1331 → 1559 (+17%) | 67.8 → 70.5 (+4%) | 37.2 → 25.1 (-32%) |
| 64k | 1183 → 1445 (+22%) | 59.2 → 73.4 (+24%) | 42.5 → 35.6 (-16%) |
| 128k | 1052 → 1227 (+17%) | 52.1 → 62.6 (+20%) | 39.5 → 40.7 (+3%) |
| 256k | 870 → 973 (+12%) | 45.5 → 59.0 (+30%) | 52.4 → 44.4 (-15%) |
| 384k | 940 → 1049 (+12%) | 44.7 → 54.0 (+21%) | 52.1 → 34.4 (-34%) |

Peak memory, restored tokens, prefill chunk and PLE arms are identical on every rung. The partial-accept counter this PR adds was not yet in the build the sweep ran on, and a non-speculative cell (the serial round cost) moved 6 to 8% too, so the gain is not attributed per change. Every cell is one boot and one generation, so the evidence is the sign across eight rungs and the same-width round cost, which fell 19% to 30% in the four buckets with ten or more samples on both sides, not any single percentage. The baseline's step in round time at 256k and above is gone. Auto-mode output is not byte-reproducible across boots by design; the byte bar with `MLX_SERVE_MTP_FORCE_DEPTH` pinned is: byte-identical at 16k and 64k between main and the branch (`MLX_SERVE_MTP_FORCE_DEPTH=4`, adaptive serial off, cold, temperature 0), with identical draft, accept and attempt counts on every request, and reproducible across boots within each arm.

Suite on 627949f: 9 of 9 steps, 2169 passed, 153 skipped, 0 failed. Swift build clean. The qsa test family is 50 tests.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@
 - **A prefix-cache hit no longer leaves Flash Next's draft head guessing.** The head's own history is saved with the cached prefix and restored with it.
 - **A partly-accepted speculative round is cheaper on Flash Next.** When verification keeps only some of the drafted tokens, the sparse-attention key history for the tokens that survived is now re-used where it already sits instead of being copied out and re-seeded on the next token. The saving grows with the conversation length, and the generated text is byte-identical.
 - **Flash Next's sparse-attention block scores are extended, not rebuilt.** Every four decoded tokens the model used to re-transpose and re-convert its whole block-score table; it now writes only the new columns into a table sized once for the request. The scores are byte-identical, and the saving grows with the conversation.
+- **Speculative drafting on Flash Next stopped re-deriving position tables it already had.** The draft head builds a fresh context per draft step, which made the sparse-attention rope tables miss their cache on every step and left the head's own key history without the space the trunk had reserved for it. The tables are now keyed on what they actually depend on, and are dropped when the conversation they belong to ends.
 
 ### Fixes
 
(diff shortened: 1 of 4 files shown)


ddalcu:
🤖 **MLXServe-PRBot approved** — reviewed at head `627949f`, no issues found.

beamivalice:
Folded into #370, which becomes the single follow-up PR with a minimal diff. This one is closed.

PR #373, opened 2026-09-07
feat(vision): append-only Qwen media history and RAM cache reuse

IridiumMaster:
Stateless clients resend media history on every turn; this avoids re-encoding unchanged images and replaying compatible cached prefixes while preserving what the model sees.
Replaces #367, not stacked. Rebased onto upstream `main` at `0814cf3` (13 September); qualification below is for head `dbdc444`.
Retains original-message media placement, ordered pixel identities/M-RoPE, compatible RAM prefix reuse, and bounded preprocessing/embedding caches. No new runtime dependencies or Swift changes.
Post-image state remains RAM-only; disk lookup may restore text strictly before media. Restart/eviction and short hybrid checkpoint tails can still replay.
Incomplete stored media histories and unsupported layouts return errors. Responses WebSocket media-validation failures emit terminal error events; both regressions were reproduced before the fix and pass after it.

- **Human tested:** the user confirms extensive hands-on testing of the previously deployed append-only implementation in DeepSeek Harness. This exact rebase received the fresh agent-driven, real-hardware qualification below; not claiming a new manual pass on this commit.
- M2 Ultra, 192 GiB, macOS 26.6.2; Qwen3.8 Flash Next Uncensored mixed-4/8-bit; 262144 context, MTP on, KV quantization off, 24 GiB hot cache / 8 entries, disk off.
- ReleaseFast build and `zig build test`: 2441 passed, 167 skipped, zero failures. Concurrent MTP streams match solo output exactly; four-stream completion also passes.
- Live media history (16 cases), API surfaces (19 cases), cancellation/isolation, and streaming/non-streaming JSON-schema reasoning checks pass. Covers append/edit/reorder/removal, image/video history and HTTP/WS error termination.
- Actual Harness UI: red image, blue edit/comparison, mid-response steering and third-image recall pass. Steering reused 16322/18654 prompt tokens and both image embeddings; appending image 3 encoded only the new image. These are cache/functionality observations, not a controlled throughput comparison.
- Bounded warm-media check: final four active-memory samples span approximately 0.27 MiB; no new system swap-ins/outs during the check. Not a long-session leak guarantee.

Earlier deployment qualification (not rerun for this rebase): Swift build/tests passed (3305 executed, 24 skipped); the first 14 frozen prior MMLU-Pro items produced identical predictions (12/14, same capped item), and five-minute idle plus post-restart cold/warm history checks passed. Related changes remain in this single PR, now ready for review.


ddalcu:

First review since this left draft, at head `dbdc444f`. Verified in depth and found sound: `media_cache.zig` in full (the LRU's byte accounting cannot underflow — `size > max_bytes` returns early so `max_bytes - size` is non-negative and the evict loop terminates before `orderedRemove` can hit an empty list; `ensureTotalCapacity` always covers the later `appendAssumeCapacity`; `Embedding.clone`/`deinit` are mlx refcount pairs reached only from the inference thread while `media_pixels` carries its own mutex for the connection-thread callers); `media_prefix.zig` in full (`expand`'s `insert_at - read` cannot underflow given the non-decreasing check, and the written length is exactly the `size` computed up front); the whole `prefix_cache.zig` integration — `spans_owned` is nulled on both transfer paths so the `defer` and `freeEntryOwnedState` never overlap, and `sharedLimit` is only ever asked about RAM entries, so the disk tier's "text strictly before media" cap is untouched (a disk entry has null spans, falls to the legacy branch and is capped at `media_start`). I also checked that lifting the commit trim limit to `tokens.len` when spans are present is safe: a trim that lands mid-block restores fine because `vision_rows_before` counts placeholder ROWS (`scheduler.zig:6297`), not blocks. Spans are duped into the slot at `scheduler.zig:643` and freed in `Slot.deinit`, and all three handlers `defer`-free `history_spans`.

- `tests/test_media_history.py` — the three new integration scripts are referenced by nothing except `MULTIMODAL-REDESIGN.md`: no `tests/*.sh`, no workflow, and `tests/CLAUDE.md` (the integration-test matrix) is not touched by this PR.
  Failure scenario: they each need a caller-owned running server plus `--url`, so nothing in the repo will ever invoke them — the append/edit/reorder/removal cases, the WS media-validation terminal-error case and the cancellation case exist as files and run zero times. That is the same gap this PR's sibling #412 hit and closed by adding a matrix row naming the boot, the flags and the invariants each gate pins; `CLAUDE.md` asks for the integration script precisely because this behaviour is HTTP-observable, and a script no one can find is not that script. Two other entries of the same growth policy are also unfilled: `src/media_cache.zig` and `src/media_prefix.zig` are new subsystems with no Layout row in `CLAUDE.md` and no `docs/reference.md` section, and `MULTIMODAL-REDESIGN.md` is a new top-level design doc rather than the "1–3-line rule here + full story in `docs/gotchas/*.md`" the policy names.

Not reviewed: the ~450 changed lines of `src/server.zig` outside the media-history block, `src/responses.zig`, and the bodies of the three Python scripts.

PR #374, opened 2026-09-07
fix(prefix-cache): honest declines, media-keyed entries, restorable SSD tier under tight budgets

codysk:
**What changed / why**

Post-mortem on a live 122k-token agent session (2 GB hot budget, client retries until timeout) found seven stacked bugs that froze both cache tiers while the log claimed commits:

- budget declines exited silently and the scheduler logged `committed N/M` for them → `CommitStatus` (ok / kept_resident / declined); the cancel-commit log fires only on `.ok`, decline reasons and disk-lookup misses are logged
- cancelled prefills shorter than the media position kept a pixel key with no boundary → keys derive from the rows an entry actually covers; with-image and no-image retries share their text prefix again
- the SSD tier blanket-skipped any `vision_key != 0` request → it serves the pre-media text prefix, capped at the request's `media_start`
- checkpoints were budgeted after the chunks and starved to zero on chunk-heavy flushes (544 MB of chunks vs a 512 MB cap) → they come off the top of the flush budget (share ≤ cap/2, write-ahead bound = kv_target; SSD-first's write-beside-the-chunk path unchanged)
- RAM budget declines discarded the candidate → spilled to SSD before discarding (flush floor 4 GB — the client is already gone; the per-flush cap prices a live request's stall)
- the disk hybrid arm ranked entries by raw usable length, not restorable checkpoint — the RAM tier's #312 lesson, which regresses the restore the moment a growing entry takes the lead

**How I verified it** (M4 Pro, macOS 26.2): `zig build test` green except `iogpu.wired_limit_mb is a FLOOR`, which fails identically on ddalcu/main @047b7eb without these changes (expected 24086, found 12934). Live on the real session: a declined 5.9 GB candidate spilled complete in one 991 ms pass (3.97 GB, 8 ssm-cps), and the next retry restored 101376/121956 tokens in 404 ms — after hours pinned at 51200. Seven new guards in prefix_cache.zig.

- [x] Searched open and closed issues, open PRs and recent commits; related: #330 (trim family), #312 (RAM-side ranking this mirrors), #354
- [ ] One fix or feature — one post-mortem family (decline observability + media keys + SSD tier + flush budget); happy to split on request


ddalcu:
🤖 **MLXServe-PRBot approved** — reviewed at head `dced3d3`, no issues found.

ddalcu:
🤖 **MLXServe-PRBot approved** — reviewed at head `9e47ddb`, no issues found.

ddalcu:
thanks, good fixes and good finds. I appreciate the small diff.

Issue #368, opened 2026-09-06
PLE n-gram gather: original prefetch gate fixed; pressure-case follow-up in #375


nikolai-vysotskyi:
**Resolved original diagnosis:** 862bddf already fixed the contradictory prefetch gate. Lowering the current 262,144-KV threshold is not supported by our follow-up measurement. The separate evicted-table threaded gather proposal is tracked in #375. Original report retained below for context.

---

## Summary

`NgramTable.gather` has a `pread` worker pool, but the gate that admits work to
it can never be true during prefill. The gate and the loop inside it contradict
each other:

```zig
// src/qwen4_exp.zig:297
if (self.pool) |p| if (self.fd >= 0 and need <= PrefetchPool.ROW_BUF and
                       row_ids.len <= PrefetchPool.MAX_ROWS) {
    ...
    // src/qwen4_exp.zig:301
    while (start < row_ids.len) : (start += PrefetchPool.MAX_ROWS) {
```

`MAX_ROWS = 64` (line 383). The inner loop is written to chunk a `row_ids` longer
than `MAX_ROWS`, but the outer gate only admits `row_ids.len <= MAX_ROWS` — so it
can never iterate more than once, and the chunking is unreachable.

Prefill passes **65,536** row ids per 4096-token chunk (`n_heads = 16` rows per
token), so the gate is always false and prefill falls through to the fully serial
mmap walk (`for (row_ids, 0..) |r, i| self.row(...)`). The pool only ever serves
decode, where `row_ids.len` is 16.

## Why it costs so much

Each row is stored as **three** separate regions (weight 80 B, scales 10 B,
biases 10 B) at unrelated offsets across 320 M rows. One 4096-token chunk touches
~176,000 distinct 16 KiB pages to read 6.55 MB of useful data — roughly **440x
page traffic amplification** — and the serial walk takes those faults one at a
time on one thread.

## Measurements

M5 Max 128 GB, macOS 26.5, `Qwen3.8-Flash-Next-MLX-Serve-4bit`,
`--prefill-chunk 4096 --prefix-cache-entries 0`, median over chunks, 3 runs per
arm, load average 2.2–3.2.

| arm | ms/chunk | pageins/chunk | what differs |
|---|---|---|---|
| serial, fresh rows | **300.5** | 2,689 | baseline |
| serial, **same** rows again (PTEs resident) | **19.1** | 0 | residency only |
| 16 threads, fresh rows | **36.6** | 2,816 | parallelism only |
[...]

Issue #369, opened 2026-09-06
round-cost learns cross-process contention as round cost, and the trial period it computes makes it self-sustaining


Sinojen:
### Before opening

Searched open and closed issues for `round_cost`, `spec_cost_solo`, `contention`
and `cost table` — no match. #360 is the same failure *class* (a table that
learned something false), different cause.

### Priority

medium (workaround exists: delete both table files and relearn idle)

### mlx-serve version(s)

26.9.1-dev (`862bddf` + three local metrics commits that do not touch this path)

### Hardware + macOS

Mac Studio M4 Max, 128 GB, macOS 26.6.2 (25G83)

### Model + settings

`qwen3.8-flash-next` (`qwen4_exp`, 4-bit, MTP loaded).
Launch: `--ctx-size 140000 --kv-quant 8 --prefix-cache-disk 100GB
--prefix-cache-mem 8GB --prefix-cache-entries 128 --max-concurrent 1 --mtp --metrics`

### Server log tail

```
[spec-stats] mode=mtp attempts=1 accepts=3 avg_per_round=3.00 per_draft_pct=100.0% depth=6 drafted=3 ext_rounds=0 runtime_disabled=false reason=none adaptive=undecided serial_cell=0.00 sync_ms=0.00 round_ms=37.73 two_ms_tok=0.00 one_ms_tok=9.43 verdict_round=0 trials=0 width_trials=0 table=<2k:w1:14.33/4390,w2:27.24/95
```

### Summary

`round_cost` learns contaminated round times on a box where several inference
servers share one GPU, and the contamination is self-sustaining. The existing
guard is intra-process only:

```zig
// scheduler.zig:6074
if (s.legacy_gen) |*g| g.spec_cost_solo = active.len == 1;
```

`solo` is "this server has one active slot". It cannot see the *other processes*
on the machine, so a round timed while something else holds the GPU is folded
into the cell as if the round itself were slow — with `solo == true` throughout.
`observe()`'s own comment ("contention only ever ADDS time, so a busy server
stops teaching the table rather than teaching it a lie") is exactly right; it
just does not hold across process boundaries.

The contending process need not be mlx-serve. On the box below it is usually
LM Studio or a llama-server, which is why a flag derived from this server's own
slot count cannot see it.

### Why it does not heal

Exploration exists — `TrialSchedule`, `EXPLORE_BLOCK = 3` rounds every period —
so this is not "the width is never re-measured". The problem is that
[...]

PR #375, opened 2026-09-07
perf(qwen4): parallelise the PLE n-gram gather at prefill width (+3.3% on 65k)

nikolai-vysotskyi:
Follow-up to #368, rebased onto current main and re-measured there — the earlier
numbers in that issue were taken on `a93e2c0`, before #363 landed.

## What this changes

At prefill width the n-gram gather walks its rows serially:

```zig
for (row_ids, 0..) |r, i| self.row(@intCast(r), out[i * self.dim ..][0..self.dim]);
```

One 4096-token chunk passes **65,536** row ids (16 rows per token), each stored
as three separate regions at unrelated offsets across 320 M rows. That is
~176,000 distinct 16 KiB pages touched to read 6.55 MB of useful data, and the
loop takes every one of those faults by itself on one thread.

The `pread` pool is there, but `PREFILL_PREFETCH_MIN_KV = 262144` gates it to
very long contexts. A typical agentic prompt — Claude Code with MCP tools, which
is where I hit this — is 65k, so the serial walk is what runs.

This patch spreads the same walk over threads (work-stealing on an atomic
counter, 256-row blocks). It is **off by default**: `QWEN4_PLE_PAR` unset or `1`
keeps today's behaviour, so the change is inert until enabled.

## Measurements

M5 Max 128 GB, macOS 26.5, `Qwen3.8-Flash-Next-MLX-Serve-4bit`, 65k prompts with
a unique salt each run, `--prefill-chunk 4096 --prefix-cache-entries 0`, arms
alternating, 3 rounds of 3. The number is the server's own `[prefill: N tok/s]`
line, not wall-clock over the HTTP request.

| arm | median tok/s | per round |
|---|---|---|
| serial (today's default) | **1471** | 1404 / 1488 / 1472 |
| `QWEN4_PLE_PAR=16` | **1519** | 1459 / 1528 / 1522 |

**+3.3 % end to end**, consistent in every round (+3.9 / +2.7 / +3.4 %).

On the gather itself (component timer, median per 4096-token chunk):

| arm | ms/chunk | pageins/chunk |
|---|---|---|
| serial, fresh rows | 300.5 | 2,689 |
| serial, **same** rows again | 19.1 | 0 |
| 16 threads, fresh rows | 36.6 | 2,816 |

The middle row is the control: byte-identical work, only whether the PTEs are
already installed differs, so ~94 % of the gather is fault servicing rather than
compute. The third row is the one that rules out alternatives — the threaded arm
takes *more* pageins and is still 8x faster, so this is overlap, not a cache
accident.

## Things I checked that did not work, so nobody repeats them

- **Forcing the existing pool on at 65k** (`QWEN4_PLE_PREFETCH_PREFILL=1`)
  measured **1358 vs 1471, −2.5 %**. The `pread` pool loses to the mmap walk at
  this depth, so `PREFILL_PREFETCH_MIN_KV = 262144` looks like the right call and
  this is not a threshold problem.
- **`--kv-quant 4`**: 1242 vs 1315 on the previous build, −5.6 %. Dequantising
  inside the sparse gather costs more than the bytes it saves.
- **Raising `HC_FUSED_MAX_ROWS` to prefill width**: 1307 → 579 tok/s, 2.2x
  *slower*. The fused hyper-connection read is decode-shaped and the 16-row cap
  is correct.

## Notes

- Thread count: normalising by pageins, ms per pagein goes 1 → 0.12, 4 → 0.031,
  8 → 0.021, **16 → 0.015**, 32 → 0.023. 16 is around the knee. That sweep is the
  weakest measurement here — treat 16 as a sane default, not an optimum.
- Default and env-var name are yours to change; I kept the `QWEN4_PLE_*` prefix
  to match the neighbours in this file.
- With 74 GB of model plus the 32 GB table on 128 GB, `Pages free` fell to 6.2 GB
  during these runs, so pageins never reach zero even after the warm pass. Unload
  any other resident model before measuring this.


ddalcu:

- `src/qwen4_exp.zig:448` — The new threaded gather arm (`ParCtx`/`parWorker`, work-stealing over 256-row blocks, spawn-failure fallback) ships with no unit test, and `parThreads()` reads `QWEN4_PLE_PAR` with no test seam (unlike `ple_prefill_prefetch_override` for the pool arm), so the arm cannot be exercised by `zig build test` at all; CLAUDE.md's "Feature = unit test that fails without it" applies.
  Failure scenario: a future edit to the block/index arithmetic in `parWorker` (e.g. an off-by-one at the `end` boundary or a `dim` mix-up) would corrupt rows silently under `QWEN4_PLE_PAR=16` while the suite stays green — the existing "4096 rows through the pool equal the direct mmap read" test is the template: add an override (e.g. `ple_par_override: ?u32`) and assert the threaded gather is byte-identical to the serial walk over ≥1024 rows.

beamivalice:
image

llmprobe runs=1 bench-only, M5 Max 128GB - Qwen3.8FlashNext 4/8 bits, main commit (047b7e) vs PR375

nikolai-vysotskyi:
Thanks both — the test is added, and I want to take the benchmark seriously
rather than wave it away.

## Test (54944d3)

`ple_par_override` sits beside `ple_prefill_prefetch_override`, and
`test "ngram prefill gather: the threaded walk is byte-identical to the serial one"`
gathers 4096 rows twice — serial and 16-thread — and asserts the outputs match.
It also asserts the 1024-row floor, so the gate is part of the contract rather
than an implementation detail.

Verified by mutation: introducing exactly the off-by-one at the `end` bound you
named makes it fail. A green run without that check would have meant nothing.

## On the ladder

@beamivalice — the baselines line up with mine, which is the useful part: your
64k main is 1475, mine is 1471 on the same box class. So we are measuring the
same thing and the numbers are comparable. That rules out the boring explanation.

Three things I would want to separate before concluding either way.

**1. `runs=1` cannot resolve this effect in either direction.** My claim is
+3.3 % against a within-arm spread of ~3 %, which is why I took 9 samples per arm
with the arms alternating. A single run per rung is inside the noise by
construction — my number would not have survived that protocol either, and I
would not have posted it.

The noise is visible in your own control column: **32k 1435 → 64k 1475**.
Prefill throughput has to fall with depth on this architecture — the QSA indexer
is O(kv) — so a rise there is spread, not signal. (2k → 4k rising is fine and
expected; fixed per-request cost amortises over more tokens.)

**2. The sign pattern does not match the mechanism.** The gather cost is
proportional to rows per chunk, so it is roughly constant per chunk, while its
*share* of prefill shrinks as attention grows with depth. If the patch does
anything, the effect should be **largest at 2k–4k and smallest at 128k**. Your
ladder shows the opposite — losses at 2k/4k, gains at 32k/128k. That inversion
is not "the patch is bad", it is "this is not resolving the patch".

**3. The pack differs, and here it is not a detail.** You are on the 4/8-bit
pack, I measured on 4-bit gs64 — roughly 7 GB more resident weights. The whole
mechanism is page-fault servicing on the mmapped 32 GB n-gram table, and how many
faults you take depends on how much RAM is left for its page cache. On my runs
`Pages free` fell to 6.2 GB even after the warm pass. A 7 GB swing there could
genuinely change the answer, and that would be worth knowing.

## The measurement I would trust

Rather than argue over end-to-end numbers, the component timer is where the claim
actually lives and it reproduces to ~3 %:

```
QWEN4_PROFILE_FWD=all ... --prefill-chunk 4096
# [qwen4-prof] ple gather S=4096: N ms
```

Serial vs `QWEN4_PLE_PAR=16`, same boot, one 64k prompt. On my pack that is
**300.5 → 36.6 ms per chunk**, with a same-rows control at 19.1 ms showing ~94 %
of it is fault servicing rather than work, and the threaded arm taking *more*
pageins (2816 vs 2689) while being 8x faster — which is what rules out a cache
accident.

If that timer does not move on the 4/8-bit pack, the patch does not help there
and I would rather find out why than defend the number. Two outcomes I would act
on immediately:

- gather drops but end-to-end does not → the win is real but too small to matter
  at your configuration, and the patch should be gated on something narrower than
  "rows ≥ 1024" — or dropped.
- gather does not drop → the table is sitting in page cache on your setup, the
  faults I was overlapping do not exist there, and the patch is solving a problem
  specific to tighter memory. Also a reason to scope or withdraw it.

Happy to run either myself if you tell me the pack and flags; I have the box.

ddalcu:
🤖 **MLXServe-PRBot approved** — reviewed at head `54944d3`, no issues found.

beamivalice:
run your proposed benches pls. @nikolai-vysotskyi

ddalcu:
@nikolai-vysotskyi  also I deleted the 4bit back from HF, the quality of the output is much better with the mixed 4-8 bit, for not much more ram.  So you should be testing with the 4-8 bit.

Let us know the `npx llmprobe@latest` command you used to see the gains. You can pass --runs, --rungs, --bench-only  to speed it up if you like.

It's the tool we should use for this project so we compare apples to apples.  Including the json it outputs would be good, or at least a screenshot from the run  `npx llmprobe@latest --library --open` opens the HTML UI so you can visualize it.

Also, for both, I added a update to main, please merge latest main for a fix on adaptive calculation  https://github.com/ddalcu/mlx-serve/issues/369

13scoobie:
Applied on main as a cherry of the gather path plus a `par_threads_override` test seam (the env is memoised per process). Default stays 1.

M5 Max, mixed Flash Next, `QWEN4_PLE_PREFETCH_PREFILL=0` (serial mmap walk), ~50k prompt, 3 reps, scratch port:

- `QWEN4_PLE_PAR=16` median **1488** tok/s (`[prefill: 1487.9 tok/s]`)
- `=1` median **1375** tok/s

About +8% prefill. Logs: `~/claude-tmp/ple-par/`. This box only: `~/.mlx-serve/launch.env` `QWEN4_PLE_PAR=16` (app merge, process env wins). The pool path past 262k kv is unchanged.

nikolai-vysotskyi:
Status, and one question before I burn the runs.

**Merged.** `3fc6b34` — latest `main` in, clean, no conflicts. Note for
@beamivalice: your ladder was against `047b7eb`, which is now ten commits back.

**Pack.** The 4-bit is gone from HF, so I am pulling
`Qwen3.8-Flash-Next-MLX-Serve-mixed-4-8bit` now (107 GB, ~1.5 h on this line).
Everything below will be on that pack.

**On the pack difference — I had the sign wrong.** I suggested the extra ~7 GB of
resident weights might explain a smaller effect on your pack. It should push the
other way: less RAM left for the n-gram table's page cache means *more* fault
servicing, which is exactly what the patch overlaps. @13scoobie's independent
number on the mixed pack (1375 → 1488, +8 %, `QWEN4_PLE_PREFETCH_PREFILL=0`,
~50k, 3 reps) is consistent with that, and it is a third box.

Worth stating plainly: at 64k kv the pool gate is off anyway
(`PREFILL_PREFETCH_MIN_KV = 262144`), so `QWEN4_PLE_PREFETCH_PREFILL=0` is the
default path there, not a special mode. The arms differ only in `QWEN4_PLE_PAR`.

**What I will run**, once the pack lands:

1. The component timer, since that is where the claim lives —
   `QWEN4_PROFILE_FWD=all`, read `[qwen4-prof] ple gather S=4096`, serial vs
   `QWEN4_PLE_PAR=16`, server restarted between arms, arms alternating.
2. llmprobe, so we are comparing like with like:

```
npx llmprobe@latest http://127.0.0.1:11234/v1 \
  -m ddalcu/Qwen3.8-Flash-Next-MLX-Serve-mixed-4-8bit \
  --bench-only --rungs 8k,32k,64k --runs 5 --label "QWEN4_PLE_PAR=16 r1" \
  --save r1_par16.json
```

with the server rebooted between arms and the arm order flipped each round.
`--runs 5` because a 3 % effect does not survive `runs=1` — mine would not have
either.

@ddalcu — is that the invocation you want, and are `8k,32k,64k` the right rungs
for this? Happy to run the full default ladder instead if you would rather have
it comparable to the existing library entries; it is just longer. I will post the
JSON for every run plus the library screenshot.

ddalcu:
🤖 **MLXServe-PRBot approved** — reviewed at head `3fc6b34`, no issues found.

ddalcu:
yes, also share the mlx-serve start command, so we are all on the same page.
I don't think you need 5, it's excessive, 3 should be enough (the default), 5 will take you too long.

ddalcu:
Measured on M4 Max 128 GB (macOS 26.5). Same setup as yours except the pack: `Qwen3.8-Flash-Next-MLX-Serve-mixed-4-8bit` (the 4bit one is not on this box). 65k prompts, unique random text per request, `--prefill-chunk 4096 --prefix-cache-entries 0`, server's own `[prefill: N tok/s]` line, 3 rounds of 3, boot order alternated, threaded arm confirmed by a log line each boot.

| round | serial tok/s | `QWEN4_PLE_PAR=16` tok/s | delta |
|---|---|---|---|
| 1 | 677 / 610 / 646 | 710 / 708 / 694 | +9.6% |
| 2 | 694 / 694 / 674 | 698 / 697 / 695 | +0.4% |
| 3 | 694 / 693 / 666 | 698 / 697 / 695 | +0.6% |

Steady state it is noise. On this box the 30 GB table stays fully page-cache resident (the warm pass reads it in 4.9 s), so the serial walk has no pageins to overlap. Your 8x on the gather comes from the table being evicted (6 GB free), which is a memory pressure case, not the general one.

The one thing I do see: the serial arm dips (610, 666, 674) and the threaded arm never does, so it does hide a fault tail when the cache is not perfect. Round 1 right after a cold boot is that.

Unit test passes here, nothing broke. I have not measured the evicted case yet (needs a purge between requests).

beamivalice:
Screenshot 2569-09-09 at 16 03 17

QWEN4_PLE_PAR=16 - One more bench from mine, this time runs=3. Decode could be noise, prefill is very slightly positive at 4k. The PR's lever is parallelising n-gram table reads, but the table is page-cache warm 3 s after boot on this machine, so there are no SSD faults left to hide.

I ran #375 on top of my upcoming #388

ddalcu:
Moving this to draft until the pressure case is shown.

Re-ran the review on current main. Main already covers the cold boot (NgramTable.startWarm preads the table at load) and kv >= 262k (PrefetchPool). What is left is a table evicted under memory pressure at kv < 262k, which is where your +3.3% and 13scoobie's +8% come from (6 GB free boxes). On this M4 Max the table stays resident and the arm is noise (+0.4/+0.6%), first post-boot round aside.

The diff is small and the test is real (dropping the last row of each block fails it), but as-is it ships QWEN4_PLE_PAR unset = serial, so merged it is dead code. If it lands:

1. Drop QWEN4_PLE_PAR and parThreads(). Pick the default (threaded at row_ids.len >= 1024, threads = min(16, cpu count)) and keep ple_par_override as the test seam only.
2. notePrefillGatherArm still logs "SERIAL mmap walk" when the threaded arm ran. An A/B arm is proven by its engagement line.
3. The "EXPERIMENT (QWEN4_PLE_PAR, default 1 = today's serial path)" comment is a war story. One line on why (faults overlap across threads).

What decides it: the mixed pack, `MLX_SERVE_NGRAM_WARM=0`, `sudo purge` before every request, 65k prompt with a unique salt, max_tokens 1, 3 reps per arm alternated per boot, plus a no-purge control that must be within noise. Post the prefill tok/s and the `ple gather S=4096: N ms` lines. If the purge arm is >= +3% with the gather timer dropping, it goes in with the fixes above.

nikolai-vysotskyi:
@ddalcu @beamivalice Agreed with the narrower pressure-case framing. The warm-table result is not a general prefill win, and the original +3.3% must not be presented as one. My recent ordinary mixed-pack run measured only **307 ms total CPU gather across a 62,176-token prefill**; older synchronous block timers were not the steady critical-path budget. [Current investigation](https://github.com/nikolai-vysotskyi/mlx-serve/blob/825cff3/research/followups/20260912-prefill-ahead-and-grouping.md).

I accept the requested admission bar for this PR: mixed pack, warmer disabled, purge before each request, alternating three requests per arm at 65K/max_tokens=1, gather timings and a no-purge control. That pressure test has **not** been run yet. I am leaving #375 draft and prioritizing the GPU path and the correctness/admission fixes in #408 for the current 2400 tok/s at 65K target. I will not spend a full llmprobe ladder trying to establish a warm-path gain this patch does not promise.

Before #375 is eligible to land, its default will need to become the automatic >=1024-row threaded path, with the test override retained and the engagement log corrected as requested. No request to merge the current opt-in version. The old claim that prefill throughput must monotonically fall with context was too strong; graph/chunk utilization and workload also matter, so that argument should not have been used to dismiss your ladder.

nikolai-vysotskyi:
Consolidating our remaining prefill work into **#408** and closing this separate PR. The final measured combination reached **2347.1 tok/s at64,947 uncached tokens** in an HTTP screen; that is **not** a validation of this PR's pressure-case claim. [Complete results, reproduction and hypothesis/status report](https://github.com/nikolai-vysotskyi/mlx-serve/blob/2ab09e8/research/followups/20260913-final-prefill.md).

Responding explicitly to the remaining review: the requested automatic >=1024-row / min(16,CPU-count) policy, corrected engagement label and purge/no-purge mixed-pack acceptance have **not been completed here**. Warm-table gains were within noise on the reported M4 tests. The gather/test-seam cherry-pick already reported on main remains upstream; closing this PR does not undo it. No pressure gain or default-on production readiness is claimed. The bounded selected-row PLE-ahead candidate is a distinct opt-in implementation, now available in #408 for review. No further tests are scheduled for this standalone proposal.

[Final consolidated PR message and review escalation](https://github.com/ddalcu/mlx-serve/pull/408#issuecomment-5653886762). Final integrated-head validation: ReleaseFast7/7 and2444 tests passed/155 skipped; merge conflicts resolved.

PR #377, opened 2026-09-07
perf(qwen4): the SSD tier writes the QSA indexer history once per entry, not once per checkpoint

beamivalice:
# perf(qwen4): the SSD tier writes the QSA indexer history once per entry, not once per checkpoint

**The problem.** On qwen4_exp an SSM checkpoint file on the SSD tier carried the QSA indexer history (raw keys and pooled keys, 13 indexers, bf16) as a full copy at every checkpoint position. The history is append-only within a lineage, so checkpoint k's arrays are a byte-prefix of checkpoint k+1's, and the cost was quadratic in turn count: 85.7% of checkpoint bytes were copies, 14.1 GB of unique checkpoint inodes on one 238k-token session, and a single checkpoint at 244k was 995 MB. A long agentic session pushed the tier to about 100 GB on disk.

**The change.** A checkpoint file holds the recurrent state only (GDN state, conv window, the PLE window on GDN layers). The entry's history is written once (`qsa.safetensors`, entry format v7; v2 to v6 entries still load), and a restore at checkpoint k slices it to k rows. The rows adopted are the tensor's own row count, never a recorded one, on every restore path (RAM latest snapshot, RAM interior checkpoint, disk hybrid, SSD-first checkout): a short or missing history is a named miss before any live state is adopted, never a wrong-state restore. A diverging entry hard-links the donor's history file the way KV chunks already inherit. In RAM the entry already held one history; each checkpoint now records its row count so the same slicing applies. Reachable only where an SSM entry carries QSA history; other hybrids write and read their checkpoint files byte for byte as before.

Measured on an incremental 4k to 768k session (M5 Max 128 GB, Qwen3.8-Flash-Next mixed-4-8bit, kv8). The "old format" column is what the same checkpoints would have carried, reconstructed from each checkpoint's position; the v7 columns are the bytes on disk:

| entry | checkpoints | top position | old format | v7 checkpoints | v7 history, once | saving |
|---|---|---|---|---|---|---|
| e5 | 9 | 66k | 1.9 GB | 0.49 GB | 0.23 GB | 61% |
| e6 | 16 | 131k | 5.6 GB | 0.88 GB | 0.47 GB | 76% |
| e7 | 16 | 262k | 10.5 GB | 0.88 GB | 0.94 GB | 83% |
| e8 | 16 | 393k | 15.2 GB | 0.88 GB | 1.41 GB | 85% |
| e9 | 16 | 524k | 21.2 GB | 0.88 GB | 1.87 GB | 87% |
| total, 9 entries, 81 checkpoints | | | 55.2 GB | 4.5 GB | 5.1 GB | 83% |

A checkpoint file is now 56 MB at any position; the only thing that grows with context is the one history file per entry.

Nothing on the forward or decode path changes; the incremental sweep's MLX peaks per rung are identical to main's at 4k to 384k.

Four read-only audit rounds and four fix rounds (a missing or short history file restoring as a hit, a swallowed overlay error, a mid-chunk checkpoint recording live rows instead of its position, the PLE window on GDN layers dropped from checkpoint files) closed with no finding. Suite on this tree: 9 of 9 steps, 2185 passed, 153 skipped, 0 failed. Diff: 3 files.

### DIFF
--- a/src/kv_disk_cache.zig
+++ b/src/kv_disk_cache.zig
@@ -137,6 +137,8 @@ pub const SpaceProbeFn = *const fn (path: []const u8) ?VolumeSpace;
 
 /// Test hook, armed through `DiskTier.armTestSpace`.
 var test_space: ?VolumeSpace = null;
+var test_qsa_overlay_mismatch = false;
+var test_ssm_write_qsa_aux = false;
 
 fn testSpaceProbe(path: []const u8) ?VolumeSpace {
     _ = path;
@@ -181,6 +183,9 @@ pub const IndexEntry = struct {
     /// A background write for this entry failed: never matched, never a donor, never reported
     /// complete; the next commit reclaims its directory. Not persisted (`scan` re-validates).
     poisoned: bool = false,
+    qsa_history_bytes: u64 = 0,
+    qsa_history_rows: u32 = 0,
+    inherited_qsa: bool = false,
     /// In-process LRU stamp; seeded from meta.json mtime order at scan.
     last_used: u64,
 };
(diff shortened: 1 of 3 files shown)


ddalcu:

- `src/kv_disk_cache.zig:689` — pre-v7 entries restored at an interior checkpoint no longer receive any QSA history overlay: their interior `s*.safetensors` carry no aux/`qsa_rows` meta (`want_rows == 0`) and their manifest has no `qsa_history` (`e.qsa_history_rows == 0`), so the gate skips the overlay entirely, whereas the removed code overlaid from the latest checkpoint file.
  Failure: a v3–v6 entry hit below its newest position lands with `entriesHaveQsaHistory == false`, `prefix_cache.zig:1034` (`qsa_history_required`) turns the hit into a cold miss until the entry is re-persisted — contradicting the "v2 to v6 entries still load" claim in the description. No test restores a pre-v7 entry at an interior checkpoint.
- `src/kv_disk_cache.zig:1241` — on the extend path the inherited `qsa.safetensors` link is bounded by `kv_len`, not by the donor/heir common-prefix length, and the manifest stores the donor's full `qsa_history_rows` (line 1302) rather than the prefix-valid rows, so `persistQsaHistory` keeps the link whenever the heir's new history rows are `<= donor rows`.
  Failure: heir committed with its newest QSA checkpoint at r1 ≤ p (common prefix; exactly the shape of the new "prefix-diverging entry hard-links qsa.safetensors" test) then extended with a checkpoint at p < r2 ≤ donor rows keeps the link; `restoreIntoHybrid` at r2 overlays the donor's indexer rows for (p, r2) — the donor's tokens, not the heir's — silently corrupting QSA selection. Storing `min(inherited_rows, prefix_rows)` as the inherited row bound (or dropping the link on any extend past it) avoids it.

beamivalice:
fixes incoming.

beamivalice:
Both findings were real. Fixed on the new head: a pre-v7 entry restored at an interior checkpoint overlays its history from the latest checkpoint file before any state is adopted (a missing or short file is a named miss); the inherited history bound is now min(donor rows, common-prefix rows), stored in the manifest, and an extend past it drops the link and writes a private file. Red-first tests for each: an old-format interior restore, and an heir extended past the prefix restoring its own rows.

ddalcu:
🤖 **MLXServe-PRBot approved** — reviewed at head `69aa4cd`, no issues found.

PR #379, opened 2026-09-07
perf(qwen4): the allocator pool is cleared once between the last prefill chunk and the first decode tick

beamivalice:
# perf(qwen4): the allocator pool is cleared once between the last prefill chunk and the first decode tick

This enabled 768k context with full prefix ram cache for qwen4exp. prefill ~770 tok/s and decode high 30-40.

---

On qwen4_exp the last prefill chunks' freed transient parks in MLX's allocator pool up to its cap (8 GB on the shipped default), and the first decode tick allocates on top of it. Measured with a per-phase memory line on a cold 393k request: pool 0.1 GB at the last chunk, 4.2 GB at prefill done, 8.1 GB at the first decode tick. A cold 524k prompt beside two idle 393k entries died on that tick with a Metal out-of-memory 0.5 GB over the plan ceiling; the parked pool was the difference.

One gated `mlx_clear_cache()` at the prefill-to-decode handover in `scheduler.runPrefill`, through the same clock `advanceStep` writes, behind `ModelConfig.longCtxGated()`; ungated archs return before any MLX call. The existing scan guard on the clear cadence is extended to pin the two clock writers.

Measured at the first decode tick on the same cold request: pool 8.1 GB → 0.8 GB, output bytes identical, decode and prefill unchanged within run-to-run variance on a 2k to 128k llmprobe ladder.

Suite on the rebased tree: 9 of 9 steps, 2195 passed, 153 skipped, 0 failed. Diff: 2 files, 63 insertions.

### DIFF
--- a/src/generate.zig
+++ b/src/generate.zig
@@ -960,6 +960,17 @@ pub fn shouldClearAllocatorCache(step: u32, last_clear: u32, interval: u32) bool
     return step -| last_clear >= interval;
 }
 
+/// PURE: does this request return MLX's allocator pool once, between its last
+/// prefill chunk and its first decode tick?
+///
+/// The per-chunk clear runs before the last chunks' transient is freed, so that
+/// transient parks in the pool up to the cap and the first decode tick allocates
+/// on top of it (8.1 GB parked at the first tick of a 393k prefill). Gated on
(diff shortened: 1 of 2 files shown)


ddalcu:
🤖 **MLXServe-PRBot approved** — reviewed at head `192044c`, no issues found.

PR #381, opened 2026-09-08
perf(qwen4): the QSA raw indexer keys are a 32-row ring; the pooled bank is the history

beamivalice:
**What the raw keys were for.** Every QSA score on qwen4_exp is against the pooled block bank (block means of the indexer keys, one row per 4 tokens); selection is top-512 of `kv/4` blocks and attention gathers trunk K/V by block id. The raw indexer keys (3,072 B per token, bf16, 12 indexers) were kept for the whole context but read in only three places: the rows that close the current block at the next append, the authority check that the history length equals the cache position, and a partial-accept rollback that reopens a block. Everything else that held them (RAM snapshots, checkpoints, the SSD history file, the MTP head's sidecar) was carrying dead weight: 18% of the per-token cache cost, 3.2 GB at 1M, and 80% of the history file on disk.

**The change.** The live entry keeps a ring of the newest 32 raw rows (the verify width plus one block, asserted at compile time against `round_cost.MAX_WIDTH` and the largest compress ratio the loader accepts); the pooled bank is the history and the three guards key on the logical row count (pooled blocks × ratio + ring rows). Every SSM checkpoint stores its own leftover (the rows that close its partial block, at most three per indexer), so a restore at checkpoint k adopts pooled blocks 0..k/ratio plus that leftover as the ring's seed; a checkpoint carries a leftover but never the bank, and one predicate (`checkpointHasQsaPooled`) answers where the bank lives at every site that reads, trims, sheds or hands it off. A rollback past the ring is a named error where MLX would have clamped a negative slice to an empty ring. The MTP head's sidecar persists logical rows with its ring.

**The MTP head.** The head commits its history at the full turn while the trunk restores at its checkpoint, so a warm clamp lands below the head's ring. The head marks its own leftover at every trunk checkpoint position (768 B per checkpoint, at most 24 KB per entry, RAM and disk) and a warm clamp re-seeds its ring from the mark at the restored position; an unmarked clamp is a named miss (`MtpHeadQsaRingGap`, the head starts blind) where the copy fallback used to hand the next prefill an empty ring and fail the request. The marks ride the spec sidecar as v8.

**Disk.** Format v8: the history file holds the pooled bank plus the ring. A v7 file (full raw keys) still restores at any position, interior mid-block positions included, by slicing its own rows; a v8 file on an older binary refuses by name.

**Verification.** Three read-only audit rounds and six fix rounds closed with no finding; two caught old-entry restore regressions (a mid-block interior checkpoint losing its leftover; a v7 file truncated before a restore could slice it) and the last two caught the live head clamp, all pinned by red-first tests that build the live construction (restore as the prefix cache restores at a hit, then append). Live bars on this box (M5 Max 128 GB, Qwen3.8-Flash-Next mixed-4-8bit, kv8, control = main):

- Byte bar, spec off: cold 16k, warm 64k (restored from the 16k entry), warm 256k (from the 64k entry) and a divergent 64k follow-up (interior restore) are byte-identical to the control.
- Byte bar, MTP on at forced depth 4: byte-identical on the same four rungs, with acceptance per round equal to the control's on every rung (1.76 / 1.73 / 1.55 / 2.06); every warm turn logs `MTP head restored`.
- Disk: the 256k entry's history file is 192 MB against 960 MB on the control.
- `tests/test_qwen4_exp.sh`: 40 passed, 0 failed.
- A warm-append ladder 4k → 1M at `--ctx-size 1048576` completed every rung; 1M decoded at 37.7 tok/s with peak MLX active 108.8 GB.

Reachable only where an SSM entry carries QSA indexer history (qwen4_exp); other hybrids write, read and decode as before.

Suite: 2220/2373 passed, 0 failed (153 skipped). Diff: 5 files.

### DIFF
--- a/src/generate.zig
+++ b/src/generate.zig
@@ -2449,6 +2449,11 @@ pub const Generator = struct {
                         doomed.deinit(allocator);
                         return e;
                     };
+                    // The head's own QSA leftover at this position: its ring is 32 rows and
+                    // the clamp back to this checkpoint comes a whole generated tail later.
+                    if (mtp_active) xfm.qwen4MtpMarkQsaLeftover(abs_end_for_cp2 -| mtp_position_base) catch |e| {
+                        log.debug("[qwen4] MTP head leftover mark failed: {s}\n", .{@errorName(e)});
+                    };
                     // Thin the interior, never the oldest (#330): drop-oldest survivors covered only
                     // the last `max * stride` tokens and left no affordable trim point.
                     if (options.ssm_checkpoint_max > 0 and
@@ -2559,6 +2564,11 @@ pub const Generator = struct {
                         var doomed = cp;
                         doomed.deinit(allocator);
                         return e;
+                    };
+                    // The head's own QSA leftover at this position: its ring is 32 rows and
+                    // the clamp back to this checkpoint comes a whole generated tail later.
+                    if (mtp_active) xfm.qwen4MtpMarkQsaLeftover(final_abs -| mtp_position_base) catch |e| {
+                        log.debug("[qwen4] MTP head leftover mark failed: {s}\n", .{@errorName(e)});
                     };
                     if (options.ssm_checkpoint_max > 0 and
                         ssm_checkpoints.items.len > options.ssm_checkpoint_max)
--- a/src/kv_disk_cache.zig
+++ b/src/kv_disk_cache.zig
@@ -219,6 +219,12 @@ pub const SpecHeadMeta = struct {
     pos_base: i32,
     ratio: i32,
     pooled: bool,
+    rows: i32 = 0,
+    /// v8: head rows the sidecar carries a QSA leftover for, ascending — the trunk's
+    /// checkpoint positions. Tensor `<prefix>h.lv<i>` holds the i-th. Inline: an IndexEntry
+    /// is scanned and copied by value.
+    marks: [transformer_mod.QSA_HEAD_MARKS_MAX]i32 = @splat(0),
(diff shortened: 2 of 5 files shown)


ddalcu:

- `src/transformer.zig:6845` — `qsaResliceToKeep` only refuses when `drop_tail > held`, so it reports success in the band where the retained ring survives but is shorter than the leftover the kept position needs (`new_held < keep % ratio`), and the caller never reaches the guard that names that case.
  Failure scenario: MTP on, a committed head at 1002 logical rows with a full 32-row ring; a warm turn clamps the head to 971 (`qwen4MtpTruncate`, `drop_tail = 31`). `qsaResliceToKeep` returns true with `new_held = 1` while `971 % 4 = 3` rows are needed, so `qwen4MtpTruncateToMark` — and with it the `MtpHeadQsaRingGap` decline that would start the head blind — is skipped; the head adopts a history claiming 971 rows whose next forward hits `if (leftover_n > held) return error.QsaHistoryGap` (`src/transformer.zig:16287`) and fails the draft instead of degrading. The same band is open for every `drop_tail` in `[QSA_RING_ROWS - ratio + 1, QSA_RING_ROWS]` that lands mid-block; the existing clamp tests cover `drop_tail = 5` (reslice) and a clamp far past the ring (fallback), never in between.

ddalcu:
🤖 **MLXServe-PRBot approved** — reviewed at head `7f7aaa3`, no issues found.

beamivalice:
Screenshot 2569-09-08 at 09 09 16

PR #383, opened 2026-09-08
fix(qwen4): NUL-truncated prompts, EOS-first spec rounds, tool-call value scans, cache unwind, stream parity, image-turn checkpoints

beamivalice:
## Summary

qwen4_exp serving hotfix. Started from one live incident (an agent session went silent) and grew into a correctness sweep of the paths an agent client exercises, plus a second live incident found while tracking the same session (a text turn after image turns cold-prefilled 348k tokens). One squashed commit; every fix has a red-first hermetic test; the full suite is green. The NUL fix ran the user's live agent session for an afternoon; the rest is hermetically tested and deployed to the same box for live soak.

## Incident 1: a NUL byte truncated the rendered prompt

An opencode agent ran `grep -a` over a binary and the shell output (28 NULs) landed in a user message. `appendJsonString` escapes 0x00 as a JSON unicode escape, nlohmann decodes it back into a real NUL in the rendered `std::string`, the shim returned a bare `char*`, and `renderChatTemplate` read it with `std.mem.span`, which stops at the first NUL. Everything after it never reached the model, including the assistant header, so the model answered with an immediate end-of-turn: `88113+0 tokens streamed [stop]`, HTTP 200, empty reply, and the agent turn ended. Resuming the session logged the identical `prompt=88113` because the new message sat past the cut. Fix: `jinja_render_chat` reports the rendered length and the caller slices by it; the byte-level tokenizer already round-trips 0x00.

## Incident 2: an image turn dropped the checkpoints the next text turn needed

The same session, 43 turns each carrying one screenshot at position 348211, then one text-only turn: `[hot-cache] hybrid miss (no checkpoint <= 348211 of 399199)` and a cold prefill of the whole conversation, which the client abandoned. An image commit lands under the pixel key, so it could neither replace nor inherit from the text entry it had just restored from: `bestCheckpointDonor` demanded exact `vision_key` equality, while the lookup side already caps a key-mismatched entry at its media boundary. Once the count cap evicted the text entry, no checkpoint at or below the boundary existed anywhere, and that boundary is exactly where the next text-only turn is capped. Fix: the donor scan applies the lookup's cross-key rule up to `min(entry.media_start, commit media_start)`, and checkpoint thinning protects the highest checkpoint under an entry's media boundary the way it protects the newest quarter (the thinning class test was red on its own: span-preserving selection had left a 2-token checkpoint under an 8-token boundary). Integration guard: `tests/test_vision_prefix_cache.sh`.

## Fixes found by the follow-up audit (all with red-first tests)

Speculative decoding
- Every block decoder (MTP, DFlash, PLD, gemma drafter, DSpark) checks its entry token before drafting. An EOS first token used to buy a full draft chain plus a verify forward, and the realized-vs-emitted divergence tripped the pad-only guard so the whole prompt prefix was not committed to the cache. Commit policy is now: only an all-pad generation declines.
- A cancel during a batched decode tick records the KV row the forward already wrote, so the committed key matches the snapshot.
- A `[short-gen]` info line names the ids and decoded bytes of any answer of two tokens or fewer.

Tool-call parsing (Qwen `<function=` dialect)
- A parameter value may contain `</parameter>`, `</function>` or `</tool_call>`; first-occurrence scans emptied the value, dropped the parameter or cut the wrapper short. The value now ends at the last close before the next opener, and outer close-tag scans skip values.
- The truncated-call branch reads `<function=` before trying the balanced-JSON arm, so a `write_file` cut mid-`package.json` is no longer renamed after the package.
- `isJsonLiteral` uses JSON's number grammar rather than `parseFloat`; an unquoted `0755` used to make the whole arguments object unparseable and the repair net answered `{}`.
- `appendJsonString` replaces ill-formed UTF-8 with U+FFFD; nlohmann rejected it exactly as a control byte and the prompt silently fell back to the generic chat format.

Prefix cache
- A lookup that fails after binding the entry's buffers now unwinds to an empty cache; the caller reads any error as "no match" and cold-prefills the whole prompt, so a half-bound KV held the prefix twice with wrong positions.
- Commit-path eviction loops exit when nothing is evictable (every entry checked out) instead of spinning the inference thread forever.
- The hybrid SSD restore adopts the MTP/DFlash sidecar from the entry it restored the trunk from, `bestHybridMatch` honours `poisoned`, and manifest scalars are bounded instead of `@intCast`.
- A trim that cannot carry the QSA history over declines instead of committing an entry that can never restore.

Server surfaces
- A spent reasoning budget withholds the rest of the thought instead of streaming it as content (matches non-streaming).
- Non-streaming tool-call replies carry the text the model said before the call, on both the OpenAI and Anthropic surfaces.
- A non-streaming client disconnect reports `client_disconnect`, not the `Slot.init` default `length`.
- Streaming cuts at a stop sequence's match index on all four surfaces, as non-streaming does.
- `seed` (negative = unseeded), `max_tokens` and `reasoning_budget_tokens` are clamped, never `@intCast`.

## Verification

- Full hermetic suite on the squashed tree: see the commit's suite line; the format-corpus and tool-traffic replay suites are part of that run. The image-turn regression test and the thinning class test were each confirmed red on revert.
- Live on the user's box (M5, Qwen3.8-Flash-Next 4/8-bit, 1M ctx, MTP on): the NUL repro message renders in full and answers correctly; the agent session that had gone silent resumed and ran 100+ tool-calling turns to 139k tokens of context with no short-generation stops. The image-turn fix is deployed to the same box for soak; `tests/test_vision_prefix_cache.sh` is the integration guard.

## Not in this PR

- The ds4 engine's own `ds4_chat_append_message` / `tokenizeText` take C strings and share the NUL class; that is a `lib/ds4` submodule change.
- A stop sequence whose match begins in bytes already streamed still leaks its prefix; closing that needs a hold-back across all emit arms.
- `CLAUDE.md` is over its 100k growth cap (it was before this branch); a trim pass is separate work.

Rules added to CLAUDE.md, stories in `docs/gotchas/{tool-calling,server-http,engine-mlx}.md`, CHANGELOG lines under Fixes.


ddalcu:
🤖 **MLXServe-PRBot approved** — reviewed at head `a61e0a0`, no issues found.

ddalcu:
🤖 **MLXServe-PRBot approved** — reviewed at head `25de4d5`, no issues found.

ddalcu:
🤖 **MLXServe-PRBot approved** — reviewed at head `adf34a4`, no issues found.

PR #387, opened 2026-09-09
launch: add opencode2 with the mlx-serve monitor plugin (sidebar stats + footer turn meter)

beamivalice:
## Summary

Adds `mlx-serve launch opencode2`: starts OpenCode 2 (`opencode2`, npm `@opencode/cli`) against the local server and registers the **mlx-serve monitor plugin**, an OpenCode 2 CLI plugin that draws live server stats in the session sidebar and a per-turn speed meter in the prompt footer. The plugin is vendored as a git submodule (`lib/opencode2-mlx-serve`, MIT) and embedded into the binary at build time, so a shipped `mlx-serve` carries it with no source tree.

sidebar panel and footer meter

The existing `opencode` agent (v1 binary) is untouched; OpenCode 2 is a different binary with a different config surface, so it gets its own `AgentKind`.

## How it works

- **Provider config** is the same `OPENCODE_CONFIG_CONTENT` JSON the `opencode` agent already builds; opencode2 honours the same env var.
- **Plugin registration** lives in OpenCode 2's `cli.json`, which is read from `$XDG_CONFIG_HOME/opencode/cli.json`. The launcher writes a dedicated `~/.mlx-serve/opencode2/opencode/cli.json`, merged from the user's own `~/.config/opencode/cli.json` when it exists (theme, keybinds and other plugins are preserved; exactly one mlx-serve entry), copies the embedded plugin to `~/.mlx-serve/opencode2/opencode/plugins/mlx-serve/`, and exports `XDG_CONFIG_HOME="$HOME/.mlx-serve/opencode2"`. The user's real OpenCode config is never written, per the existing launcher rule.
- **Server URL as plugin option.** `options.metricsUrl` is set to `<base_url>/metrics.json`, so the plugin polls the server the launcher actually connected to instead of discovering a port. `metricsToken` is omitted for loopback (the server exempts loopback from `--api-key`) and set for non-loopback URLs.
- **Swift twin** per the documented policy: `LauncherCLI.opencode2`, an OpenCode 2 tab in the setup-instructions panel, and `app/build.sh` copies the plugin sources into `Contents/Resources/opencode2-mlx-serve` for the DMG.

## Files

- `.gitmodules`, `lib/opencode2-mlx-serve` (submodule at `758b5b2`), `lib/opencode2_plugin.zig` (`@embedFile` table), `build.zig` (module wired into the exe and test graph)
- `src/launch.zig`: `AgentKind.opencode2`, `mergeOpencode2CliJson`, `isLoopbackBaseUrl`, `scriptFor` and `writeConfigs` arms
- `app/Sources/MLXServe/Services/{CLILauncher,AgentBudget,CLISetupInstructions}.swift`, `app/build.sh`
- `tests/test_launch_cmd.sh` (case 7), `app/Tests/MLXCoreTests/CLISetupInstructionsTests.swift`
- `docs/integrations.md` (OpenCode 2 subsection), `docs/cli.md`, `README.md`, `CLAUDE.md`

## Tests

Written red-first (failing output recorded before each fix):

- `src/launch.zig`: `fromName` recognizes `opencode2`; cli.json merge keeps theme and unrelated plugins with one mlx-serve entry; a prior mlx-serve entry is replaced, not duplicated; `metricsToken` omitted on loopback, written otherwise, and a known key wins; the script exports `XDG_CONFIG_HOME` and `OPENCODE_CONFIG_CONTENT`, invokes `opencode2 --model mlx/<id>`, carries the install hint and passthrough args.
- `tests/test_launch_cmd.sh` case 7 pins `launch opencode2 --print` against a live server (not executed here: the default test model is not on this machine).
- Swift: tab id, launcher script, and merge pins in `CLISetupInstructionsTests`.

Results: `zig build test -Doptimize=ReleaseFast` 0 failures; `cd app && swift test` 3308 passed, 0 failures; `swift build` clean.

Manual check against a live server on 11234: `launch opencode2 --print` wrote the merged `cli.json` (existing theme, tabs and session settings intact, one plugin entry with the live `metricsUrl` and no token) and a plugin directory byte-identical to the submodule; `check-submodules.sh` is clean after `git submodule update --init lib/opencode2-mlx-serve`.

## Notes for review

- After merging, `git submodule update --init lib/opencode2-mlx-serve` is required before building; `@embedFile` reads the submodule.
- The embed table includes `fixtures.ts` and `probe-live.ts` (test fixtures and a manual probe). They are small and harmless; drop them from `lib/opencode2_plugin.zig` if you prefer a leaner copy.
- The MAS `project.yml` gains no resource folder, so the App Store build finds the plugin only when run from source; the DMG path is covered by `app/build.sh`.
- Requires OpenCode 2 `beta-19296` or later (CLI-plugin API with `sidebar.content` and `prompt.footer.status` slots).


ddalcu:

- `docs/integrations.md:127` — The OpenCode 2 setup omits the `--metrics` prerequisite the monitor plugin's only data source depends on: `metrics_enabled` defaults to false (`src/main.zig:523`), so `GET /metrics.json` answers `503 {"error":"metrics not enabled — start with --metrics"}` (`src/server.zig:1922-1924`) on any CLI-started server, and neither the doc nor `mlx-serve launch opencode2` mentions the flag.
  Failure scenario: a user runs `mlx-serve serve` (no flags), then `mlx-serve launch opencode2`. The launcher writes `options.metricsUrl = http://127.0.0.1:11234/metrics.json` unconditionally (`src/launch.zig:232-234`) with no probe or warning, so the sidebar panel and footer turn meter — the entire point of the plugin — poll a 503 forever with nothing in the launch output to explain it. This is invisible when testing from the app, because `ServerOptions.enableMetrics` defaults to `true` there (`app/Sources/MLXServe/Models/ServerOptions.swift:55`) and passes `--metrics`; the CLI path is the one that breaks.

beamivalice:
> `docs/integrations.md:127` — The OpenCode 2 setup omits the `--metrics` prerequisite … neither the doc nor `mlx-serve launch opencode2` mentions the flag.

Agreed, fixed in a33b2cf.

- `cmdLaunch` now probes `GET /metrics.json` for `opencode2` before launching (and before `--print`). A 503 prints a warning naming `--metrics` and the restart command; 401/403 name the api key; any other non-200 says the panel will read `unreachable`; 200 stays silent. The text is a pure `metricsFeedNote(status)` with a unit test covering all four branches.
- `docs/integrations.md` states the `--metrics` prerequisite, what the panel shows without it, and that the MLX Core app enables metrics by default. `docs/cli.md` gets one sentence.

One nuance on the failure scenario: the plugin does not poll silently. Its Server log section shows `feed --metrics off` in the warning colour, and the footer turn meter keeps working from the streamed response. What was missing was the launcher saying so up front and the docs naming the flag.

Verified: full `zig build test` 0 failures; against a live server with metrics on, `launch opencode2 --print` produces no warning. The 503 branch is unit-tested only, since exercising it live means starting a second model server.

ddalcu:
🤖 **MLXServe-PRBot approved** — reviewed at head `a33b2cf`, no issues found.

PR #385, opened 2026-09-08
perf(qwen4): use cooperative NAX tiles for sparse prefill attention

nikolai-vysotskyi:
Qwen4 sparse prefill attention can keep Q and output fragments in cooperative NAX registers while gathering the same selected KV blocks. This adds an opt-in path for G17, macOS 26.3+, bf16, head dimension 256 and GQA 12; `MLX_SERVE_QSA_NAX=0` retains the stock gather path.

Softmax probabilities are decomposed into three bf16 terms for PV multiplication, retaining float32 precision with relaxed NAX MMA. Block selection and causal tails are unchanged.

Validation on M5 Max 128 GB, macOS 26.5, against main `16a47ec`:
- ReleaseFast build: 7/7 steps succeeded. Full tests: 2230 passed, 153 skipped, zero failures.
- Independent CPU float64 oracle: max FP32-output error 1.61e-7 across batch 2, strided inputs, sentinel rows, ragged causal tails and KV=65599.
- `tests/test_qsa_nax_prefill.py` checks long HTTP prompts and the selected kernel's engagement line.
- Rebuilt with an independent Zig cache after detecting stale build artifacts from a shared cache. Fresh full-model performance checks are in progress; this remains a draft until results can be tied to the rebuilt binary.

PR #388, opened 2026-09-09
perf(qwen4): the QSA block select splits across threadgroups and the score sheet is one NAX kernel

beamivalice:
## Summary

Two kernels for the qwen4_exp sparse-attention indexer, both bit-identical to the chains they replace, plus a per-shape threadgroup policy for the block select. One squashed commit; every piece has red-first hermetic tests, three audits (grok + two Opus rounds per kernel, one Opus round on the merge), and live A/Bs on the M5 Max at 42k, 162k, 361k and 805k tokens.

### The exact top-k block select splits across threadgroups at decode widths

`msv_qsa_select` ran one threadgroup per query row. Right for a 4096-row prefill chunk, wrong for decode, MTP verify and every draft step, where a single threadgroup walked the whole score row four times, latency-bound at ~20x under the bandwidth floor: 0.73 ms per layer at 860k tokens, most of the 7 ms by which a long-context tick exceeds a short one, and paid seven times per depth-6 round. Vectorized loads, unrolling and split histogram banks inside one threadgroup did not move it (occupancy, not ILP). At rows <= 15 and nb >= 24576 each row now splits over 16 slices that run the same exact radix select (one shared body composed into three kernels), and a second dispatch selects over the 16*K candidates. The merge is exact because the order is total (ordinal, then lowest original index): every global top-K element is in its slice's top-K, and candidates reach the merge ascending by original index with sentinel slots skipped, so a cross-slice tie at the K-th score resolves to the lowest index. 0.72 -> 0.26 ms per layer at 215k blocks; ids bit-equal on 1550/1550 probe cases (random, mostly-zero, ties across the K boundary, NaN/-0.0, visibility cut-offs, ragged sizes). `MLX_SERVE_QSA_SELECT_SPLIT=0` restores the single kernel.

The floor: an A/A boot pair measured +4% serial / +5% MTP on identical code, so the floor sits where the win clears that (40k blocks: +4% MTP net; 200k blocks: +18% serial). 16k blocks is the design minimum (`2*G*K <= nb` at G=16).

### The indexer score sheet is one NAX kernel; the f32 score bank retires

The sheet was `astype -> mlx_matmul -> maximum -> sum_axis` over a `[4, rows, nb]` f32 sheet re-chunked to a 256 MB budget: at 1M tokens ~3 s of every 4096-token prefill chunk, all memory bandwidth, linear in context. `msv_qsa_score` (Metal 4 `matmul2d`, bf16 in / f32 accumulate) writes the f32 sheet straight from bf16 `q_rope` and the bf16 pooled bank. Both operands are bf16-valued so every product is exact, and the kernel reproduces the stock tf32 GEMM's 8 ascending 16-deep k-steps and `mlx_sum_axis`'s sequential head order: bit-identical on 100% of elements at every width, ragged nb, a sliced q parent and the strided capacity-buffer bank. Two layouts share one source (`h4` per-head accumulator at rows >= 128, head-interleaved `base` for decode and verify widths). Default on; one predicate `qsaScoreFusedActiveFor(batch, n_idx, idx_hd)` (env + NAX incl. the GPU-family fallback switch + TF32 under atoi semantics + a probe apply armed once in `Transformer.init` on the inference thread + the model's indexer geometry) governs the forward, the row chunker, the prefill transient bill and `qsaScoreBankBytesPerToken`. Under it the f32 score bank (1,536 B/token, rebuilt on every restore) is never built and the sheet budget buys 4x the rows; off, or on an unsupported geometry, the composed chain and its bank run exactly as before. `MLX_SERVE_QSA_SCORE_FUSED=0` restores the composed chain; `QWEN4_DUMP_QSA_BLOCKS=<path>` dumps block ids per QSA layer.

### Second pass on the select

Measured as marginal in-graph cost, the split pair was 0.119 ms per layer at 215k blocks, of which the two launches were 0.022 ms; each radix level cost ~0.010 ms regardless of element count (histogram clear, a single-simdgroup scan over 64 bins, barriers) and the compaction ~0.025 ms in a per-chunk loop every thread ran. Two exact changes: the compaction is a `simd_prefix_exclusive_sum`, and the split kernels plus the single kernel at rows <= 15 use four 8-bit digits (256 bins, 1.3 KiB of threadgroup memory) while prefill widths keep 11/11/10 (8-bit digits lose 3-9% there). Split pair 0.120 -> 0.077 ms at S=1, 0.189 -> 0.106 at S=7; the prefill single kernel unchanged (6.71 vs 6.74 ms at 4096 rows); floor 32768 -> 24576. Bit-equal over the exactness grid on both arms.

805k prompt rerun on this binary: serial decode 46.7 tok/s (stick 46.6), serial cell 20.45 ms (20.44), MTP 65.9 tok/s (57.8, planner variance), cold prefill 1110 tok/s (960; the fused score kernel and the threadgroup policy entered since the stick).

### Select threadgroup as a per-shape choice

256 threads at rows >= 256, 512 at 64..255, 1024 below; `MLX_SERVE_QSA_SELECT_TG` forces one width. Outputs identical at every width.

### The round-cost table is keyed on the engine build

The persisted MTP round-cost table (`~/.mlx-serve/round-cost/rc3-<key>.txt`) was keyed on chip, model, quant and OS build but not the engine build. A binary that makes rounds or serial ticks cheaper restored width cells measured on the old binary next to a serial cell re-measured on the new one; the planner saw every width as relatively worse and drafted narrower (llmprobe's 4k-16k rungs read 85-92 tok/s instead of ~100 on this branch's first run; the same binary on a fresh table read 104 / 92 / 95, identical with the fused kernel off). The key now carries a build id: with `-Dgit-sha` the sha stands for the executable bytes and the binary is not read (a release build's sha is the packager's promise of a clean commit); without one, a dev build hashes its own executable once at load, so an edit-and-rebuild never shares a table. The MLX dylib and metallib are always mixed in by size plus a hash over sampled windows of their bytes, never mtime, so a byte-identical reinstall keeps its table and a rebuilt library rotates it. Within a build, a bucket whose serial cell moves past the planner's own switch margin drops its width cells so trials re-measure them, and the table is never written while a barrier diagnostic (`QWEN4_PROFILE_*`, `MLX_SERVE_MTP_TRACE`, `MLX_SERVE_MTP_FORCE_DEPTH`) is armed. `MLX_SERVE_ROUND_COST_PERSIST=0` still disables persistence.

## Measurements (main `cfb0a3a` vs this branch, same prompts, same launch flags, one run per cell)

| prompt | serial decode tok/s | serial cell ms/tok | MTP decode tok/s | MTP ms/token | cold prefill tok/s |
|---|---|---|---|---|---|
| 361k | 47.7 -> 49.6 | 21.3 -> 19.7 | 57.7 -> 59.0 | 16.5 -> 15.8 | 1115 -> 1168 |
| 805k | 39.5 -> **46.6 (+18%)** | 25.1 -> **20.4** | 53.5 -> 57.8 | 21.3 -> **17.1** | 930 -> 960 |
| 162k (all three) | 50.3 -> 52.8 | | 54.8 -> 60.3 | | 1393 -> 1413 |

Boot-to-boot noise on this box is +4% serial / +5% MTP (A/A pair). Greedy answers are byte-identical between arms at 42k, 162k, 361k and 805k; block ids identical on a real 42k prompt via the dump hook; 40/40 `tests/test_qwen4_exp.sh` cells with both kernels engaged. Full suite 2253 pass, 153 skip, 0 fail.

### llmprobe ladder on the final branch (fresh table by construction)

| rung | decode tok/s | tok/step | first token |
|---|---|---|---|
| 4k | 109.7 | 3.84 | 2.3 s |
| 8k | 109.7 | 4.57 | 4.8 s |
| 16k | 102.1 | 4.27 | 9.9 s |
| 32k | 96.3 | 3.43 | 20.6 s |
| 64k | 106.7 | 3.69 | 42.1 s |
| 128k | 84.1 | 4.00 | 87.2 s |

Headline cells: decode 104.6 tok/s, prefill 1763 tok/s, speculative predictable 142 / novel 70.5 tok/s. The same binary on the stale table read 92 / 85 / 89 at 4k / 8k / 16k.

## Guards

Hermetic: `qsa select split: ids are identical to the single-threadgroup kernel` (exactness grid), the predicate test, the single kernel's exactness test pinned to the single arm, the kill-switch seam test; `qsa score kernel: fused score sheet is bit-identical to the composed chain` (both layouts, sliced-Q parent, prefix-view bank, ragged nb, 262k blocks), ids through `qsaSelectTopBlocks`, geometry fallback, bill-follows-predicate, default-on/off, NAX and TF32 declines, sizing policy, dump hook; `qsa select tg policy`. Rules in CLAUDE.md, stories in `docs/gotchas/engine-mlx.md`, two CHANGELOG bullets.

## Added: the QSA gather attention on NAX cooperative tensors (Nikolai V.'s kernel from #385, default on)

Two more commits on this branch. The first is @nikolai-vysotskyi's `msv_qsa_nax_precise` from #385, carried here with his authorship unchanged: the block-gathered sparse attention of qwen4_exp prefill on M5 cooperative tensors (Metal 4 `matmul2d`, bf16 in / f32 accumulate). The second is our review commit on top of it:

- **Two bf16 softmax terms** instead of three: two match the stock kernel's bf16-store error exactly and beat three on the chained graph (33.5 vs 36.2 ms per 4096-row call); one term is worse than stock.
- **Default on wherever the hardware can run it** (`qsaNaxEligible`: G17 GPU by MLX's own generation parse, macOS 26.3 for the cooperative-tensor API, bf16 q/k/v, head_dim 256, gqa 12, q_len >= 16); `MLX_SERVE_QSA_NAX=0` restores `msv_attn_qsa256`. Never built off NAX; declines a non-GPU stream; `NSG 2` / `BK 32` asserted at compile time.
- **JIT-at-eval fallback**: the Metal compile of a cooperative-tensor kernel runs at first eval, not at construction, so `qsaNaxArm` runs it once on the inference thread and a failure latches the stock gather for the process; a kernel miss falls to the stock gather, never to the dense-mask arm.
- **Not bit-identical to stock**, so the bar is per-element error vs float64 of the bf16-rounded inputs, no worse than the stock bf16 store: `tests/qsa_nax_precision.py` (float64 oracle, bf16-store bar on every case incl. 65599 keys, replaces the C++ oracle) and Zig parity tests against an f32 gather-softmax reference that cross 32-key tiles (a broken online-softmax rescale fails red).

Live on the M5 Max, same binary, env off vs on:

| prompt | stock gather | NAX gather | prefill |
|---|---|---|---|
| 162k | 1480 tok/s | 1667 tok/s | +12.6% |
| 384k | 1322 tok/s | 1465 tok/s | +10.8% |

Isolated gather at S=4096: 2.06x at kv 4k, 1.93x at 65k, 1.64x at 262k. Greedy answers fork after ~200 characters as expected for a non-bit-identical kernel; both arms recall the passphrase. Three audits (Opus, grok-4.6, GPT-5.6) with every P0/P1 closed; suite 2272 pass, 153 skip, 0 fail. #385 can be closed as merged through this PR; the kernel commit keeps its author.

## Next levers (measured, not in this PR)

On the MTP round: the head could select once per round and reuse blocks across draft steps (output-safe), and the EV cost model is context-blind with an unbounded 256k+ bucket.


ddalcu:
🤖 **MLXServe-PRBot approved** — reviewed at head `6849d70`, no issues found.

ddalcu:

- `src/round_cost.zig:851` — `-Dgit-sha` is documented as replacing the executable hash (`build.zig:101`: "overrides hashing the executable") but is only mixed in ahead of it, so `mixExeBytes` reads and FNV-hashes the entire binary unconditionally and the escape hatch does not exist.
  Failure scenario: nothing in the tree passes `-Dgit-sha` (its only two references are the option and the iOS `""`), so every build takes the full-binary read on the first model load inside `doLoadOnInferenceThread` — and a packaging step that starts passing the flag to avoid exactly that still pays it, because the branch that would skip `mixExeBytes` was never written. Two consequences ride along: the key stays sensitive to any byte of the executable, which is what the flag exists to decouple, and `mixMlxArtifacts` folds `libmlx.dylib`'s mtime in, so a reinstall of byte-identical artifacts still rotates the key and discards the learned table the header promises "the same binary across boots shares".

ddalcu:

Nothing new in this head's increment (8-bit digits, simd-scan compaction, the 24576 floor); the finding below is unchanged from `db32755` and still open.

- `src/round_cost.zig:851` — `-Dgit-sha` is documented as replacing the executable hash (`build.zig:101`: "overrides hashing the executable") but is only mixed in ahead of it, so `mixExeBytes` reads and FNV-hashes the entire binary unconditionally and the escape hatch does not exist.
  Failure scenario: nothing in the tree passes `-Dgit-sha` (its only two references are the option and the iOS `""`), so every build takes the full-binary read on the first model load inside `doLoadOnInferenceThread` — and a packaging step that starts passing the flag to avoid exactly that still pays it, because the branch that would skip `mixExeBytes` was never written. Two consequences ride along: the key stays sensitive to any byte of the executable, which is what the flag exists to decouple, and `mixMlxArtifacts` folds `libmlx.dylib`'s mtime in, so a reinstall of byte-identical artifacts still rotates the key and discards the learned table the header promises "the same binary across boots shares".

beamivalice:
Both points taken; fixed at `0b85647`.

- **`-Dgit-sha` now stands for the executable.** When it is set, `engineBuildId` uses the sha and never reads the binary (`buildIdReadsExe` in `src/round_cost.zig`); the option text in `build.zig` says exactly that. Without a sha, a dev build hashes its own executable once at first load so an edit-and-rebuild at the same commit never shares a table. The release packaging owns the promise that a sha build is a clean commit.
- **No mtime in the key.** `libmlx.dylib` and `mlx.metallib` are fingerprinted by size plus an FNV over six sampled 64 KiB windows of their bytes (`fileFingerprint`), so a byte-identical reinstall keeps its learned table while a rebuilt library still rotates it. The metallib read is 384 KiB, not 182 MB.

Tests added: `round_cost: a git sha stands for the executable; without one the executable is hashed` and `round_cost: an artifact fingerprint follows the bytes, not the mtime` (two copies with different mtimes fingerprint equal; one flipped byte differs). Full suite 2261 pass.

ddalcu:
🤖 **MLXServe-PRBot approved** — reviewed at head `0b85647`, no issues found.

beamivalice:
Two commits added on top of `0b85647`: @nikolai-vysotskyi's NAX sparse-attention kernel from #385 (his commit `bd9a2b8`, authorship unchanged) and one review commit of ours (`9db87d9`): two-term bf16 PV, default on wherever the hardware gate holds (`MLX_SERVE_QSA_NAX=0` restores the stock gather), a JIT-at-eval probe that latches the stock gather on a compile failure, and the float64 precision bar in `tests/qsa_nax_precision.py` plus Zig parity tests that cross 32-key tiles. Details and the live table are in the new section of the description: 162k prefill +12.6%, 384k +10.8% on the M5 Max on top of this branch's select and score kernels. Suite 2272 pass / 153 skip / 0 fail. #385 can close as merged through here.

ddalcu:
🤖 **MLXServe-PRBot approved** — reviewed at head `9db87d9`, no issues found.

Issue #212, opened 2026-08-17
MLX Core does not show models held by an mlx-serve process it did not start


Fe2-O3:
### Observed

MLX Core (26.8.8, Homebrew cask) shows no models and reports the server as not running when an `mlx-serve` process was started from a terminal rather than by the app. A model that is resident in that server is invisible in the GUI, so there is no way to see it or unload it from the app.

### Expected

MLX Core lists any reachable `mlx-serve` server, including one it did not start, and shows the models each has resident.

### Reproduction

1. From a terminal: `mlx-serve serve --port 11235`
2. Send a request that loads a model, so the server holds weights in VRAM.
3. Open MLX Core.

The app does not show the server or the resident model.

### The data is already available

`mlx-serve` exposes resident models over HTTP. Against a running server:

```
$ curl localhost:11235/api/ps
{"models":[{"name":"Ling-3.0-tiny-MLX-4bit:latest", ...
  "details":{"family":"bailing_hybrid","quantization_level":"4bit"},
  "size_vram":589824}]}
```

An idle server returns `{"models":[]}`. `/api/ps` is present in the `mlx-serve` binary shipped inside the app bundle.

### What the bundle suggests

From `strings`/`nm` over `Contents/MacOS/MLXCore` (the app was not launched):

- `ServerManager` holds a single private `Process` (NSTask), a single `var port: UInt16`, and `start(modelPath:options:)`, `startHeadless(modelsDir:options:)`, `launch`, `launchCommandLine`, `stop`, plus health polling.
- There is also a `killOrphanedServers` member, so a server the app did not start is treated as something to terminate rather than something to attach to.
- `/api/ps` does not appear anywhere in the `MLXCore` binary. The HTTP paths it does reference are `/v1/models`, `/api/models`, `/v1/models/rescan`, `/v1/unload-model`, and the generation endpoints.

This reads as: the app tracks only its own child process, and never probes for other servers. Stated as a reading of the shipped binary, not of the source.

### Why it matters

A large model can stay resident and invisible. In one case roughly 73 GB of weights stayed pinned in a terminal-started server with no GUI affordance to see or release them, and the machine went deep into swap. The CLI does not help either: `pull`, `list`, `run`, and `serve` are the only subcommands and none of them talks to an already-running server, so `curl` is the only way to ask.

### Suggestion

Probe the standard local port range on launch and list every `mlx-serve` found, each with its `/api/ps` resident models and an unload control, whether or not the app started it.

[...]

PR #389, opened 2026-09-09
fix(app): reach a server bound to a specific interface host

t2tx:
**What changed / why**
When the Settings host is a specific interface address (e.g. a LAN IP), the app's `APIClient` kept hitting hardcoded `127.0.0.1` and silently lost the server it had just started — no health, no model list, no chat. `APIClient` now carries the host, `ServerManager` sets it from `ServerOptions.host` at launch, and every request URL goes through one `serverURL(port:path:)` helper. Wide binds (`0.0.0.0` / `::`) stay on loopback, which keeps the no-api-key trust boundary. Related to #212 (servers the app did not start); this one is the app losing contact with a server it did start.

**How I verified it** (chip, macOS, `swift test`, live run)
M5 Max, macOS 26.5.1, Xcode 26.2. `FAST_DEV=1 app/build.sh`; `swift test` 3314/0 (9 new in `APIClientServerURLTests`). Served `Qwen3.8-27B-MLX-Serve-8bit` with the host set to a specific LAN IP: health, `/v1/models` and a chat completion all answered through that host while `127.0.0.1` correctly refused.

### DIFF
--- a/app/Sources/MLXServe/Services/APIClient.swift
+++ b/app/Sources/MLXServe/Services/APIClient.swift
@@ -84,6 +84,13 @@ struct RetryPolicy {
 }
 
 class APIClient {
+    /// The host address used to connect to the server. Defaults to loopback;
+    /// set to the configured `ServerOptions.host` so the app can monitor a
+    /// server bound to a specific LAN IP (e.g. `192.168.1.10`).
+    /// When the server binds `0.0.0.0`, loopback still works, so the default
+    /// is safe for the common case.
+    var host: String = "127.0.0.1"
+
     private let session: URLSession = {
         let config = URLSessionConfiguration.ephemeral
         config.timeoutIntervalForRequest = 600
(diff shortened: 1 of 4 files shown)


ddalcu:

- `app/Sources/MLXServe/Services/APIClient.swift:104` — An IPv6 host is interpolated unbracketed, so the string is not a valid URL authority (the port component parses as `:1:11234`), and the force-unwrap on `URL(string:)` has nothing to fall back on.
  Failure scenario: the Settings host field is free text (`ServerOptions.swift:582`) and this very function already special-cases `"::"`, so IPv6 is expected input — but a user who types the natural explicit-loopback value `::1`, or any link-local/global v6 address, gets `URL(string: "http://::1:11234/health")`, which is malformed and returns nil, trapping on `!`. Every caller is affected (health polling, `/v1/models`, chat streaming, TTS), and the pre-change code could not hit it because the host was hardcoded. Bracketing a host that contains `:` before interpolating fixes it; the six new tests cover `0.0.0.0`, `::` and an IPv4 LAN address, but no v6 literal that survives the wide-bind check.

t2tx:
Fixed in 34744f4: hosts containing a colon are now bracketed before interpolation, and testIPv6HostIsBracketed pins http://[::1]:11234/health. swift test 3312/0 on M5 Max / macOS 26.5.1.

ddalcu:

- `app/Sources/MLXServe/Services/APIClient.swift:107` — The bracketing fixes the bare-IPv6 case but the force-unwrap still trusts a free-text Settings field to form a valid URL, and the new unconditional bracketing makes an already-bracketed host malformed where it used to work.
  Failure scenario: the host field is free text (`ServerOptions.swift:582`), so a user who types `[::1]` — the form every URL and most docs show — now gets `effectiveHost.contains(":")` true and a second pair of brackets, `http://[[::1]]:11234/health`, which is not a valid IP-literal, so `URL(string:)` returns nil and `!` traps; at the previous head that exact input produced a working URL. The same trap covers every other non-authority string a free-text field invites (a pasted `http://192.168.1.10`, or any value with a space), and health polling calls this on a 1 s timer whether or not the server bound, so the app crashes rather than showing the server as down. Returning `URL?` (or falling back to loopback when the string does not parse) removes the whole class; the new test covers `::1` but no already-bracketed or otherwise unparseable host.

t2tx:
Fixed in 9db5ba7 at the root rather than per-case: serverURL now normalises a pre-bracketed host (so [::1] no longer double-brackets) and falls back to loopback whenever the free-text host does not form a valid URL, so the 1 s health poll reports the server down instead of trapping. New tests pin both: testAlreadyBracketedIPv6IsNotDoubleBracketed and testUnparseableHostFallsBackToLoopback. swift test 3314/0 on M5 Max / macOS 26.5.1.

ddalcu:
🤖 **MLXServe-PRBot approved** — reviewed at head `9db5ba7`, no issues found.

PR #392, opened 2026-09-09
perf(qwen4): batched slots gather their selected QSA blocks; the batched verify carries PLE capture (stacked on #391)

beamivalice:
## Summary

Stacked on #391 (`feature/batch-decode-mtp`), taking its two open follow-ups plus the batching lever from our audit. Five commits (`14fde82`, `3e5b83e`, `cb41abe`, `dff98f3`, `c2807f9`):

- **Gather arm (default on).** Batched qwen4_exp slots each gather their own selected QSA blocks instead of padding one dense `[N,H,S,kv_max]` mask over the group's stacked KV; a slot with a mask but no blocks keeps its mask through the masked-SDPA ladder, plain causal only when a slot has neither. Forced N=1 is byte-identical to the solo arm; N=2 equals the concat of two solos. `MLX_SERVE_QSA_BATCHED_GATHER=0` restores the stacked mask.
- **Deferred PLE on the batched arm.** The batched tick plants the same zero `[N,S,C]` leaf the solo tick does, claims the spec-capture length per slot at build time and fills every row from its own slot's history after the build, removing the host read from inside the layer-1 graph.
- **Batched verify on Flash Next behind `MLX_SERVE_MTP_BATCHED_QWEN4=1`** (off until measured): `pleGatherBf16` is row-correct at `[N,S]`, `qsaMaskBatched` passes the real S so the indexer history advances per slot, `splitSsmToSlots` carries `spec_ple_input` so a partial accept rolls back per slot.
- **Billing.** The pad-waste cap bills the SELECTED length on qwen4_exp, reading exactly the switches, kv floors and the group's vision flag the arm reads, so a group is never admitted on a sparse bill and served the padded stack.

## Bars

- `tests/test_batched_equivalence.sh` two-stream on qwen4_exp: 4/4 (byte-identical single vs batched at N=1, near-ties acquitted at N=2, kv-quant 8 alive).
- Tick time, two concurrent streams (68.6k and 163k tokens), same binary, `MLX_SERVE_QSA_BATCHED_GATHER=1` vs `0`, counterbalanced 1,0,0,1, `MLX_SERVE_ROUND_COST_PERSIST=0`, fans max (M5 Max):

| arm | 64k stream ms/token | 162k stream ms/token | per-slot tok/s |
|---|---|---|---|
| stacked dense mask (as #391) | 113.6 / 114.9 | 113.1 / 114.2 | 8.8 |
| per-slot gather (this stack) | 28.6 / 29.1 | 27.5 / 27.7 | 35 |

  4.0x per tick; `[qsa-arms] qsa=decode_gather` on both slots with the gather, `qsa=mask` without. Concurrent throughput at short context (default 200-token prompt, N=4): base 1.71x, stack 1.70x (the gather does not engage below 8k, no regression).
- Model-backed deferred-PLE identity test (`QWEN4_TEST_MODEL`, pack loaded): 3/3 pass (logits and per-slot PLE state bit-identical deferred vs eager at S=1 and S=4, incl. a partial-accept rollback).
- Note for the maintainer: `tests/test_mtp_batched.sh` fails at N=3 (streams 0 and 2 differ from solo) on this stack **and on the unmodified `ed3ba65`**: pre-existing.

A base-vs-stack tick A/B with the unmodified `ed3ba65` binary follows as a comment (the OFF arm above still carries the deferred PLE, so 4.0x is a lower bound on the kernel term and compound with admission).

Suite: 2258 pass, 154 skip, 0 fail. Audits: one Opus design pass and three Opus review rounds; every P0/P1 closed.

### DIFF
--- a/src/generate.zig
+++ b/src/generate.zig
@@ -6109,6 +6109,7 @@ pub const Generator = struct {
         // the point at which the entry's n-gram history + `spec_ple_tokens`
         // advance. Must precede BOTH the first eval of anything downstream of
         // the leaf (Phase 4) and `ssmRollbackFromCapture` (Phase 5).
+        errdefer xfm.discardDeferredPle(&self.ctx);
         try xfm.flushDeferredPle(&self.ctx);
         st.verify_logits = verify_logits;
         st.new_hidden = new_hidden;
@@ -9256,6 +9257,7 @@ fn lazyForward(xfm: *Transformer, ctx: *ForwardCtx, lazy_token: mlx.mlx_array) !
         return e;
     };
     xfm.flushDeferredPle(ctx) catch |e| {
+        xfm.discardDeferredPle(ctx);
         _ = mlx.mlx_array_free(logits);
         return e;
     };
--- a/src/model.zig
+++ b/src/model.zig
@@ -924,6 +924,14 @@ pub const ModelConfig = struct {
         return self.isQwen4();
     }
 
+    pub fn batchedEffectiveKvLen(self: *const ModelConfig, kv: u32, gather_on: bool, gather_min_kv: u32) u32 {
+        if (!self.isQwen4() or !gather_on) return kv;
+        if (kv <= gather_min_kv) return kv;
+        const cap = self.indexer_budget + self.indexer_compress_ratio;
+        if (cap == 0) return kv;
+        return @min(kv, cap);
+    }
+
     /// SSD-first prefix cache arch predicate; delegates to `longCtxGated`.
     pub fn ssdFirstCapable(self: *const ModelConfig) bool {
         return self.longCtxGated();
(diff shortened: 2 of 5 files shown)


ddalcu:
🤖 **MLXServe-PRBot approved** — reviewed at head `c2807f9`, no issues found.

beamivalice:
Base-vs-stack control, as promised: the unmodified `ed3ba65` binary vs this stack's final binary (`c2807f9`), interleaved base/stack/base/stack, two concurrent streams of 68.6k and 163k tokens, `--max-concurrent 4 --prefix-cache-entries 0 --kv-quant 8`, `MLX_SERVE_ROUND_COST_PERSIST=0`, fans max, M5 Max.

| arm | 64k stream ms/token | 162k stream ms/token |
|---|---|---|
| base `ed3ba65` (stacked dense mask) | 114.7 | 114.1 |
| stack `c2807f9` (per-slot gather) | 30.3 / 29.0 | 27.5 / 27.7 |

Both batched (`[batched] ... engaged (slots=2)` on every boot, no serial-reason line). So the per-tick cost of a long batched pair drops ~3.9x against the base itself, not only against the stack's own kill switch.

One note for you: on the first base boot the pair did not complete. The 163k request was admitted (`[admission] ... verdict=admit`, `prompt=163105 tokens`), never produced a completion line, and the client saw the connection closed with nothing logged; the 64k stream ended after 12 tokens with `[length]`. The second base boot ran the same pair to completion at the numbers above, so it is intermittent on `ed3ba65`; the stack completed all four of its boots. Log: `~/claude-tmp/qsa-live/tick391b/server-base-023858.log` on my box if you want the excerpt.

PR #394, opened 2026-09-09
fix(qwen4): inherited prefix-cache checkpoints restore their QSA history (#390); NAX probes verify correctness; round-cost persistence opt-in

### LINKED ISSUE #390: `QsaHistoryGap`: a conversation that inherited checkpoints from a shared prefix is permanently broken when the donor entry is LRU-evicted
**Model:** `ddalcu/Qwen3.8-Flash-Next-MLX-Serve-mixed-4-8bit` (`qwen4_exp`)
**Version:** mlx-serve 26.9.2 (MLX 0.32.2)

## Summary

Several agent conversations share one large system prompt + tool-schema prefix. The hot prefix cache lets one conversation **inherit checkpoints from a shared prefix** owned by another entry, but the QSA indexer history — which by design exists as **one copy per entry**, pinned to the newest checkpoint — lives on the donor. When the LRU byte budget evicts the donor (`evicted LRU entry (byte budget; …)`), the inheriting conversation is poisoned: the cache keeps reporting a hit on it, and every subsequent request fails `QsaHistoryGap` → `error.GenerationFailed`. It did not self-heal over three consecutive requests and only cleared when the model was unloaded.

Prompt length at failure: **15,771 tokens**. Nothing about this required a long context.

```
[scheduler] prefill failed for slot: QsaHistoryGap
  -> streaming error: error.GenerationFailed
```

## Environment

| | |
|---|---|
| Hardware | MacBook Pro `Mac16,9`, Apple M4 Max, 128 GB |
| macOS | 26.6.2 (25G83) |
| mlx-serve | 26.9.2 release, MLX 0.32.2 |
| wired limit | `[wired] mode=max limit=110100 MB` |
| model | `qwen4_exp` (48 layers, 2560-dim, head_dim 256, 24q/2kv, 4-bit affine) |
| per-model settings | `ctx=524288 kv=default mtp=default` — a manual context raised in Model Settings, **not** RoPE scaling; see "Why the raised context ceiling is not the trigger either" below. |
| prefix cache | hot cache live, **16 entries / 8,192 MB** (as reported by the server's own `resident=… / 8192.00 MB (n/16 entries)` lines); SSD tier **off** (no `~/.mlx-serve/kv-cache`) |
| kv-quant | off |
| MTP | enabled, depth 6 |
| workload | agent client: `sys=17938b` + `tools=46973b` ≈ 15.6k shared prefix tokens, several concurrent conversations, turns of ~10 tokens each |

## Observed symptom

One specific conversation returned a streaming error on every turn. Other conversations in the same boot were fine — **83 requests completed between the `ctx=524288` setting being applied (line 95002) and the first failure**. Client-side there was no remedy: retrying the identical request reproduced it, and a fresh conversation on the same prefix worked. Unloading/reloading the model cured it.

## Log evidence

`~/.mlx-serve/logs/mlx-serve-11234.log`, line numbers are from that file.

**A. A ~21k-token conversation on the same shared prefix, healthy** (note `MTP head restored … from base 15662` — the prefix boundary is 15,662 tokens):

```
96284:   [hot-cache] reused 20774/20897 tokens (matched 20875; entry 5/5)
96285:   [qwen4] MTP head restored (5112 tokens from base 15662)
96288:   [qsa-arms] qsa=verify_gather calls=168 mask=0 decode=0 verify=144 prefill=24
96289:   [hot-cache] resident=7147.29 / 8192.00 MB (5/16 entries)
96290:   <- 20897+70 tokens streamed [prefill: 285.7 tok/s (20774 cached / 20897 total), decode: 101.2 tok/s] [tool_calls]
```

**B. A second conversation restores, is declined the MTP head, and inherits checkpoints** (this turn still succeeded):

```
96297:   [hot-cache] reused 15662/15674 tokens (matched 15666; entry 3/5)
96298:   [qwen4] MTP head restore declined (MtpHeadQsaRingGap) — head starts blind
96302:   [hot-cache] inherited 5 checkpoints (<= 15666 tokens) from a shared prefix
96303:   [hot-cache] resident=7920.73 / 8192.00 MB (6/16 entries)
96304:   <- 15674+90 tokens streamed [prefill: 17.3 tok/s (15662 cached / 15674 total), decode: 58.1 tok/s] [stop]
```

**C. The inheriting conversation works twice more, then the donor is evicted — then it never works again:**

```
96311:   [hot-cache] reused 15673/15779 tokens (matched 15764; entry 6/6)
96312:   [qwen4] MTP head restored (11 tokens from base 15662)
96321:   <- 15779+108 tokens streamed [prefill: 80.2 tok/s (15673 cached / 15779 total), decode: 63.3 tok/s] [stop]
96328:   [hot-cache] reused 15673/15674 tokens (matched 15674; entry 6/6)
96329:   [qwen4] MTP head restored (11 tokens from base 15662)
96333:   [hot-cache] inherited 5 checkpoints (<= 15681 tokens) from a shared prefix
96334:   [hot-cache] evicted LRU entry (byte budget; key=3b32b7fb534b0cc6; was 18529 tokens, 928.24 MB; ssm 407.10 MB)
96335:   [hot-cache] resident=7810.80 / 8192.00 MB (6/16 entries)
96336:   <- 15674+84 tokens streamed [prefill: 1.0 tok/s (15673 cached / 15674 total), decode: 53.2 tok/s] [stop]

96343:   [hot-cache] reused 15673/15771 tokens (matched 15760; entry 6/6)
96344:   [qwen4] MTP head restored (11 tokens from base 15662)
96345: [scheduler] prefill failed for slot: QsaHistoryGap
96346:   -> streaming error: error.GenerationFailed
96353:   [hot-cache] reused 15673/15771 tokens (matched 15760; entry 6/6)
96354:   [qwen4] MTP head restored (11 tokens from base 15662)
96355: [scheduler] prefill failed for slot: QsaHistoryGap
96363:   [hot-cache] reused 15673/15777 tokens (matched 15760; entry 6/6)
96364:   [qwen4] MTP head restored (11 tokens from base 15662)
96365: [scheduler] prefill failed for slot: QsaHistoryGap
96367: [registry] unloading model id=ddalcu/Qwen3.8-Flash-Next-MLX-Serve-mixed-4-8bit (70.13 GB resident)
```

The last successful turn and the first failed turn are **the same request shape, the same entry (`6/6`), the same restore position (15,673), 11 new tokens** — with the donor eviction (96334) as the only intervening event.

Note also that the cache still reports a *hit* on every failed turn (`reused 15673/15771`, `MTP head restored`). The entry is not detected as broken; it is served, and dies downstream.

## Where the error is raised, and why I think it is not the restore guard

`error.QsaHistoryGap` has several raise sites. The ordering in the log identifies this one:

* `prefix_cache.zig:1195` — `checkpointQsaCoversPos(cp, src_cp, cp.pos)` fails during `lookupAndRestore`. This runs **before** the `[hot-cache] reused …` line (`prefix_cache.zig:1295`).
* `transformer.zig:16376` — inside `qsaMaskFromQk`: `if (qsaHistoryRows(entry) != offset) return error.QsaHistoryGap;`, i.e. the logical indexer-history row count no longer equals the KV position at the start of the forward.

In all three failures the log prints `reused …` **and** `MTP head restored …` before failing, so the restore-side coverage predicate passed and the failure is the **downstream one at `transformer.zig:16376`** — the entry looked restorable and then could not account for its own history rows during the prefill.

For contrast, the SSD tier treats the same condition as a *miss*: `prefix_cache.zig:1104` logs `hybrid restore carries no QSA history — falling back to RAM/cold path` and the request proceeds. The hot path has no equivalent recovery.

## Hypothesis

`e0a4264` ("the QSA raw indexer keys are a 32-row ring; the pooled bank is the history") made the **pooled bank the history**, kept **one copy per entry** (`ce5b072`, and `inherited_qsa` in `nonChunkBytes`, `kv_disk_cache.zig:3288`), and added `checkpointHasQsaPooled` to answer "where does the bank live" at "every site that reads, trims, sheds or hands it off". `a93e2c0` added the inheritance this log exercises: "a commit from a restored prefix inherits the donor's checkpoints".

So: entry `6/6` holds checkpoint metadata inherited from a donor, while the pooled bank that those checkpoints' positions are measured against is owned by the donor entry. `evictOneLruProgress("byte budget")` (`prefix_cache.zig:1817/1823/1836`) then releases the donor — including the one copy of the bank. After that:

* the inheritor's logical row count (`pooled blocks × ratio + ring rows`, i.e. `qsaHistoryRows`) no longer equals its cache position,
* `qsaMaskFromQk` refuses, and
* nothing invalidates or repairs the entry, so the prefix cache keeps hitting it forever.

Questions I'd like the maintainers to sanity-check:

1. When an entry inherits checkpoints from a shared prefix, does it take (or share) the pooled bank, or does it depend on the donor staying resident? If it shares, is there a refcount or a handoff on eviction, and does the inheritor's `qsa_pooled` get nulled/shortened by the donor's free?
2. Is `checkpointHasQsaPooled` consulted on the **evict** path, or only on read/trim/restore? The commit message enumerates "reads, trims, sheds or hands it off" — eviction of a *donor with inheritors* isn't obviously in that list.
3. Should a `QsaHistoryGap` surfaced from `qsaMaskFromQk` be downgraded to a cache miss — drop the entry, `resetSsmEntries`, and re-prefill from token 0 — as the disk tier already does? The difference is "this turn was slow" vs "this conversation is dead until the model is reloaded". Even a one-shot self-heal would have hidden this entirely.
4. Is the sibling symptom the same root cause? Same boot, twice: `[qwen4] MTP head restore declined (MtpHeadQsaRingGap) — head starts blind` (96298, 95525). Degrades rather than fails, but MTP acceptance visibly collapses on those turns (`avg_per_round=1.20` / `1.39`, versus `5.00` at 96287 on the healthy conversation) — which is the "head starts blind" cost being paid silently.

## Repro

Not minimized to a hermetic test yet. Natural conditions:

1. App defaults on a 128 GB machine: hot prefix cache 16 entries / 8 GB, SSD tier off. Load `qwen4_exp`, MTP on.
2. Open **two or more agent conversations that share one large prefix** (here: an ~18 KB system prompt plus ~47 KB of tool schemas ≈ 15.6k tokens) — this is the ordinary shape of two Claude-Code-style sessions on the same client.
3. Drive them until the log prints `inherited N checkpoints (… ) from a shared prefix`, and keep going until the byte budget forces `evicted LRU entry (byte budget; …)` for the entry that owns the shared history. `resident=7977.08 / 8192.00 MB` — the budget was genuinely pinned, not slack.
4. Send another turn on the conversation that inherited. Every turn after that fails.

The existing hermetic bars already build "restore as the prefix cache restores at a hit, then append" (per `e0a4264`'s notes). The missing bar looks like: **inherit checkpoints from a donor, evict the donor, then append on the inheritor** — assert the inheritor either self-heals to a cold prefill or stays byte-identical, and does not error.

## Why the raised context ceiling is not the trigger either

That is the obvious suspect — I raised the per-model context to 524,288 minutes before this — so "not an input" would be too strong a claim. `getEffectiveContextLength` reaches three sizing paths, and one of them (the prefill chunk, which sets the SSM checkpoint stride) could in principle change *which checkpoints exist*. Checked each against the log's load map. The setting took effect at the load beginning line 95002; the failing window is 96345-96365, inside that same load.

| path the context reaches | code | under 524288 | under 262144 | verdict |
|---|---|---|---|---|
| hot-cache byte budget | `prefixCacheMemForLoad` → `ctx_kv = (kv/tok + statePerTokenBilled) × ctx` | not re-resolved for this load (no `Hot prefix cache:` line after 95002); every `resident=…` line in the window reads `/ 8192.00 MB` | 8,192 MB (94942) | identical |
| prefill chunk → checkpoint stride | `sizerCtxKvBytes`, `pinPrefillChunk`, `effectiveSsmCheckpointStride` | `Prefill chunk: 4096` (95127) | `Prefill chunk: 4096` (94940 before, 96669 after) | identical |
| per-request admission reserve | `reservedCacheTokens(seq, max_tokens, chunk, ctx)` | bounded by `prompt + max_tokens` = 15,674 + 32,000 = **47,674 rows**, far under both ceilings | same | identical |

And the check that actually fired compares indexer-history row count to the cache position (`transformer.zig:16376`) — it reads neither the context nor the rope spectrum. The failing turns were 15,771 tokens in, well inside the trained window, so no extended position or unscaled angle was involved in them.

**Residual uncertainty, stated plainly:** this is one boot. In this log 6,820 request lines report `ctx=262144` and 87 report `ctx=524288`; all three failures are in the 524288 set. I found no mechanism, which is not the same as proving none exists — so two cheap ways to settle it:

* If it recurs with the context back at 262,144 (which it can, whenever the hot cache reaches its budget with several long-prefix conversations live), the ceiling is exonerated outright.
* If it never recurs under the same workload, start with the three paths in that table rather than with the prefix cache — the prefill chunk is the one that changes checkpoint geometry.

Separately, and this part *is* settled: **YaRN was never configured on this box.** The server logs `[rope] YaRN: factor … (window A -> B), rotary N dims, ramp …, mscale …` at model load whenever the checkpoint declares `rope_parameters` (`transformer.zig:11221`), and there is **no `[rope]` line anywhere in the 6.8 MB log** — all 20 "yarn" matches are conversation text. The raised number moved the advertised ceiling on a model with an unscaled 262,144-token RoPE spectrum; the rope scaling I blamed for my own error was never running. That is a separate finding, below.

Worth noting for context anyway: `statePerTokenBilled` = `qsaHistoryBytesPerToken` (3,840 B) + `qsaScoreBankBytesPerToken` (1,536 B) is ≈5.4 KB/token of state that `--kv-quant` does not reduce, so on this arch a few long shared prefixes eat the hot budget faster than the KV term alone suggests. That's what held the box at 7,977/8,192 MB and made the eviction fire.

## Secondary finding: a per-model context above the checkpoint's RoPE window is applied silently

Unrelated to the crash, but it is why I misdiagnosed it, and I suspect I won't be the last.

`getEffectiveContextLength` (`server.zig:2783`) returns `manualContext(config)` first — `model-settings.json`'s `ctx_size`, else `--ctx-size` — **with no comparison against `contextCap()`**. The only startup comparison lives in the *auto* branch (`server.zig:1560-1569`), which is exactly the branch a manual setting skips; and a per-model setting picked up by an already-running model logs no context line at all. So this box ended up serving an effective context of 524,288 on a checkpoint whose `contextCap()` is 262,144, with no rope scaling, no clamp and no warning: beyond 262,144 tokens the model attends with unscaled positions and quietly gets worse, while the client — and the Model Settings pane — believe it has half a million tokens.

Suggestions, in rough order of value:

1. On the manual path, check against `contextCap()` and say so, e.g. `Context size: 524288 tokens (manual — above this checkpoint's 262144-token RoPE window; long-range quality degrades unless rope_parameters (rope_type: yarn, factor 2) is set)`. Whether to also clamp is your call; the warning alone would have stopped me.
2. Log the same line when a running model adopts a new `ctx_size`, so the live-pickup path isn't silent.
3. In Model Settings, print the checkpoint's own window next to the context field. There's currently no UI at all for rope scaling on an arch that supports it (`8058076`), so the natural reading of "set 512k on a 262k model in an app that offers 512k" is "I now have 512k".

## Two smaller things from the same window (may be unrelated)

1. **96336:** `prefill: 1.0 tok/s (15673 cached / 15674 total)` — a turn that re-prefilled 1 uncached token reported 1.0 tok/s, versus 80.2 tok/s for the same shape two turns earlier (96321). Timing or accounting oddity in the turn where the donor eviction happened.
2. **96288 vs 96301/96319/96332:** the 20.9k-token conversation used `qsa=verify_gather (verify=144)` while the 15.7k ones fell back to `qsa=mask (mask=432-504, verify=0)`. Consistent with `QSA_VERIFY_GATHER_MIN_KV = 16384`, so probably working as designed — flagging only because the decode rates differ visibly across the threshold.

## Workaround in use

Reverting the per-model context to the default and reloading made it go away, which was really the **unload** clearing the poisoned entry rather than the context change. Until this is fixed, the reliable cures are unloading/reloading the model, or keeping the hot cache from reaching its byte budget while several long-prefix conversations are live (`--prefix-cache-mem` well above the working set; `--prefix-cache-entries 0` disables the cache and checkpoint capture entirely).

## Relevant history

The two changes that create the dependency are both from **26.9.2**, which shipped hours before this failure — `a93e2c0` lets one conversation inherit another entry's checkpoints, `e0a4264` moved the history into a bank that one entry owns. The one-copy-per-entry design they interact with is from 26.9.1.

| commit | released | why it matters here |
|---|---|---|
| `e0a4264` | 26.9.2 | Changed what the QSA history *is* — raw indexer keys cut to a 32-row ring, the pooled block bank became the history, on-disk format v8, `checkpointQsaCoversPos` added. Every question about "where does the bank live" dates from this change. |
| `a93e2c0` | 26.9.2 | Introduced the inheritance this log exercises: "a commit from a restored prefix inherits the donor's checkpoints." This is the mechanism that lets one conversation depend on another entry's state. |
| `ce5b072` | 26.9.1 | Introduced the one-copy-per-entry design at the centre of this report — stride captures keep GDN/PLE only, one copy pinned to the latest snapshot, and "no history after restore is a miss on both tiers". Its review notes describe three holes in the same attach / trim / restore area; this looks like a fourth, approached from the eviction side. |
| `f5e7ef8` | 26.9.2 | Hot-cache budget now follows actual residency rather than being fixed at startup. Relevant because the budget is what forced the eviction. |
| `539a39e` | 26.9.2 | "A batch of unrelated requests no longer evicts your live conversation." Nearest neighbour to this report — eviction striking a conversation still in use — so possibly the same ownership gap seen from the other direction. |

For the avoidance of doubt about versions on this box: the failing boot is the **26.9.2 release**. Two earlier boots in the same log were `26.9.2-dev` builds from main, and 26.9.1 precedes them, so the machine has run all three — but with the SSD tier off there are no on-disk entries carried across the v7→v8 boundary.

## Attachment

Wider verbatim excerpt (boot header of the failing session, the per-model settings line, and log lines 96282-96372 with original line numbers): `evidence.txt`.

<details>
<summary><strong>evidence.txt</strong> — verbatim log excerpt (boot header, load map, lines 96282-96372, context-size lines)</summary>

```
### boot header of the failing session (release 26.9.2)
94790: mlx-serve 26.9.2 (MLX 0.32.2)
94791: [args] model: /Volumes/External/AI/Models/ddalcu/Qwen3.8-Flash-Next-MLX-Serve-mixed-4-8bit
94792: [args] drafter: <none>
94793: [args] serve: 127.0.0.1:11234, ctx-size=0, pld=on, no-vision=false, prevent-sleep=true
94794: [args] kv-quant: off
94795: [args] kv-attn-mode: auto
94796: Metal GPU: true
94797: Model: qwen4_exp (48 layers, 2560-dim, head_dim=256, 24h/2kv, 4-bit affine quant)
94798: Loading tokenizer...
94799:   parsed tokenizer (12 MB) in 43ms
94800:   loaded 248044 vocab entries in 4ms
94801:   loaded 247587 merges in 7ms
94802: Tokenizer loaded: 248077 vocab, 247587 merges, 33 special tokens (byte-level BPE)
94803: EOS token from tokenizer: <|im_end|> (id=248046)
94804: User turn marker: "<|im_start|>user
94805: " -> 3 tokens
94806: [registry] max_resident_models=3, max_resident_mem=86.0 GB
94807: [preflight] weights ~70.13 GB, available 116.47 GB
94808: Loading model-00070.safetensors...
94809: Loading model-00042.safetensors...
94810: Loading model-00032.safetensors...
94811: Loading model-00091.safetensors...
94812: Loading model-00084.safetensors...

### per-model settings line
95002: [model-settings] /Volumes/External/AI/Models/ddalcu/Qwen3.8-Flash-Next-MLX-Serve-mixed-4-8bit: ctx=524288 kv=default mtp=default

### failure sequence
96282:   spec-gate: ngram-score=0.177 (threshold=0.010)
96283:   mtp=enabled (streaming, depth=6)
96284:   [hot-cache] reused 20774/20897 tokens (matched 20875; entry 5/5)
96285:   [qwen4] MTP head restored (5112 tokens from base 15662)
96286: [mtp] regime gate: two-chunk 10.56 ms/tok vs single 13.02 ms/tok -> two-chunk every round (period 47)
96287:   [spec-stats] mode=mtp attempts=12 accepts=60 avg_per_round=5.00 per_draft_pct=100.0% depth=6 drafted=60 ext_rounds=9 partial_rounds=0 runtime_disabled=false reason=none adaptive=undecided serial_cell=0.00 sync_ms=3.90 round_ms=68.06 two_ms_tok=10.40 one_ms_tok=13.00 verdict_round=15 trials=0 width_trials=0 table=16-32k:w1:29.07/558,w2:12.93/1558,w3:19.99/521,w4:14.40/807,w5:27.03/355,w6:20.73/617 table_drops=t483/c0/b0/i0 serial_drops=t12/c0/b0
96288:   [qsa-arms] qsa=verify_gather calls=168 mask=0 decode=0 verify=144 prefill=24
96289:   [hot-cache] resident=7147.29 / 8192.00 MB (5/16 entries)
96290:   <- 20897+70 tokens streamed [prefill: 285.7 tok/s (20774 cached / 20897 total), decode: 101.2 tok/s] [tool_calls]
96291:   pld=enabled (draft_len=5, key_len=3)
96292: POST /v1/chat/completions (2 msgs, max_tokens=32000, temp=1.00, top_p=0.95, top_k=20, stream=true, thinking=true, sys=17938b, user=5b, tools=46973b, tool_msgs=0) 
96293:   > "Hello"
96294:   prompt=15674 tokens, max_gen=32000, ctx=524288
96295:   spec-gate: ngram-score=0.156 (threshold=0.010)
96296:   mtp=enabled (streaming, depth=6)
96297:   [hot-cache] reused 15662/15674 tokens (matched 15666; entry 3/5)
96298:   [qwen4] MTP head restore declined (MtpHeadQsaRingGap) — head starts blind
96299: [mtp] regime gate: two-chunk 28.47 ms/tok vs single 12.13 ms/tok -> two-chunk throttled (period 128)
96300:   [spec-stats] mode=mtp attempts=41 accepts=49 avg_per_round=1.20 per_draft_pct=54.4% depth=6 drafted=90 ext_rounds=2 partial_rounds=25 runtime_disabled=false reason=none adaptive=undecided serial_cell=0.00 sync_ms=3.22 round_ms=36.81 two_ms_tok=28.47 one_ms_tok=17.46 verdict_round=15 trials=0 width_trials=0 table=8-16k:w1:19.52/133,w2:17.47/449,w3:29.35/175,w4:21.87/42,w5:21.88/30,w6:23.04/8 table_drops=t485/c0/b0/i0 serial_drops=t12/c0/b0
96301:   [qsa-arms] qsa=mask calls=516 mask=504 decode=12 verify=0 prefill=0
96302:   [hot-cache] inherited 5 checkpoints (<= 15666 tokens) from a shared prefix
96303:   [hot-cache] resident=7920.73 / 8192.00 MB (6/16 entries)
96304:   <- 15674+90 tokens streamed [prefill: 17.3 tok/s (15662 cached / 15674 total), decode: 58.1 tok/s] [stop]
96305:   pld=enabled (draft_len=5, key_len=3)
96306: POST /v1/chat/completions (4 msgs, max_tokens=32000, temp=1.00, top_p=0.95, top_k=20, stream=true, thinking=true, sys=17938b, user=9b, tools=46973b, tool_msgs=0) 
96307:   > "Huh?"
96308:   prompt=15779 tokens, max_gen=32000, ctx=524288
96309:   spec-gate: ngram-score=0.155 (threshold=0.010)
96310:   mtp=enabled (streaming, depth=6)
96311:   [hot-cache] reused 15673/15779 tokens (matched 15764; entry 6/6)
96312:   [qwen4] MTP head restored (11 tokens from base 15662)
96313: [mtp] regime gate: two-chunk 9.83 ms/tok vs single 12.11 ms/tok -> two-chunk every round (period 47)
96314: [mtp] regime gate: two-chunk 12.97 ms/tok vs single 12.11 ms/tok -> two-chunk throttled (period 15)
96315: [mtp] regime gate: two-chunk 13.54 ms/tok vs single 13.74 ms/tok -> two-chunk every round (period 8)
96316: [mtp] regime gate: two-chunk 14.37 ms/tok vs single 13.55 ms/tok -> two-chunk throttled (period 13)
96317: [mtp] regime gate: two-chunk 14.37 ms/tok vs single 14.88 ms/tok -> two-chunk every round (period 8)
96318:   [spec-stats] mode=mtp attempts=36 accepts=72 avg_per_round=2.00 per_draft_pct=56.3% depth=6 drafted=128 ext_rounds=17 partial_rounds=20 runtime_disabled=false reason=none adaptive=undecided serial_cell=0.00 sync_ms=3.74 round_ms=45.72 two_ms_tok=13.80 one_ms_tok=14.56 verdict_round=15 trials=1 width_trials=0 table=8-16k:w1:19.52/133,w2:16.29/459,w3:29.35/175,w4:21.87/42,w5:21.88/30,w6:23.04/8 table_drops=t491/c0/b0/i0 serial_drops=t12/c0/b0
96319:   [qsa-arms] qsa=mask calls=456 mask=432 decode=0 verify=0 prefill=24
96320:   [hot-cache] resident=7977.08 / 8192.00 MB (6/16 entries)
96321:   <- 15779+108 tokens streamed [prefill: 80.2 tok/s (15673 cached / 15779 total), decode: 63.3 tok/s] [stop]
96322:   pld=enabled (draft_len=5, key_len=3)
96323: POST /v1/chat/completions (2 msgs, max_tokens=32000, temp=1.00, top_p=0.95, top_k=20, stream=true, thinking=true, sys=17938b, user=5b, tools=46973b, tool_msgs=0) 
96324:   > "Hello"
96325:   prompt=15674 tokens, max_gen=32000, ctx=524288
96326:   spec-gate: ngram-score=0.156 (threshold=0.010)
96327:   mtp=enabled (streaming, depth=6)
96328:   [hot-cache] reused 15673/15674 tokens (matched 15674; entry 6/6)
96329:   [qwen4] MTP head restored (11 tokens from base 15662)
96330: [mtp] regime gate: two-chunk 22.05 ms/tok vs single 37.01 ms/tok -> two-chunk every round (period 128)
96331:   [spec-stats] mode=mtp attempts=36 accepts=50 avg_per_round=1.39 per_draft_pct=44.6% depth=6 drafted=112 ext_rounds=19 partial_rounds=24 runtime_disabled=false reason=none adaptive=undecided serial_cell=0.00 sync_ms=3.65 round_ms=45.55 two_ms_tok=18.99 one_ms_tok=30.80 verdict_round=15 trials=0 width_trials=0 table=8-16k:w1:19.52/133,w2:15.76/460,w3:29.35/175,w4:21.87/42,w5:21.88/30,w6:23.04/8 table_drops=t493/c0/b0/i0 serial_drops=t12/c0/b0
96332:   [qsa-arms] qsa=mask calls=444 mask=432 decode=12 verify=0 prefill=0
96333:   [hot-cache] inherited 5 checkpoints (<= 15681 tokens) from a shared prefix
96334:   [hot-cache] evicted LRU entry (byte budget; key=3b32b7fb534b0cc6; was 18529 tokens, 928.24 MB; ssm 407.10 MB)
96335:   [hot-cache] resident=7810.80 / 8192.00 MB (6/16 entries)
96336:   <- 15674+84 tokens streamed [prefill: 1.0 tok/s (15673 cached / 15674 total), decode: 53.2 tok/s] [stop]
96337:   pld=enabled (draft_len=5, key_len=3)
96338: POST /v1/chat/completions (4 msgs, max_tokens=32000, temp=1.00, top_p=0.95, top_k=20, stream=true, thinking=true, sys=17938b, user=10b, tools=46973b, tool_msgs=0) 
96339:   > "Hello"
96340:   prompt=15771 tokens, max_gen=32000, ctx=524288
96341:   spec-gate: ngram-score=0.155 (threshold=0.010)
96342:   mtp=enabled (streaming, depth=6)
96343:   [hot-cache] reused 15673/15771 tokens (matched 15760; entry 6/6)
96344:   [qwen4] MTP head restored (11 tokens from base 15662)
96345: [scheduler] prefill failed for slot: QsaHistoryGap
96346:   -> streaming error: error.GenerationFailed
96347:   pld=enabled (draft_len=5, key_len=3)
96348: POST /v1/chat/completions (4 msgs, max_tokens=32000, temp=1.00, top_p=0.95, top_k=20, stream=true, thinking=true, sys=17938b, user=10b, tools=46973b, tool_msgs=0) 
96349:   > "Hello"
96350:   prompt=15771 tokens, max_gen=32000, ctx=524288
96351:   spec-gate: ngram-score=0.155 (threshold=0.010)
96352:   mtp=enabled (streaming, depth=6)
96353:   [hot-cache] reused 15673/15771 tokens (matched 15760; entry 6/6)
96354:   [qwen4] MTP head restored (11 tokens from base 15662)
96355: [scheduler] prefill failed for slot: QsaHistoryGap
96356:   -> streaming error: error.GenerationFailed
96357:   pld=enabled (draft_len=5, key_len=3)
96358: POST /v1/chat/completions (5 msgs, max_tokens=32000, temp=1.00, top_p=0.95, top_k=20, stream=true, thinking=true, sys=17938b, user=15b, tools=46973b, tool_msgs=0) 
96359:   > "Hello"
96360:   prompt=15777 tokens, max_gen=32000, ctx=524288
96361:   spec-gate: ngram-score=0.155 (threshold=0.010)
96362:   mtp=enabled (streaming, depth=6)
96363:   [hot-cache] reused 15673/15777 tokens (matched 15760; entry 6/6)
96364:   [qwen4] MTP head restored (11 tokens from base 15662)
96365: [scheduler] prefill failed for slot: QsaHistoryGap
96366:   -> streaming error: error.GenerationFailed
96367: [registry] unloading model id=ddalcu/Qwen3.8-Flash-Next-MLX-Serve-mixed-4-8bit (70.13 GB resident)
96368: [wired] mode=max limit=110100 MB
96369:   parsed tokenizer (12 MB) in 39ms
96370:   loaded 248044 vocab entries in 5ms
96371:   loaded 247587 merges in 7ms
96372: Tokenizer loaded: 248077 vocab, 247587 merges, 33 special tokens (byte-level BPE)

### context-size lines around the 524k setting (same boot)
94942: Hot prefix cache: ENABLED (capacity=16, mem-cap=8192.0 MB)
94943: Concurrency: --max-concurrent=2 (continuous batching enabled)
94944: Context size: 262144 tokens (auto: the model's maximum; memory would allow 582656) [pinned]
94945: Request timeout: 300s
94946: Model context length: 262144 tokens
94947: Reasoning budget: unlimited
94948: PLD speculative decoding: ENABLED (draft_len=5, key_len=3; default for new requests)

### load map: context / chunk / cache-cap lines, last loads (line numbers as in the log)
94944: Context size: 262144 tokens (auto: the model's maximum; memory would allow 582656) [pinned]
94946: Model context length: 262144 tokens
95000: [registry] unloading model id=ddalcu/Qwen3.8-Flash-Next-MLX-Serve-mixed-4-8bit (70.13 GB resident)
95002: [model-settings] /Volumes/External/AI/Models/ddalcu/Qwen3.8-Flash-Next-MLX-Serve-mixed-4-8bit: ctx=524288 kv=default mtp=default
95127: Prefill chunk: 4096 tokens (memory-sized down from 8192; --prefill-chunk overrides)
96367: [registry] unloading model id=ddalcu/Qwen3.8-Flash-Next-MLX-Serve-mixed-4-8bit (70.13 GB resident)
96493: Prefill chunk: 4096 tokens (memory-sized down from 8192; --prefill-chunk overrides)
96518: mlx-serve 26.9.2 (MLX 0.32.2)
96669: Prefill chunk: 4096 tokens (memory-sized down from 8192; --prefill-chunk overrides)
96671: Hot prefix cache: ENABLED (capacity=16, mem-cap=8192.0 MB)
96673: Context size: 262144 tokens (auto: the model's maximum; memory would allow 582656) [pinned]
96675: Model context length: 262144 tokens

```

</details>

beamivalice:
## Summary

Fixes #390.

Follow-up to #388's post-merge review, plus the fix for #390 (reported by @d-b). Six themes, one commit each:

- **NAX probes verify correctness, not just compilation.** `qsaScoreFusedRunProbe` and `qsaNaxRunProbe` run a deterministic pseudo-random input through the kernel and the composed chain (bit equality for the score sheet, the bf16-store bar for the gather) and latch the stock arm on a mismatch, so a driver change that shifts the cooperative-fragment lane map can no longer mis-score silently. A test fixture no longer leaks `qsa_score_fused_override`; release builds pass `-Dgit-sha`.
- **Round-cost persistence is opt-in.** `MLX_SERVE_ROUND_COST_PERSIST=1` saves the measured table (keyed on the engine build); by default every boot re-measures. The serial-drift cell wipe is deleted: boot-to-boot drift is 5–12%, so a 5% trigger erased cells on most reboots.
- **#390: the QSA indexer bank follows checkpoint inheritance.** A conversation that inherited checkpoints from a shared prefix could be left with checkpoints and no pooled bank; the lookup accepted the shape and the forward raised `QsaHistoryGap` on every later turn until the model was unloaded. Now the donor's bank is sliced onto the newest cloned checkpoint, "a committed checkpoint list carries a bank or lands KV-only" is a checked invariant at every list-producing site, and lookup and forward share one coverage predicate asked at the position that will actually be restored.
- **Restore walks down to the highest covering checkpoint** instead of declining a backed heir; the stale-position commit capture is gone.
- **The self-heal retry is cold**: a forward-side `QsaHistoryGap` drops the restored entry (poisons an SSD entry), resets SSM entries, the MTP head and the pooled-rope cache, skips the cache once, and retries. Inherit/merge drops unbill their SSM bytes; the shed rescues a bank only downward.
- **The QSA leftover at a restore position is `pos % ratio` rows of the source ring by absolute position**, zero at an aligned position, never a 32-row overlay; the pooled truncation is materialised; the full-match clamp to `prompt_len-1` applies only when the covering checkpoint sits at the prompt end.
- **Inherited checkpoints stop at `min(donor shared prefix, token divergence, the heir's own prompt length)`**: a greedy continuation regenerating the donor's text made the token walk report a shared prefix past the heir's prompt, cloning donor checkpoints from inside the donor's generated tail onto the heir.
- **`MLX_SERVE_DUMP_RESTORE=<dir>`** (debug, inert when unset) dumps every restored per-layer SSM/QSA/PLE array and the KV tails after a warm restore or a cold prefill; `tests/diff_restore_dump.py` diffs two dumps per tensor. It is how the last round was settled.

## Live

`tests/test_prefix_inherit_qsa.sh` (new) and the coordinator's driver reproduce d-b's log on main (`full reuse N/N+1` → `MTP head restore declined (MtpHeadQsaRingGap)` → `prefill failed for slot: QsaHistoryGap` → 500) and pass on this branch with no gap and no self-heal event; with the dump hook, a heir restore and a donor restore at the same position are bit-identical across all 506 dumped tensors, and both differ from a true cold prefill to that position only by the documented chunking-class noise.

One thing this PR does not fix, found while chasing it and reproduced on main by restoring straight from the donor entry: a warm restore followed by a short tail forward (about 25–29 rows) at a degenerate cut point can answer EOS where a cold prefill answers text. Pre-existing, independent of #390, filed separately.

## Speed (llmprobe `--bench-only --rungs 4k,8k,16k,32k,64k --runs 3`, main vs this branch, fans max, both unpersisted)

| rung | main decode | branch decode | main prefill | branch prefill |
|---|---|---|---|---|
| 4k | 101.3 | 108.6 | 1834 | 1871 |
| 8k | 104.2 | 102.0 | 1625 | 1683 |
| 16k | 98.7 | 100.6 | 1703 | 1752 |
| 32k | 111.5 | 100.4 | 1704 | 1769 |
| 64k | 98.9 | 107.2 | 1722 | 1755 |
| 2k headline | 105.6 | 109.4 | 1751 | 1709 |

Prefill +2..4% at every rung (the NAX probes and the unpersisted table cost nothing), decode inside the ±5–10% boot noise (main `ecf8d35` binary vs this branch, one boot each).

Suite: 2309 pass, 153 skip, 0 fail (executable built ReleaseFast). Audits: four Opus rounds on the #390 fix, one grok and one GPT-5.6 round on the review items; every P0/P1 closed.


ddalcu:
🤖 **MLXServe-PRBot approved** — reviewed at head `60db51d`, no issues found.

PR #396, opened 2026-09-09
serve: prefill_tokens_expected for the monitor plugin; #394 follow-up (self-heal cache, view restore, per-lookup skip, sequential live test)

beamivalice:
## Summary

Two commits: the prefill gauge for the monitor plugin, and the post-merge follow-up of #394 (five review findings plus the live script).

The opencode2 monitor plugin draws a prefill progress bar, and until now it had to guess the denominator. The chunk loop already publishes how far the in-flight prefill has got (`prefill_tokens_live`); how far there is to go had no surface, so the panel inferred it from the last step. This PR gives it the real target and updates the plugin to use it.

- **`prefill_tokens_expected`**: a new gauge beside `prefill_tokens_live`, on the same scale (the post-cache tail this prefill will forward), published once per prefill and cleared on every prefill exit next to the progress atomic. Wired through the Prometheus text and `/metrics.json`. 0 when idle and on the bypass engines, which the plugin reads as "no bar".
- **Plugin** (`lib/opencode2-mlx-serve`, three bumps): the bar targets the server gauge and falls back to the rate shape when it reads 0 (older builds); the footer settles into three fixed shapes; steady Throughput lines that never disappear at 0; the Server cluster (gpu, running, tokens, requests, messages, tool calls), Model & sampling, Server log as a feed state; the Memory gauge on its own bar row with the reading bright and the ceiling dim.

Server side is 34 lines across `generate.zig`, `metrics.zig`, `scheduler.zig`, `server.zig`, with the metrics test extended for the new gauge in both emitters and the at-rest zero.

## #394 follow-up (second commit)

- The QSA self-heal drops the entry the SLOT's model cache restored, not whichever model loaded last: with two models resident the retry used to evict an innocent entry and re-fail forever.
- A warm restore rebinds the entry's pooled QSA bank as a view instead of copying it (~768 MB per restore at 1M tokens); the SSD-tier overlay and the inherit slice still own their rows, and the commit handoff never shares a never-appended view into a new entry.
- `inherit_limit` is `min(donor.shared, prompt_cap)`; the token-divergence walk recomputed what `bestCheckpointDonor` already stored.
- `skip_prefix_cache` is an argument of the lookup instead of a field on the shared cache, so an early-returning arch cannot leave the next request cold.
- `MLX_SERVE_DUMP_RESTORE` is read once per process; the restore-dump locals collapse into one struct; production dump sites 5 → 2.
- `tests/test_prefix_inherit_qsa.sh` runs its warm and cold phases sequentially on one port instead of booting a second 70 GB server beside the first. Live on the M5 Max: the one-token shape occurs, no `QsaHistoryGap`, warm greedy text equals cold.

Suite on the branch: 2309 pass, 153 skip, 0 fail; executable built ReleaseFast; one Opus audit round, every P1 closed.

### DIFF
--- a/src/generate.zig
+++ b/src/generate.zig
@@ -1951,6 +1951,9 @@ pub const Generator = struct {
         /// prefill saturates the GPU while both tiles read 0 / "—". The
         /// scheduler resets it to 0 when the prefill ends.
         prefill_progress: ?*std.atomic.Value(u64) = null,
+        /// Total tokens this prefill will forward (post-cache tail, same scale
+        /// as `prefill_progress`), published once; null on the bypass engines.
+        prefill_expected: ?*std.atomic.Value(u64) = null,
 
         /// Called once per completed prefill chunk boundary (chunk state
         /// evaluated, allocator cache cleared), except the final one. The
@@ -2350,6 +2353,7 @@ pub const Generator = struct {
             );
 
             var pos: usize = 0;
+            if (options.prefill_expected) |e| e.store(@intCast(loop_end), .monotonic);
             while (pos < loop_end) {
                 // Abandoned-request abort: the client disconnected and the
                 // conn thread flagged the slot. Bail before the next chunk —
--- a/src/kv_disk_cache.zig
+++ b/src/kv_disk_cache.zig
@@ -786,7 +786,7 @@ pub const DiskTier = struct {
             }
         }
         if (qsa_overlay) |*qsa_cp| {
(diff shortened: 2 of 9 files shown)


ddalcu:
🤖 **MLXServe-PRBot approved** — reviewed at head `b49e4d8`, no issues found.

ddalcu:
🤖 **MLXServe-PRBot approved** — reviewed at head `40eeb93`, no issues found.

Issue #366, opened 2026-09-06
Prefill: no MoE profiler for S>1, and where the time actually goes on Qwen3.8-Flash-Next (M5 Max, measured)


nikolai-vysotskyi:
> Latest attribution update, 2026-09-13: actual-weight MoE replay retains an exact-output **1.11–1.13x component gain**. Same-process model controls drift downward about10%; GPU telemetry measures **1619→1333 MHz and64.9→97°C**. The earlier cross-boot throughput drop is preserved below, but its entire size cannot be causally attributed to the combined research options. **2400 tok/s at65K remains unproven.** [Code, measurements and discussion](https://github.com/ddalcu/mlx-serve/issues/366#issuecomment-5649448839).

**Current scope (2026-09-13):** investigate further mixed-pack prefill gains toward 2400 tok/s at 65K. The original unsorted-MoE/utilization diagnosis below was disproven: production already sorts, and the early comparison mixed chunk sizes. Current implementation and measurements: [research checkpoint](https://github.com/ddalcu/mlx-serve/issues/366#issuecomment-5647960600); production fusions and review: #408. The original report is retained below as historical context.

---

### Summary

On `ddalcu/Qwen3.8-Flash-Next-MLX-Serve-4bit` (qwen4_exp, 512 experts, top-10, `moe_intermediate_size` 640), prefill reaches **~8 TFLOP/s on an M5 Max whose measured 4-bit GEMM ceiling is ~50 TFLOP/s** — about 16 % utilisation. The dense 27B on the same machine and server reaches 33 TFLOP/s (66 %).

A microbenchmark points at the cause: `mx.gather_qmm` with **`sorted_indices=True`** is up to **19× faster** than the unsorted call at these shapes, and the unsorted path is completely flat in batch size — which matches the behaviour I observe from the server (raising `--prefill-chunk` does not help).

### Measurements

**End to end**, M5 Max 128 GB, macOS 26.5, mlx-serve 26.9.1, `--ctx-size 131072 --prefix-cache-entries 0`, 65 000-token prompt of real source code, median of 3, otherwise-idle machine:

| | 65k prefill |
|---|---|
| Ollama 0.33.3 MLX, Qwen3.8-27B nvfp4 (dense) | 651 tok/s |
| mlx-serve, Qwen3.8-27B-MLX-Serve-4bit (dense) | 604 tok/s |
[...]

PR #397, opened 2026-09-10
fix(gdn): the conv_state tail no longer pins the prefill chunk (-2.5 GB); the admission bills follow the buffers; --wired-margin-gib; --mtp-head-kv-quant; checkpoint default 16

beamivalice:
## Summary

Five memory items for the conv-cached archs (Qwen3.8 Flash Next first), one squashed commit; items 1 to 4 are output byte-identical, item 5 is an opt-in flag for the speculative head's cache precision, measured.

**1. The GDN conv_state tail no longer pins the whole prefill chunk input.** `conv1dWithCache` kept a 3-row *view* of `conv_input`, and the residual's cadence eval materialises `conv_input`, so every linear-attention layer held its full chunk input until its next forward: 36 × 4096 × 10240 × 2 B ≈ 3 GB per 4096-token chunk on qwen4_exp. The tail is now an owned copy, and it is **named in the cadence eval vector inside the layer loop** (`evalCadencePoint`), which is the load-bearing part: a lazy copy is not in the residual's graph, so a plain `mlx_array_eval(h)` leaves it lazy and still holding the parent (that is why an earlier attempt measured null). The copy is gated on a chunk-wide parent (T ≥ 32), so composed decode on lfm2/nemotron/KDA pays nothing. Found by **Nikolai V.** in #366.

Measured on the production mixed 4/8-bit pack, production flags, fans max, one boot per cell (`--prefill-chunk 8192` is clamped to 4096 on this model, so 4096 is the real production width):

| cell | before | after | saved |
|---|---|---|---|
| cold 24k-token prompt, width 4096, peak MLX bytes | 78.50 GB | 76.03 GB | 2.47 GB |
| llmprobe 4k–128k ladder, peak after the 128k rung | 86.53 GB | 83.73 GB | 2.80 GB |

Greedy replies byte-identical on vs off. Speed neutral: llmprobe 4k–128k × 3 runs, prefill +1–3% on every rung with the copy on (1858/1671/1743/1744/1685/1601 vs 1828/1649/1718/1704/1632/1562 tok/s), decode within spec-decode noise both ways.

**2. The prefill admission bill follows it.** `prefillStreamBytesPerToken` charged every linear layer's chunk-wide stream (737 KB/token here) because that term *was* this pin; it now bills `min(linear_layers, cadence + 1)` layers. Live: the `[admission]` line for a 4096 chunk on the same prompt went from `needed=8495 MB` to `needed=5184 MB`, same greedy output. The measured-peak rows for qwen3.5 27B/4B were re-derived as measured peak minus the exact pin bytes (the packs are not on this box to re-measure); the test says so.

**3. The QSA raw-key ring is billed once per slot, not per token.** Since #381 the raw indexer keys are a fixed 32-row ring, but `qsaHistoryBytesPerToken` still charged 3 KB/token of them: 0.8 GB of phantom at 256k and 1.6 GB at 512k that shrank the advertised context and refused prompts that fit. Only the pooled bank (768 B/token) is per token now; `qsaRingBytes` (96 KiB) is added once per live slot at every bill site.

**4. Two knobs.** `--wired-margin-gib <n>` (default 8, the previous fixed value; 2..32, named error otherwise) sets how far under a raised `iogpu.wired_limit_mb` the planner may reach; a 96 GB Mac at a 90–96 GB limit can take 6. `--ssm-checkpoint-max` defaults to 16 (was 32): it halves the checkpoint bytes a cached session holds (~57 MB each on Flash Next) and at 256k still leaves a checkpoint every 16k tokens; the production launcher already passed 16.

**5. `--mtp-head-kv-quant` (default off) lets the MTP head's KV follow `--kv-quant`; the head KV is billed either way.** The qwen4 speculative head kept its own KV cache dense bf16 regardless of `--kv-quant 8`, and never counted it in admission. With the flag it takes the trunk's scheme at load (both construction sites, retained on the cold-load path), every consumer reads through the same `denseView`/fused-kernel path as the trunk, the prefix-cache snapshot and the SSD spec sidecar carry the quantized triple, and a sidecar written under the other scheme is declined at restore and rewritten on the next commit. The head KV is billed per request at its effective width (2048 B/token dense, 1088 at 8-bit). It stays opt-in because short prompts sit under the fused-kernel floor (about 2048 head rows) and dequantize per draft step, a 6% short-prompt cost in the A/B below.

A/B, same binary, production flags with MTP on, llmprobe 4k–128k × 3 runs, boots A/B/B/A (A = 8-bit head via the flag, B = dense, the default):

| rung | decode A1 / B1 / B2 / A2 (tok/s) | A/B pair 1, pair 2 | prefill A/B pair 1, pair 2 |
|---|---|---|---|
| 4k | 95.2 / 109.1 / 108.1 / 103.4 | 0.87, 0.96 | 0.97, 1.02 |
| 8k | 101.4 / 98.8 / 96.8 / 103.2 | 1.03, 1.07 | 0.98, 1.02 |
| 16k | 98.2 / 103.3 / 95.7 / 98.4 | 0.95, 1.03 | 1.00, 1.01 |
| 32k | 110.1 / 106.0 / 89.5 / 94.9 | 1.04, 1.06 | 1.01, 1.01 |
| 64k | 96.8 / 99.0 / 92.5 / 98.8 | 0.98, 1.07 | 1.01, 1.00 |
| 128k | 99.2 / 91.5 / 85.5 / 93.0 | 1.08, 1.09 | 1.02, 1.00 |

Per-draft acceptance from the server's spec-stats is within a few points of dense at every context bucket (e.g. 64–128k: 83.5/87.0 vs 86.4/84.9), inside boot-to-boot scatter; peak MLX bytes −240 MiB at the 128k rung on both pairs. The short-prompt cell leans dense by about 6%: below ~2048 head rows the quantized cache misses the fused kernels and dequantizes per draft step, which is why the quantized head is opt-in. `tests/test_mtp_equivalence.sh` 11/11, `tests/test_qwen4_exp.sh` 40/40. One Opus audit (its P1, old dense sidecars never rewritten, and three bill items are in) and one DeepSeek V4.1 audit of the whole PR: its P1 (the head was billed at a request's `kv_quant` override rather than the scheme it was loaded with) and five P2s (truncate covering every scheme, the head's own QSA state billed, the MoE coexist constant derived from the cadence, one home for the flag, the reverse sidecar-rewrite test) are in.

Full suite on the squashed tree: 2325 pass, 154 skip.

## Tests

- Retention: `conv1dWithCache tail drops the prefill chunk parent` (+ a batch-2 strided case) — red with the copy stashed (`retention delta=6311426 chunk=2097152`, `delta=12604418 chunk=4194304`), green with it.
- Identity: on/off byte-equal at T=1/5/4096, batch 1/2, and across two chunks (`compact_conv_state_override` test seam; no env switch ships, it is an exact lifetime change).
- Cadence: `hybrid prefill: the cadence eval names every conv_state inside the layer loop` — fails with the four cadence sites reverted.
- Bills: `prefillStreamBytesPerToken` expectations, the re-derived measured-peak rows, `a 512k QSA session drops the raw-key overbill and bills the ring once` (ring invariance + a refused-then-admitted 512k prompt against a fixed ceiling), `wiredLimitFloor` at margins 8 and 6 on a 96 GB box, `parseWiredMarginGib`.
- Full suite: 2325 pass, 154 skip (squashed tree). `tests/test_qwen4_exp.sh`: 40/40; `tests/test_mtp_equivalence.sh`: 11/11.

Two Opus audit rounds; the mechanism was clean in both, the second round only re-derived four admission guards that had been loosened and renamed the cadence test to what it asserts.

Journal: `conv-state-compaction.md`; gotcha story in `docs/gotchas/engine-mlx.md`.


ddalcu:
🤖 **MLXServe-PRBot approved** — reviewed at head `89dcf47`, no issues found.

ddalcu:
🤖 **MLXServe-PRBot approved** — reviewed at head `30485f5`, no issues found.

ddalcu:
🤖 **MLXServe-PRBot approved** — reviewed at head `6c253f3`, no issues found.

ddalcu:
🤖 **MLXServe-PRBot approved** — reviewed at head `ca4bf0f`, no issues found.

PR #398, opened 2026-09-10
fix: implement the --idle-evict-secs sweep

### LINKED ISSUE #241: Feature request: unload models after idle timeout
Would it be possible to add an optional idle timeout for loaded models?

For example, after 10 minutes without requests, the model would be unloaded and its MLX/Metal memory released. The next request would load it again.

This would be useful on Macs with limited unified memory, especially when running large models.

For example https://github.com/ggml-org/Llama-macOS supports it via

Image

Thanks!

latent-variable:
## Summary

Implements the `--idle-evict-secs` sweep. The flag has been accepted since multi-model support landed but never did anything — `main.zig` parses it, `ModelRegistry` stores it, and no code ever read it, so the "inference thread's idle tick (Phase D)" its doc comment named never existed.

## Changes

- **`server.zig`** — `idleEvictLoop` at `serve()` scope with its own stop flag, the same shape as `gaugeSamplerLoop`. Not inside `Scheduler`: a thread joined in `deinit` cannot gate on `scheduler.shutdown` (deinit is the earliest-registered defer, runs last), and a sweep blocked in `unloadModel` waits on the inference thread, which `deinit` only rescues *after* joining — so a join there hangs the process on shutdown.
- **`scheduler.zig`** — `unloadModelIfIdle`, which is `unloadModel` with the idle precondition re-checked under the registry mutex at commit. Without it, a request arriving between the pick and the commit finds its model marked `.evicting`, and every other client blocks in `ensureLoaded` until that request finishes, then pays a cold reload. Plus `idleEvictTickMs`, hoisted so the interval is testable.
- **`model_registry.zig`** — `LoadedModel.last_used_ms` on the boot clock, stamped in `release()` and `markReadyLocked()`, and `pickIdleEvictable`. `last_used_ns` is a bare LRU ordering counter and cannot answer "how long has this been idle". Boot rather than wall so an NTP step cannot stall the timer.
- Corrected the two comments in `model_registry.zig` and `main.zig` that described the mechanism that was never built.

The eviction path itself is untouched: mark `.evicting`, drain refcount to 0, hand the mlx free to the inference thread. Nothing in the sweep touches mlx directly.

## Test plan

- [x] `zig build test` — 2315 pass, 153 skip, 0 fail
- [x] `zig build -Doptimize=ReleaseFast`
- [x] End to end: request loads the model (0.58 GB resident), sweep unloads it 24s after the request on a 20s window, log line emitted
- [x] Shutdown stress: server killed mid-sweep at randomised offsets, 6 rounds, a sweep confirmed fired in each. Exits in 2-3s every time, no hangs
- [x] Soak under contention: 25 requests against a 1s window. All succeeded, zero eviction failures, no client blocked
- [x] Controls: removing the refcount guard fails the live-request test by name; moving the join back into `Scheduler.deinit` fails the ordering test by name

## Notes

`--idle-evict-secs 0` still parses to `null` (off), unchanged. All five flag-accepting paths (`runDs4Serve`, `runLlamaServe`, `runHeadlessServe`, `runGenServe`, and the primary) route through one `Scheduler.init`, so the sweep covers every mode.

Fixes #241


ddalcu:
🤖 **MLXServe-PRBot approved** — reviewed at head `1a3275d`, no issues found.

latent-variable:
Disclosure before you spend time on this: **eviction leaks ~196 MB per cycle**, and this PR makes that leak periodic rather than manual. Filed separately as #399 because it is not caused by this change.

I checked it both ways on the same binary and the same model. Driven by this PR's sweep:

```
cycle 1:  load 665 MB  ->  evict 260 MB
cycle 2:  load 862 MB  ->  evict 457 MB
cycle 3:  load 1059 MB ->  evict 654 MB
```

Driven by `POST /v1/unload-model` with `--idle-evict-secs` off entirely:

```
cycle 1:  load 665 MB  ->  unload 260 MB
cycle 2:  load 862 MB  ->  unload 457 MB
cycle 3:  load 1058 MB ->  unload 654 MB
```

Same numbers, so the leak is in `unloadResident`, which this sweep only calls. Linear over 8 cycles with no plateau (table in #399), and `/props` reports `active_bytes` far below RSS, so it is not MLX-side.

It still bears on whether you want to merge this as-is. Right now the leak is reachable only by someone calling unload in a loop; with `--idle-evict-secs` on, a long-running server hits it every idle window. On a 900s window that is roughly 196 MB per cycle, indefinitely.

Three ways to go, your call: merge this and fix #399 separately, hold this until #399 lands, or I fold a fix for #399 into this branch. Happy to do the last one — the likely fix is small, but whether the tokenizer is meant to survive an unload is a design call I would rather you made.

ddalcu:
feel free to add #399 fix here as well.

latent-variable:
Added the #399 fix to this PR as you suggested — second commit.

`unloadResident` retains `config`/`tokenizer`/`chat_config`/`token_bytes`/`tokenize_cache` by contract, so the leak is on the reload: it installs a fresh set over the retained one and orphans the previous generation. `LoadedModel.releaseRetainedCpuState` frees exactly those five in `deinit`'s order, called at all four load paths before installing.

**Freeing it exposed a use-after-free the leak had been hiding, so that is fixed here too.** `/api/tags` and `/api/show` run ahead of model resolution: they hold the registry mutex but no refcount and no state check, and read `config`/`chat_config` for every entry. The release runs on the inference thread without that mutex. Previously the read was benign because the old object was leaked and still valid; freeing makes it a genuine UAF, and `/api/show` copies `chat_template` into the response body. Both handlers now skip retained state when an entry is `.loading` — the only state it can be freed in, since a load starts from `.unloaded` via `tryBeginLoadLocked`, which takes the mutex to set `.loading`. Deliberately `!= .loading` and not `== .ready`, so an unloaded entry stays fully listable.

Two smaller things in the same commit: `ModelConfig.ngram_table_path` is `allocPrint`'d and was not freed by `deinit` either, so both paths free it now; and `Scheduler.config`/`.tok`/`.chat_config` are removed, boot-time borrowed views with no readers whose comment assumed their addresses stayed valid.

Measured: RSS flat at 457-459 MB across cycles 2-12, against 1634 MB and still climbing at cycle 8 before. Survived 12 reloads with 240 concurrent `/api/tags` + `/api/show` racing them. `/api/tags` still reports `family=qwen3_5_moe` after an unload rather than falling back to the arch hint.

Three controls, each verified to fail with the fix removed: the tokenizer free (3 leaked allocations), the `ngram_table_path` free (1 leak, 18 bytes), and the `.loading` guard (behavioural test on `ollamaTagEntryOf` across `.unloaded`/`.ready`/`.loading`).

I dropped two source-scan tests I had written first — CLAUDE.md bans them, and mine managed to hit the exact failure mode it names by matching its own needle. Comments swept for issue refs and measurements per the same section.

CI has not run on this PR; happy to rebase or split the two commits if you would rather review them separately.

ddalcu:
🤖 **MLXServe-PRBot approved** — reviewed at head `7a9c1d8`, no issues found.

ddalcu:
Not merging yet. The sweep itself is fine and the suite passes here on 7a9c1d8. Commit 2 has a hole.

1. You free the retained config on reload and guarded /api/tags and /api/show, but missed the pre-load text-gen gate in `handleConnection`. It does `registry.peek(requested_model_id)` then `textGenTargetOf(peeked)`, which reads `config.is_encoder_only` with no mutex and no refcount (server.zig:2120 on your head). Model gets idle-evicted, request A starts the reload, request B hits that peek while the inference thread is in `releaseRetainedCpuState`. That is a read of freed memory, and with idle eviction on it is the normal reload path, not a corner case. Take the registry mutex around the peek and the snapshot, and skip the config when the entry is `.loading`, same as you did for the Ollama handlers.

2. Drop the `the idle-evict sweep is stopped and joined outside Scheduler.deinit` test in scheduler.zig. It's a source scan and CLAUDE.md bans them. You removed your other two and missed this one.

3. handleOllamaShow has two stacked comments saying the same thing, and the first one calls it a `.ready` gate, which it isn't. Keep one.

4. The app never passes `--idle-evict-secs`, so for app users this feature doesn't exist. Add it to `ServerOptions` the same way `maxResidentMemGB` is wired (field, `toCLIArgs`, decode, equality, `SettingsReset`, explainer) plus a control in the server settings UI, default off. Post a screenshot of the new setting here.

latent-variable:
All four addressed.

- Text-gen gate: entry and snapshot now happen under one registry-mutex hold (`peekLocked`), and `textGenTargetOf` skips the config while the entry is `.loading`. Same predicate as the Ollama handlers.
- Source-scan test: gone.
- `handleOllamaShow`: one comment, naming the right predicate.
- App: `idleEvictSecs` wired like `maxResidentMemGB` — field, `toCLIArgs`, decode, equality, `SettingsReset`, explainer, plus a snapping slider in Settings ▸ Server. Off by default. Screenshot below.

Three more, found while validating, each its own commit:

1. `freeCpuState` destroyed `ModelConfig` without freeing `ngram_table_path` — a leak on every failed cold load of a qwen4 entry, while `preloadCpuState`'s doc asserted the config owned nothing. `ModelConfig.deinit` is the one owner now.
2. The `.loading` precondition the guards rest on was true by accident: the boot load frees while `.unloaded`. Asserted rather than assumed.
3. `GET /props` fell through to `ensureLoaded`, so a status poll cold-loaded the model. The app's tray polls it every 3s, so with idle eviction on the sweep frees the memory and the next poll takes it back — the feature couldn't survive its own client. Nothing resident now answers from the counters. Separate commit in case you'd rather take it on its own.

`zig build test` 2320 pass / 153 skip / 0 fail; `cd app && swift test` 3319 / 0 fail.

New `tests/test_idle_evict.sh`: evicts after the window, reloads under `/api/tags` + `/api/show` + a chat request fired concurrently, and compares RSS across cycles that all end with nothing resident.

| build | RSS, cycles 2..6, nothing resident either end |
|---|---|
| this branch | 891 → 903 MB (+12) |
| four `releaseRetainedCpuState()` calls removed | 891 → 1829 MB (+938) |
| `/props` fix removed | fails at the status-poll arm |

Live on an M5 128GB with Qwen3.8-Flash-Next and the app's own slider at 5 min:
`[registry] idle-evicting model id=... (70.13 GB resident, idle 307s)` — server RSS 11.9 GB → 0.20 GB, and it stayed unloaded under `/props` polling.



Unload idle models, default Off

latent-variable:
Ready. Conflict resolved: merged `main` (your #396/#397), your five new CHANGELOG bullets kept above ours.

Re-validated on the merge, since #397 touches the same files:

- `zig build test` 2336 pass / 154 skip / 0 fail, and confirmed live rather than cached (inverting one assertion fails at 2335 naming the right test).
- `cd app && swift test` 3319 / 0 fail.
- `tests/test_idle_evict.sh` green: RSS 892 → 897 MB across cycles that all end with nothing resident.
- Flash-Next on an M-series box, real config: cold load 14.8s → correct answer → evicted at 73s idle (70.13 GB, RSS to 0.23 GB) → reload 10.4s → same correct answer. Zero `[mlx]` lines.

Also: the auto-context caveat I flagged above did not reproduce. Across evict→reload the advertised `context_length` held at 262,144 on both the explicit `--ctx-size` path and the auto path. It would still need real memory pressure at reload time to bite, so I'd leave it as a known interaction rather than something this PR introduces.

`scheduler.zig` is fmt-unclean on `main` already, so I left it rather than adding reformat noise to the diff.

PR #400, opened 2026-09-10
perf(batch): one sampling eval per tick, a fused-rows MoE arm, and one persistent group state for qwen4_exp batched decode

beamivalice:
## Summary

Three levers on the qwen4_exp batched decode tick, one squashed commit onto `feature/batch-decode-mtp`. Every lever is bit-identical per row against the solo path; the sampled/seeded and mixed-row cases have their own tests.

**1. One sampling eval per tick.** N rows draw in one eval (one argmax or one categorical draw over the stacked `[N, V]` logits); seeded or mixed-parameter rows concatenate per-row draws, each reshaped to rank one first (a greedy draw is `[1,1]`, a categorical one `[1]`; the mix used to throw inside MLX's concatenate and error every slot in the group). Per-row `seedKey` semantics unchanged.

**2. A fused-rows MoE arm at decode.** Row-axis twins of `gatherQmvGateUp` / `gatherQmvDownReduce` for S = 1 and 2 ≤ B ≤ 8, dispatched ahead of the sorted chain, which stays the fallback whenever the arm declines; the ten-expert reduction runs per row in the single-row order, and the Metal body is shared with the single-row kernels. N=1 keeps the old kernel byte for byte.

| cell (MTP off, A/B/B/A, fans max) | sorted chain | fused rows |
|---|---|---|
| ctx 50, N=2 tick | 24.9 ms | 22.0 ms |
| ctx 50, N=2 aggregate decode | 77 tok/s | 86 tok/s (+10–13% paired) |
| N=1 | unchanged | unchanged |
| 4k, N=8 | 70.7 ms | 71.1 ms (wash) |

`MLX_SERVE_MOE_ROWS_FUSED=0` restores the sorted chain (a genuine two-arm tradeoff at B=8).

**3. One persistent group state.** The tick used to concatenate every slot's GDN/conv/PLE state and hand back slice views each tick. It now owns one group generation across ticks: a stable membership reads the current generation and writes the next with no concat, slice or allocation; a member sent to any non-group forward (serial decode, prefill, the MTP solo arms) is released with its row materialized and evaluated first; a membership generation on the entries refuses a slot written outside the group; departures export the row, and a failed repack exports every survivor. Hermetic: the N=4 stable tick issues 0 merge/split ops after the first (2555 before over 7 ticks). Wall effect at these cells is within noise (+2–5% at short context across N=2/4/8, 4k inconclusive), as the research's 0.9–2.9 ms/tick estimate implied; the value is the ownership contract the batched KV append and the stacked indexer are designed on. Test seam `persistent_group_state_override` runs the old path for the byte comparisons; no env switch.

Also: a production-geometry test pinning the fused GDN prework at B=6/8/16 (Hk 16, Hv 48).

## Verification

Two Opus audit rounds on each lever (a P0 on the mixed greedy/sampled group and three P0s on group ownership were found and fixed red-first); full suite green; `tests/test_batched_equivalence.sh` on the production pack: forced-N=1 byte identity, first 30 tokens, both concurrent streams vs serial, the logprobs-beside-batched reason, kv-quant 8.

Design and measurements: `batched-decode-tick-research.md`, `batched-decode-max-design-codex.md` in the journal.

## L4 — batched affine-8 KV append (5f2e977)
One quantization over the stacked `[N, kv_h, 1, hd]` rows, then per-slot writes: `24N → 8+18N` ops per layer per tick; affine-8 only, seam `batched_kv_append_override`, one-shot engagement line. ABBA at 384k kv8 (medians of two arms): +1.4/+1.9% agg at ctx 50 N=4/8, +1.5/+4.4% at 4k N=4/8, neutral N<=2; byte-identical (`test_batched_equivalence.sh` 5/5 incl. kv8).

### DIFF
--- a/src/generate.zig
+++ b/src/generate.zig
@@ -14,6 +14,7 @@ const drafter_mod = @import("drafter.zig");
 const mtp_mod = @import("mtp.zig");
 const round_cost = @import("round_cost.zig");
 const ane_mod = @import("ane.zig");
+const scheduler_mod = @import("scheduler.zig");
 
 const Transformer = transformer_mod.Transformer;
 const Tokenizer = tokenizer_mod.Tokenizer;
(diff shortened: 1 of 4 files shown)


ddalcu:

- `src/generate.zig:9800` — The `errdefer` and the `defer` cover the same `parts[0..n_parts]` range and are both live once the loop ends, so an error out of the concat frees every part twice.
  Failure scenario: the `errdefer` at :9800 is needed for a mid-loop `mlx_reshape` failure (the `defer` is not registered yet), but from :9811 both are armed. `return concatLogitRows(parts[0..n_parts], s)` (:9814) propagates an error union, which is an error return from this scope — so a failing `mlx_concatenate_axis` inside `concatLogitRows` runs the `defer` and then the `errdefer`, double-freeing each of the up-to-8 per-row draws. That is the mixed seeded/greedy path this PR adds, and it is the arm taken whenever any row carries a seed, so the crash lands on a group of live requests rather than one. Registering a single `defer` before the loop (or zeroing `n_parts` before the return) covers both exits without the overlap.

beamivalice:
Fixed in 6359898: `sampleRowsLazyConcat` now registers one `defer` before the loop that releases `parts[0..n_parts]` on every exit, so a mid-loop reshape failure and a concat failure both free each draw exactly once; the concat output holds its own references. No behaviour change on the success path.

The error path has no hermetic red-first test: every part is a rank-one int32 draw, so `mlx_concatenate_axis` cannot be made to fail from inputs without a fault-injection seam, which is more machinery than the fix. Verified: ReleaseFast build, the `sampleRows` filter (22/22) and the full suite (2284 passed, 154 skipped) on the amended commit.

ddalcu:
🤖 **MLXServe-PRBot approved** — reviewed at head `6359898`, no issues found.

ddalcu:

- `src/transformer.zig:6167` — Each field is freed before its replacement slice is computed, so a failing slice leaves the struct holding a freed handle that the `defer deinit()` frees again.
  Failure scenario: the pattern at :6167-6172 (and the `row_vq` twin at :6179-6184) is `_ = mlx_array_free(row_kq.q); row_kq.q = try sliceStackedRow(...)`. If `mlx_slice` fails — OOM is the realistic trigger on a server that admits against a memory ceiling — the `try` returns with `row_kq.q` still pointing at the handle freed one line earlier, and `QuantizedKV.deinit` (`src/kv_quant.zig:145`) frees all three fields unconditionally, so that handle is freed twice. There are six such windows per row and up to eight rows per layer, and it is the affine-8 path that every batched qwen4_exp group takes. Computing the slice into a temp and assigning after the free (or building the struct straight from the three slices) closes it; this is the same shape as the `sampleRowsLazyConcat` double free fixed at `6359898`.

PR #401, opened 2026-09-10
perf(batch): batched affine-8 KV append for qwen4_exp decode (L4)

beamivalice:
Stacked on #400 (its two commits show here until it merges; L4 is `bb92af4` + the engagement line `40f41a3`).

## What
Batched decode appended each slot's new K/V row with its own quantize + slice + update chain: `24N` ops per layer per tick. L4 quantizes ONCE over the stacked `[N, kv_h, 1, hd]` rows (`KVCache.appendBatched` → `appendAffineBatched`) and then does only the per-slot writes: `8 + 18N`. Affine-8 only (the served width); dense, affine-4 and turbo stay on the per-slot path. Seam: `batched_kv_append_override` (default on). A one-shot `[batched] affine-8 kv append engaged (slots=N)` line proves an arm.

## Numbers (ABBA, base/l4/l4/base, medians of 2 arms each, ctx 384k kv8, `bench_batching` at ctx 50 and 4k)

| ctx | N | base agg tok/s | L4 | Δ |
|---|---|---|---|---|
| 50 | 1 | 55.5 | 53.4 | −3.9% (kernel not on the N=1 path; boot drift) |
| 50 | 2 | 79.6 | 78.5 | −1.3% |
| 50 | 4 | 106.8 | 108.2 | +1.4% |
| 50 | 8 | 124.5 | 126.8 | +1.9% |
| 4k | 1 | 34.4 | 35.1 | +2.2% |
| 4k | 2 | 41.0 | 41.2 | +0.5% |
| 4k | 4 | 47.7 | 48.4 | +1.5% |
| 4k | 8 | 50.9 | 53.1 | +4.4% |

Small and positive where it engages (N ≥ 4), neutral at N ≤ 2. Byte-identical: `tests/test_batched_equivalence.sh` 5/5 incl. the kv8 arm on `bb92af4`.

## Tests
Four unit gates in `src/transformer.zig` (`batched KV append: …`): affine-8 N=2/4/8 matches N solo updates at production geometry; two restored siblings append and truncate independently with a byte-identical snapshot; per-layer op count is quantize-once plus per-slot writes; dense/affine-4/turbo stay per-slot. `zig build test -Doptimize=ReleaseFast` filtered runs green; the equivalence script as above.

### DIFF
--- a/src/generate.zig
+++ b/src/generate.zig
@@ -14,6 +14,7 @@ const drafter_mod = @import("drafter.zig");
 const mtp_mod = @import("mtp.zig");
 const round_cost = @import("round_cost.zig");
 const ane_mod = @import("ane.zig");
+const scheduler_mod = @import("scheduler.zig");
 
 const Transformer = transformer_mod.Transformer;
 const Tokenizer = tokenizer_mod.Tokenizer;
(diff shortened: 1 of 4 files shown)


beamivalice:
Folded into #400 as a single squashed commit per the owner.

Issue #116, opened 2026-08-01
Support Qwen3 embedding models with last-token pooling on /v1/embeddings and /api/embed


dpearson2699:
## Feature request

Please add native MLX serving support for the Qwen3-Embedding family, including `Qwen3-Embedding-4B`, through both the OpenAI-compatible `POST /v1/embeddings` endpoint and the Ollama-compatible `POST /api/embed` endpoint.

## Use case

I use [GitNexus](https://github.com/abhigyanpatwari/GitNexus) v1.6.9 to index source code with a configurable OpenAI-compatible embedding endpoint. I want to replace Ollama with MLX Core and serve [mlx-community/Qwen3-Embedding-4B-4bit-DWQ](https://huggingface.co/mlx-community/Qwen3-Embedding-4B-4bit-DWQ) natively through MLX.

GitNexus expects 2,560-dimensional vectors from this model and sends batched requests to `/v1/embeddings`. mlx-serve's Ollama compatibility is also useful for other local clients, so equivalent behavior on `/api/embed` would make this a true drop-in migration.

## Current compatibility gap in v26.7.12

The HTTP routes exist, and the native MLX engine can load Qwen3-family transformer weights, but the current embedding implementation is not semantically correct for Qwen3-Embedding:

- `computeEmbeddingsBatch` in `src/generate.zig` always performs masked **mean pooling** followed by L2 normalization.
- The official [Qwen3-Embedding-4B model card](https://huggingface.co/Qwen/Qwen3-Embedding-4B) requires **last-token pooling** followed by L2 normalization.
- Its Sentence Transformers metadata explicitly sets `pooling_mode_lasttoken: true` and `pooling_mode_mean_tokens: false`.
- The MLX conversion identifies itself as `Qwen3ForCausalLM` / `model_type: qwen3`, with `hidden_size: 2560`; it is not marked `is_encoder_only`.
- The README currently describes `/v1/embeddings` as supporting “BERT and encoder-only models.”

As a result, pointing this Qwen3 checkpoint at the existing embedding endpoint would either fail capability discovery or, worse, return correctly shaped but semantically incorrect mean-pooled vectors. I did not migrate the production index because silent embedding corruption is harder to detect than an explicit unsupported-model error.

## Requested behavior

1. Detect the checkpoint's Sentence Transformers pooling metadata, particularly `1_Pooling/config.json`, or otherwise recognize Qwen3-Embedding checkpoints.
[...]

PR #404, opened 2026-09-11
fix(embeddings): each sub-batch starts from an empty KV cache

### LINKED ISSUE #403: Qwen3-Embedding: a request split into sub-batches returns wrong vectors or crashes the server
### Before opening

- [x] I searched open and closed issues, open PRs and recent commits. #116 (item 4) asked for exactly this when Qwen3-Embedding support landed ("a Qwen causal forward must reset or bypass its KV cache for each independent chunk"), and `docs/reference.md` describes the trunk as cache-less, but the shipped path resets once per request. #266 has the same `freeKVEntry` crash stack from a different trigger and is fixed. #343/#351 do not touch the sub-batch loop. Still present on `main` (fa76a4b).

**Priority:** high (blocks my use of mlx-serve) · **Area:** Server / API (embeddings), Crash (SIGSEGV)
**mlx-serve version:** 26.9.2 (MLX 0.32.2), MLX Core app; present since v26.8.1
**Hardware + macOS:** Mac mini M4 Pro 48 GB (Mac16,11), macOS 26.6.2
**Model:** `chrisyuan/Qwen3-Embedding-0.6B-mlx-bf16`; also `mlx-community/Qwen3-Embedding-0.6B-4bit-DWQ`. Default flags (kv-quant off).

### What happened vs what you expected

A Qwen3-Embedding request that `computeEmbeddingsBatch` splits into sub-batches (rows × longest input over `EMBED_TOKEN_BUDGET`, e.g. 64 inputs with one over 512 tokens) should embed every input as if it were sent alone. Instead the sub-batches share one KV cache:

1. Where the cache write can carry the first sub-batch's rows into the second's shape (typically a second sub-batch no larger than the first), the request answers 200 with wrong vectors for the later rows. On real text (1500-character chunks of this repo's sources and docs, 64 per request over `/api/embed`) sampled rows came back as low as cosine 0.31 to the same input embedded alone, with nothing in the log.
2. Otherwise the cache write fails with `[mlx] [broadcast_shapes]` and a 500, and the model's **next** embeddings request (or unloading the model) crashes the server (SIGSEGV in `freeKVEntry`, stack below).

Cause: Qwen3-Embedding runs the ordinary decoder forward, which appends to `xfm.cache`, and `runEmbedRequest` resets that cache once per request, not per sub-batch. Sub-batch 2 writes its keys after sub-batch 1's and attends to them at shifted positions. The crash is a second defect: `KVCache.updateDense` frees `key_view`/`value_view` before the write and reassigns them only after it, so a failed write leaves freed handles that the next `resetCache` frees again. Case 1 happens under any `--kv-quant`; the crash only with kv-quant off, since `updateAffine` already resets its handles.

Found while indexing a Zotero library through zotero-mcp's Ollama provider (`/api/embed`, 64 chunks per request): two crashes in about 35 minutes. Lowering the client batch to 48 happened to stop the observed crashes, probably because fewer requests split. It is not a workaround: a request that still splits can return wrong vectors with a 200 and, depending on the sub-batch sizes, can still hit the 500 and the crash.

### Reproduction steps

`mlx-serve --serve --port 11234 --model <Qwen3-Embedding-0.6B dir>`, then (stdlib Python; the repeat counts give ~1006 and ~1106 Qwen3 tokens per input, so 33 of the first or 30 of the second exceed 64 × 512):

```python
import json, math, sys, urllib.request

URL = sys.argv[1] if len(sys.argv) > 1 else "http://127.0.0.1:11234"

def embed(xs):
    r = urllib.request.Request(URL + "/v1/embeddings", json.dumps({"model": "mlx-serve", "input": xs}).encode(),
                               {"Content-Type": "application/json"})
    rows = json.load(urllib.request.urlopen(r, timeout=300))["data"]
    return [d["embedding"] for d in sorted(rows, key=lambda d: d["index"])]

cos = lambda a, b: sum(x * y for x, y in zip(a, b)) / math.sqrt(sum(x * x for x in a) * sum(y * y for y in b))
short = [f"note {i}" for i in range(35)]
long_ = [f"document {i}: " + "lorem " * 1000 for i in range(32)]

# 1. Sub-batches of 32 short + 32 long: HTTP 200, but the long rows are wrong.
rows = embed(short[:32] + long_)
print("worst cosine, batched vs alone:", round(min(cos(r, embed([t])[0]) for r, t in zip(rows[32:], long_)), 4))

# 2. Sub-batches of 29 longer + 35 short: the cache write fails.
try:
    embed([f"document {i}: " + "lorem " * 1100 for i in range(29)] + short)
    print("second request: 200")
except Exception as e:
    print("second request:", e)

# 3. The next embeddings request resets the damaged cache: the server segfaults.
try:
    embed(["anything"])
    print("third request: 200")
except Exception as e:
    print("third request:", e)
```

On 26.9.2 this prints `worst cosine, batched vs alone: 0.86`, `second request: HTTP Error 500`, `third request: Remote end closed connection without response`, and the server is gone.

### Server log tail

```text
POST /v1/embeddings (64 inputs)
  <- 64 embeddings (22705 tokens)
POST /v1/embeddings (64 inputs)
[mlx] [broadcast_shapes] Shapes (28,8,1134,128) and (36,8,1134,128) cannot be broadcast. at /Users/runner/work/mlx-serve/mlx-serve/lib/mlxc-src/mlx/c/ops.cpp:3196
  embedding error: MlxError
POST /v1/embeddings (64 inputs)
```

Crash report (identical on both production crashes), faulting thread:

```text
EXC_BAD_ACCESS (SIGSEGV) KERN_INVALID_ADDRESS
0 libmlx.dylib    mlx::core::array::~array() + 36
1 libmlxc.dylib   mlx_array_free + 16
2 mlx-serve       transformer.freeKVEntry + 72
3 mlx-serve       transformer.KVCache.deinit + 44
4 mlx-serve       transformer.Transformer.resetCache + 116
5 mlx-serve       scheduler.inferenceLoop + 7924
```

I have both fixes with tests and will open two PRs linking this issue: the per-sub-batch reset, and the `updateDense` handles.

josk0:
**What changed / why**

Fixes #403. Qwen3-Embedding forwards through the decoder's KV cache, and `runEmbedRequest` reset it once per request instead of once per sub-batch as #116 (item 4) asked, so a request split by `EMBED_TOKEN_BUDGET` embedded its later sub-batches against the earlier rows: a 200 with wrong vectors, or a failed cache write that crashed the next embeddings request (#405 makes that failure cost one request).
- The reset moves into `computeEmbeddingsBatch`, before every sub-batch. The per-request resets in `runEmbedRequest` and the server fallback are removed, and the two comments that called embeddings encoder-only are corrected.
- Test [4c] in `tests/test_embeddings.sh`, a rule + gotcha, the `docs/reference.md` row that called the trunk cache-less, a CHANGELOG line.

**How I verified it** (M4 Pro 48 GB, macOS 26.6.2)
- `zig build test`: 9/9 steps, 2315 pass, 166 skip, 0 fail.
- `tests/test_embeddings.sh` [4c] on `mlx-community/Qwen3-Embedding-0.6B-4bit-DWQ` and `chrisyuan/Qwen3-Embedding-0.6B-mlx-bf16`: `main` fails all three checks (later sub-batch vs the same inputs sent alone: cosine 0.86 / 0.84; HTTP 500; server segfault), this branch passes (1.0000); the BERT checks pass on both.
- Served it to a zotero-mcp-shaped client on the bf16 model: 80 `/api/embed` requests of 64 real 1500-character chunks, all 80 split: no errors, sampled rows ≥ 0.9997 cosine to the same input embedded alone (`main`: as low as 0.31 within 12 requests).

https://claude.ai/code/session_017uayEz8B1LK1dexi5FDxxe


ddalcu:
🤖 **MLXServe-PRBot approved** — reviewed at head `780a525`, no issues found.

PR #405, opened 2026-09-11
fix(kv): a failed dense KV write leaves no freed view handle behind

josk0:
**What changed / why**

Refs #403. `KVCache.updateDense` freed `key_view`/`value_view` before the grow and the writes and assigned fresh handles only after them, so an MLX error in between (catchable since #353) left freed handles in the entry, and the next reset or free of that cache freed them again: SIGSEGV in `freeKVEntry`. #403 hit it through Qwen3-Embedding sub-batches (#404 removes that trigger), but any failing dense KV write reaches it.
- The handles are reset at the free, as `updateAffine` already does (`updateTurboQuant` goes through `updateAffine`); `writeAtOffset` releases its result on the error path.
- A fault sweep beside the `growQuantBuf` one, over every checked op of an in-capacity and a growing update; a rule + gotcha; a CHANGELOG line (it conflicts with #404's `### Fixes` hunk; I'll rebase whichever lands second).

**How I verified it** (M4 Pro 48 GB, macOS 26.6.2)
- `zig build test`: 9/9 steps, 2316 pass, 166 skip, 0 fail. With the `updateDense` hunk reverted the sweep aborts the test runner: a stale view fails its assertion and the cleanup double-frees it.
- The issue's script against this branch alone, where the sub-batch write still fails: the second request still gets its 500, and the server now answers the third instead of crashing.

https://claude.ai/code/session_017uayEz8B1LK1dexi5FDxxe


ddalcu:
🤖 **MLXServe-PRBot approved** — reviewed at head `7c18d5d`, no issues found.

PR #406, opened 2026-09-11
launch(opencode2): vendor the monitor plugin in-tree

beamivalice:
The opencode2 monitor plugin now lives in this repository as plain tracked files at `lib/opencode2-mlx-serve`, instead of a submodule pinned to another repo. One repository owns the code that `lib/opencode2_plugin.zig` embeds into the binary and `app/build.sh` copies into the app bundle: a plain clone builds and ships the plugin with no submodule step, and a plugin change can land in the same commit as the server change it needs.

The vendored tree is the plugin's current state, three changes ahead of the old pin:

- **A session served by another provider is metered from its own stream.** Its wait counts up, `session.step.streamed` fixes TTFT, the streamed bytes estimate decode, and the step's usage settles a prefill rate — instead of borrowing this server's feed and reporting this Mac's numbers for a cloud turn.
- **Prefill bars cap at 12 blocks**, and the sidebar's Throughput line draws the rate it names.
- **The provider gate takes a list**, defaulting to both ids our config writers register: `mlx-serve` and the launcher's `mlx`.

That last one supersedes the writer-side workaround the previous revision of this PR carried. `opencodeJson`, `mergeOpencode2CliJson` and the Swift `opencode2CliJSON` no longer write a `provider` option; the shared `opencode_provider_id` constant, its Zig and Swift tests, and the `tests/test_launch_cmd.sh` assertion are gone. The plugin's default gate already accepts the id the launcher registers, and naming a single id in the written config would only narrow it.

`NOTICE` gains the plugin's MIT entry — its copyright holder differs from mlx-serve's and it now ships inside this tree. `.gitmodules` keeps only the mlx, mlx-c and ds4 entries; `scripts/check-submodules.sh` reads `.gitmodules`, so it stops covering the path on its own.

Gates, all green on the final tree:

| Gate | Result |
|---|---|
| `cd lib/opencode2-mlx-serve && node --test` | 149 pass, 0 fail |
| `zig build -Doptimize=ReleaseFast` | exit 0; the binary carries the vendored bytes |
| `zig build test -Doptimize=ReleaseFast --summary all` | 9/9 steps, 2327 pass, 154 skip, 0 fail |
| `zig build test -Dtest-filter="launch"` / `"opencode2"` | exit 0 |
| `cd app && swift test --filter CLISetupInstructionsTests` | 23 tests, 0 failures |
| `tests/test_release_workflow_gates.sh` | 75 PASS, 0 FAIL |
| `git ls-files -s lib/opencode2-mlx-serve` | 22 files, all `100644`, no gitlink |


ddalcu:
🤖 **MLXServe-PRBot approved** — reviewed at head `21628fd`, no issues found.

PR #407, opened 2026-09-11
Generalize JSON-constrained reasoning across supported formats

perretv:
**What changed / why**

This generalizes #333 (related: #331) from Qwen-style reasoning to bare and suffixed think tags, Gemma, Inkling, Harmony, and Muse on the MLX path. Models can reason before a schema-constrained JSON answer, or answer directly where their format allows it. JSON with thinking disabled still works.

One shared state machine handles the format differences, with tokenizer data cached per model. It checks tokens that cross into JSON before sampling. All three APIs share output routing, so reasoning markers inside JSON strings remain data. `max_tokens` still covers reasoning and the answer together; exhausting it during reasoning can return empty content with a length finish reason. Finite response-side reasoning budgets and unsupported headers keep the thinking-off fallback.

**How I verified it**

- Apple M4 Max, macOS 26.6.2: ReleaseFast build; full Zig suite, 2,350 passed / 166 skipped; Swift suite, 3,317 tests / 24 skipped / zero failures; Swift release build.
- `tests/test_json_schema_thinking.sh`: existing schema/token-limit checks plus 12 live Qwen3.5-0.8B combinations covering Chat Completions, Messages, Responses, streaming on/off, and thinking on/off.
- Deterministic tests cover all six formats, split tokens, partial headers, recovery, invalid JSON at boundaries, direct answers, and output routing. Other formats still need live checkpoint validation; no model-specific performance improvement is claimed.


ddalcu:
thank you @perretv , i merged latest main, added a small fix, and merged.

PR #408, opened 2026-09-12
perf(qwen4): consolidate prefill fusions, bounded PLE and MoE paths

nikolai-vysotskyi:
Superseded by **#438**, the requested HC/GDN-only PR on fresh main. This branch retains the combined implementation for reproduction and further investigation.

**Reproduced on September 15: 2323.423 tok/s**, within 1.01% of the earlier record ([data](https://github.com/nikolai-vysotskyi/mlx-serve/blob/ce3371d/research/followups/replay-20260915/result.json)).

**2347.1 tok/s is a recorded M5 Max 128 GB prefill result:** 64,947 uncached input tokens, 27.671 s server prefill, 27.888 s client wall time. Reproducing it and identifying the changes required for normal serving remains an open task. Materially lower comparable M5 results require investigation; describing the measurement as special does not resolve the discrepancy.

The recorded `087c210` profile enabled QSA-pair, HC/GDN, PLE-ahead/packed, MoE-group and HC-upmix. Its matched comparison was 2173.8→2347.1 (+8%); both arms already had QSA+HC+GDN. The default-path reviewer results used a different enabled set: [M4 Max A/B](https://github.com/ddalcu/mlx-serve/pull/408#issuecomment-5674934215), [M5 Max A/B](https://github.com/ddalcu/mlx-serve/pull/408#issuecomment-5658856761). Isolating these differences is part of the investigation. #438 carries the independently reviewed HC/GDN subset.


beamivalice:
Screenshot 2569-09-12 at 12 15 54


Very good indeed, but not that 2500 tok/s on the 8bit/4bit hybrid quant. You also need to enable 8192 chunks otherwise the mlx-serve will silently use 4096.

nikolai-vysotskyi:
@ddalcu @beamivalice — long-context validation is now complete for PR #408, head `092ce2e`, on M5 Max 128 GB / macOS 26.5.

| Checkpoint / arm | Input tokens | Long prefill tok/s |
|---|---:|---:|
| Original 4-bit, off | 68,651 | **1644** |
| Original 4-bit, on | 68,447 | **1775** |
| Mixed 4/8-bit, off A | 68,447 | **1626** |
| Mixed 4/8-bit, on | 68,539 | **1879** |
| Mixed 4/8-bit, off B | 68,651 | **1643** |

Mixed checkpoint gain: **+14.36% to +15.56%**, with disabled-control drift +1.05%. Original 4-bit gain: +7.97%. These are `bench.contextScaling[0].prefillTokPerSec` medians from llmprobe 0.6.6, three coding requests per cell. All three opt-ins are toggled together in the same executable; no weights changed.

**2509.4 tok/s was the earlier 15,715-token HTTP cell; it does not reproduce as a 64K+ result. We have not reached 2500–3000 tok/s or >1.5× on long context.**

I checked your comments: explicit `--prefill-chunk 8192` is required and every long request's trace confirms nine chunks at width8192. Following the request in #375, the recommended `mixed-4-8bit` checkpoint is measured separately, in off/on/off order. The old `4bit` pack has 793 quantized 4-bit tensors plus an 8-bit lm_head; the mixed pack has 148 at 4 bits and 646 at 8 bits (group64). They are not interchangeable baselines.

Fixed settings: ctx131072, prefix cache entries0, KV quant off, MTP off. PLD is requested on at launch/the HTTP gate, but the Qwen4 scheduler disables it before Generator init; neither the short smoke nor these long runs execute PLD. Paired QSA, HC and GDN engagement is present only in ON logs. Standard llmprobe nonces/calibration give a small recorded prompt-length difference across boots. Both control runs and all samples are retained.

[Raw JSON, logs, source/runtime/model fingerprints, exact launch and probe reproduction](https://github.com/nikolai-vysotskyi/mlx-serve/tree/6a0ee0617630f5c2936c6ffaaf6a7b77ef928139/research/long_context). `reproduce.sh` defaults to the recommended mixed pack. PLE #375 is absent and its gains are not included. The PR body now names the measured long-context result.

Could you review the paired planner/gather and HC numerical contract? The patch remains opt-in and draft for startup fallback, planner admission accounting and broader HC numerical coverage. The frozen implementation already passed ReleaseFast and the full Zig suite, and the new long evidence is now attached for review.

beamivalice:
Is this branch on top of main? and on/off = pr408 vs main? or last release?

Good numbers at 64k context, btw.

nikolai-vysotskyi:
@beamivalice Yes, PR #408 is based directly on upstream `main` at `fa76a4b50b3f54af7e9cd927279f5ba2870f02c6`; the tested PR head is `092ce2e`. It does not include PLE PR #375. OFF/ON here uses **the same PR executable**, toggling `MLX_SERVE_QSA_PAIR`, `MLX_SERVE_HC_PREFILL` and `MLX_SERVE_GDN_PREFILL_FUSED` together. OFF dispatches the pre-existing paths; it is not a separately built clean-main or last-release executable. I have named that control explicitly in the report and PR body.

For the mixed ON arm, the server's own coding-request median is 1882.5 tok/s and the llmprobe long-context median is 1879, so the client-versus-server timing definition is only about 0.2% here.

Could you share the benchmark and server launch commands, actual prompt-token count, tested commit/runtime and timing source for your 2060.2 / 2083.1 screenshot? We want to account for that gap and reproduce your setup. Our fixed settings are chunk8192, ctx131072, KV quant off, prefix cache entries0, MTP default off, PLD requested on but disabled by Qwen4 scheduler wiring, llmprobe 0.6.6 rung64k/runs3. Your screenshot's disabled cells (1877.4 / 1684.5) also differ from our 1626 / 1643, so the absolute rates need a matched workload before attribution.

A fresh short bridge using the existing HTTP recall script on the same PR binary and the mixed checkpoint gives **2141.6 tok/s at 15,715 tokens** (first request: 1985.5 at 13,515), correct recall, zero cached tokens. This confirms the 2080+ range on a short request here; it is not a matched reproduction of your screenshot or a new long-context improvement.

Correction to the initial settings prose: Qwen4 `moduleSpecWiring()` causes `specInitWiring()` to return `use_pld=false`. The HTTP gate’s “pld=enabled” line is intent, not final engagement. All measured Generator traces lack `[pld]` and PLD `spec-stats`; PLD is therefore not an explanation for the short/long difference. The reports now distinguish these two levels.

[Short bridge evidence and corrected execution settings](https://github.com/nikolai-vysotskyi/mlx-serve/tree/6a0ee0617630f5c2936c6ffaaf6a7b77ef928139/research/long_context).

beamivalice:
16k prompt, greedy, n=512. byte identical.

cameronbergh:
I ran an isolated A/B on an **M4 Max (128 GiB)** at the current PR head `092ce2e0d11ebe42467f681347c19ac92da8a4e4`.

Setup:

- retained Qwen3.8 Flash-Next 4-bit mlx-serve checkpoint
- fresh process per arm, ordered OFF1 / ON1 / OFF2 / ON2
- `--ctx-size 131072 --prefill-chunk 8192`
- prefix cache disabled; KV quantization off
- MTP and PLD disabled
- greedy, fixed seed, 512 generated tokens
- `cached_tokens=0` for every measured request

Results:

| Prompt | OFF prefill | ON prefill | Delta | OFF gen | ON gen | TTFT delta |
|---|---:|---:|---:|---:|---:|---:|
| 15,753 tokens | 396.801 tok/s | 381.372 tok/s | **-3.89%** | 61.299 | 61.378 | **+3.96%** |
| 64,170 tokens | 749.071 tok/s | 771.546 tok/s | **+3.00%** | 58.668 | 58.720 | **-2.91%** |

The 64K improvement reproduced closely in both alternating pairs. The 16K result was noisy: ON was -9.30% in the first pair and +1.93% in the second.

Kernel engagement:

- HC prefill and fused GDN prefill engaged in both ON arms at width 8192.
- QSA pair did **not** engage because NAX is unavailable on this M4 (`applegpu_g16s`).
- Therefore this validates only the HC+GDN portion locally, not the complete QSA+HC+GDN path.

Correctness note: requests were byte-identical and outputs were deterministic within each feature state, but OFF and ON outputs differed at both prompt lengths (first differing bytes 597 and 108 respectively). Decode speed was effectively unchanged.

Resource use remained bounded: minimum sampled headroom was about 31 GiB and peak process footprint was about 75.6 GiB. Production was stopped during the test and restored afterward.

My takeaway: there is a small reproducible long-context prefill signal on M4 Max, but this result does **not** validate the combined PR for production—especially without NAX/QSA coverage and with deterministic output divergence. Happy to provide raw logs or rerun a corrected head with more fresh-process repetitions.

nikolai-vysotskyi:
@cameronbergh Thank you for isolating the M4 path and reporting deterministic output divergence. I pushed a concrete correction at **1b1a4dfbfd3d5f5c2bc9c4e61963902d5089d918**: HC now uses native MLX matmul for the four inject channels, retaining the fused pending write, normalization and stream mixing.

The old one-hot test missed this reduction-order issue. A dense BF16 inject-weight regression fails on the previous kernel (max raw difference 0.0000076293945) and passes exactly on the corrected kernel across cold/pending B2/S17, B1/S65 and B1/S513. On M5, all five HC component outputs now match stock bit-for-bit with both 4-bit and 8-bit projections. Component time is 4.43 ms vs stock 6.43–6.46 (4-bit), and 4.73 vs 6.54–6.63 (8-bit). The corrected path is slower than the previous fused-inject path; the historical 1879 tok/s long result is **not** a measurement of this new head.

ReleaseFast, the full Zig suite (2329 passed, 154 skipped) and mixed-checkpoint uncached HTTP recall/engagement checks passed on the corrected PR binary. [Code, red/green evidence, component results and fingerprints](https://github.com/nikolai-vysotskyi/mlx-serve/blob/af8107f/research/followups/20260912-hc-native-inject.md).

This removes a known HC numerical difference; it does not prove that HC caused all of the M4 generation divergence or that model quality is validated. Could you attach the raw logs/request metadata from the existing run? If you rerun, one identical OFF vs HC-only pair on this corrected head, followed by GDN-only if needed, would isolate more than additional repetitions of the old combined head. No QSA claim on the M4 fallback. Keeping the PR draft pending that isolation and the outstanding readiness work.

@ddalcu The HC correction is ready for review. Larger MoE work remains separate research; no additional whole-model speedup is claimed here.

beamivalice:
please submit your new speed benchmark results based on the new correct binary too. @nikolai-vysotskyi - llmprobe at 4k/8k/16k/64k/128k --runs 3, both prefill and decode.

ddalcu:
All three fusions are opt-in `=1` levers, default off. Merged, it is dead code. CLAUDE.md is explicit: a bit-identical win ships default on with a `=0` kill switch (like MLX_SERVE_HC_FUSED, MLX_SERVE_GDN_DECODE_FUSED); a lossy path gets a lever plus a stated bar. QSA pair is lossy and NAX-only, so it never engages on M4 at all. HC and GDN are the only parts that run here and cameronbergh's M4 Max A/B on 092ce2e was -3.9% at 16k, +3.0% at 64k with greedy divergence, and the 1879 tok/s headline was measured before the inject fix that makes HC slower. So today there is no number on the current head that says this pays anywhere.

Things that have to change before I look again:

1. Lever polarity. HC and GDN default on with `=0` switches if you claim bit-identical; QSA pair keeps its lever with the bar written in a comment and a CLAUDE.md rule line.
2. Bit-identical against what runs in production. The hc prefill test compares against an eager chain the test builds. Production runs compiled_hc_write/compiled_hc_mix. mlx_compile is not output preserving. Compare against a Transformer instance, or state the bar as "vs eager, greedy flips legit" and prove the flips are near-ties.
3. Unbilled transient. qsa_pair.zig allocates tilepos + tilemask per full-attention layer per chunk, ~70 MB at S=8192 (x2 at batch 2). Under-billing is a Metal OOM. Add it to the qwen4 admission terms or cap supports() below it. You listed this as outstanding yourself.
4. supports() silently declines past kv 131072 on a 1M context model, so a 200k prompt flips kernels mid-request with no log.
5. hc_prefill.zig hardcodes hc==4, hidden==2560 and 10240/40960/640 literals. Derive from hc*hidden or the next pack gets a silent fallback.
6. No CLAUDE.md rule, no gotcha, no CHANGELOG. tests/test_qsa_pair_prefill.py needs a live server and is not wired into any tests/*.sh; wrap it or drop it, and default it to the mixed pack, the -4bit one is gone.

Then post OFF/ON on the current head, same binary, fresh process per arm, with the engaged lines in the ON log, on both an M5 and an M4. I can run the M4 side once the head is stable. (using the method @beamivalice mentioned above llmprobe at 4k/8k/16k/64k/128k --runs 3)

nikolai-vysotskyi:
@ddalcu @beamivalice @cameronbergh I have read the full review. The current PR head is still **1b1a4df**; the old **1879 tok/s** llmprobe number belongs to **092ce2e** and is not a benchmark of the corrected HC path.

There is newer corrected-HC evidence, but it must be labelled accurately: on the mixed pack, an identical-binary HTTP screen measured **2071.3 tok/s at 62,176 tokens** with QSA+corrected HC+GDN, versus **2154.0** with the separate bounded PLE-ahead experiment. These are one sequential pair, not the requested llmprobe ladder, and PLE/MoE research is **not in this PR**. [Raw data and provenance](https://github.com/nikolai-vysotskyi/mlx-serve/blob/825cff3/research/followups/20260912-prefill-ahead-and-grouping.md).

I am fixing the review points in this PR before freezing the next measured head: use the production compiled HC closures as the comparison, derive HC geometry, bill the pair planner, report unsupported long-context fallback, add the precision/default rules and wire the HTTP check. I will only claim bit identity/default-on where the production comparison supports it; the eager fixture alone did not establish that. QSA remains explicitly approximate with its existing float64 accuracy bar.

The acceptance command will use the mixed pack, **llmprobe 4k,8k,16k,64k,128k, --runs 3**, both prefill and decode, with the same binary, fresh-process OFF/ON arms and engagement logs. I will post the M5 results against the stable new head and then tag you for the M4 side. I am keeping the PR draft until then. The 65K throughput target being pursued locally is 2400 tok/s; it has not been demonstrated yet.

nikolai-vysotskyi:
@ddalcu @cameronbergh The reviewed head is now **d23d9df** (binary SHA256 `580f495dce825202ea2da7976e10f6b1c4db39158827e7d63735f60a79f0d476`).

Implemented the six code/documentation points:

- HC and GDN now default on with `=0` kill switches; paired QSA stays opt-in with its explicit approximate float64 bar.
- HC parity uses `Transformer.compileQwen4Hc()` / the actual compiled write and mix closures; GDN also compares with `computeGdnGate()` after compilation. Both passed on this M5. This is fixture evidence, not a claim that the M4 greedy difference has been resolved.
- Compressed planner positions to one base per four-key block: **71,909,376 → 19,611,648 bytes per S8192/B1 layer** including masks; B2 doubles it. The same allocation is now added to the QSA admission transient.
- Removed the arbitrary 131072 cap: supports now reaches **1,048,576 KV**, with a once-only log for unsupported fallback. Independent float64/key-coverage checks passed at 200003 and 1048576 KV as well as batch/tail/65K fixtures.
- HC derives widths and threadgroup size from the weight geometry; the compiled-path fixture also exercises HC2/H1536.
- Added CLAUDE rules, gotcha, changelog, a disposable-server `tests/test_qsa_pair_prefill.sh` wrapper defaulting to the mixed pack, and a portable fresh-process llmprobe driver.

Validation: ReleaseFast **7/7**; full Zig suite **2330 passed / 154 skipped**; HTTP mixed-pack recall at 13515 and 15715 tokens, `cached_tokens=0`, QSA+HC+GDN engaged. The requested M5 4k/8k/16k/64k/128k `--runs 3` OFF/ON ladder is running on this frozen binary, with a final 64K OFF drift control. No new speed headline until it finishes. `tests/bench_qwen4_prefill.py` records launch flags, binary hash, engagement and JSON reports; MTP/PLD/drafter are explicitly off.

@ddalcu This is the stable head for the M4 comparison you offered. Leave QSA disabled there and compare the HC/GDN kill switches; the first two unit tests above can be run before the ladder to catch an architecture-specific compiled-path difference cheaply. PR remains draft pending the current measurements and M4/quality follow-up.

nikolai-vysotskyi:
@ddalcu @beamivalice @cameronbergh The requested **current-head full ladder is complete on M5 Max 128 GB**, frozen **d23d9df** (`580f495d…0d476`), mixed-4-8bit, chunk8192, ctx262144, cache entries0, KV quant off, MTP/PLD/drafter explicitly off. PLE and MoE research opt-ins were disabled in both arms.

| Target | Actual OFF / ON tokens | OFF prefill | ON prefill | Observed change | OFF / ON decode |
|---:|---:|---:|---:|---:|---:|
| 4K | 4239 /4235 |1434|1489|+3.8%|55.1 /50.0|
| 8K |8235 /8231|1599|1623|+1.5%|56.2 /50.5|
| 16K |16298 /16311|1531|1735|+13.3%|57.2 /50.8|
| 64K |65841 /65778|1517|1761|+16.1%|49.6 /50.9|
| 128K |131086 /131133|1447|1771|+22.4%|48.5 /49.1|

All rates tok/s, llmprobe 0.6.6 contextScaling, three runs per rung. [Raw OFF/ON JSONs, logs, exact manifest and validation](https://github.com/nikolai-vysotskyi/mlx-serve/tree/4b9dcb6/research/reviewed_ladder_20260913). Peak MLX allocation was80.113/80.132 GB OFF/ON.

Caveats that matter: OFF decode drift -10.9%, ON +4.8%; the lower short decode rates cannot be cleanly attributed to these changes. Final standalone 64K OFF was1564 at68,651 tokens, drift+0.7%. Inspecting llmprobe showed that filler sizing uses preceding rungs, so that final standalone arm is a **coarse drift check, not a matched A-B-A**. Both complete ladders used the same rungs. At8K, tail coalescing produced S8230, above the specialization cap; the newly added QSA fallback log exposed that. Longer rungs use eligible chunks.

**2400 at65K is not reached.** This closes the missing-current-head measurement request; it does not claim the outstanding M4 greedy divergence or broad quality gate is resolved. The PR description is updated to the current defaults/code/evidence. Next research is the inclusive expert pipeline and normal cadence costs, without another full ladder for a weak component hypothesis.

nikolai-vysotskyi:
@ddalcu Two further concrete fixes are now pushed in **bf38063**:

1. `nextChunkEnd` absorbs up to 511 extra tokens. The old S<=8192 guard therefore disabled HC/GDN/QSA for an 8K prompt after coalescing. All three now accept up toS8703; the shared limit is checked against the actual chunker, and QSA reserves the enlarged planner. Real-model integration at 8296 prompt tokens logged HC/GDN/QSA engagement at S8295, including cold GDN, and retained exact passphrase recall. Subsequent 13515/15715-token requests passed too.
2. The newly generalized HC geometry exposed an actual rounding bug at HC=3: native `mean` rounds its reciprocal to BF16 before multiplication, whereas FP32 division differed. The production compiled-reference test failed at 0.001953125 before the fix and passes afterward for HC2/3/4/8, plus HC4/H2560/S8703. The HC4 reciprocal is unchanged, so this is **not an explanation of the M4 HC4 divergence**.

ReleaseFast 7/7, full Zig 2330 passed /154 skipped, expanded float64 QSA oracle and integration passed. [Validation artifacts](https://github.com/nikolai-vysotskyi/mlx-serve/tree/bd70e05/research/followups/pr408-tail-validation). New binary SHA256 `d62beaf96ffb72c377e9c8cb845ae20022a4e0cfc00667a567252936f47fb203`.

The full context ladder above remains explicitly a measurement of **d23d9df**, not of this new head. I did not rerun the entire ladder for these localized fixes. Paired QSA remains opt-in; M4 output isolation and broad quality validation remain outstanding.

nikolai-vysotskyi:
**Reproduction clarification (September14):**2347tok/s is the warmed, prefix-uncached HTTP profile on087c210, not a universal65K or llmprobe result. The same local M5 earlier recorded1761 in a different-head llmprobe ladder. [Frozen exact request, original binary/library fingerprints, settings and limits](https://github.com/nikolai-vysotskyi/mlx-serve/tree/f50ca9f/research/followups/repro-64947). The manual width pin and the incorrect MLX0.32.3 label are corrected below.

@ddalcu @cameronbergh @beamivalice **Final consolidated submission: approximately 2300 tok/s at65K, one remaining PR (#408).**

The stable integrated head is **3e1dc4d10cb31c1e0b97ee3c7e70d8ee40f6f556**, incorporating upstream0814cf3ce2; the merge conflicts are resolved. The measured snapshot is preserved separately. **2173.8 →2347.1 tok/s at64,947 tokens (+7.97%)** is the matched HTTP screen on087c210, with Automatic fans and documented passive cooldown. Later optional-MPP observation:2299.9. These are not a new llmprobe cell, not a>1.5× claim and not a benchmark of the upstream integration.

**Final integration validation:** **2444 tests passed /155 skipped, 9/9 build steps, exit0**. ReleaseFast7/7; binary SHA256`7dcdbb0a5516d81b3a14c7cb41fcc19c1de2045973f1577113845a7113487bdb`. The previous identical-source consolidation also retained its2333passed/154skipped and actual-model parity evidence. The new test window did not load the full model or rerun performance tests.

**Review requested:** @ddalcu please review the consolidated default-on HC/GDN portion and the explicit opt-in PLE/MoE/HC-upmix additions, including admission/pressure and layout-fallback requirements. @cameronbergh the remaining M4 action is one corrected-head OFF/HC-only/GDN-only isolation, then the agreed ladder if parity permits; the old HC4 greedy divergence is not claimed fixed. @beamivalice the requested old-core4K/8K/16K/64K/128K runs3 prefill/decode table and logs are below; final-combination acceptance is still a distinct missing cell. QSA stays approximate and opt-in. The PR remains draft for these actual readiness gates.

I am closing standalone#375 and the#366/#368 research trackers with links here and precise dispositions;#385/#388 and#365 are already closed/landed. The old#375 purge-pressure acceptance is explicitly unproven. There will be no further competing PR or unbounded local experiment loop for this research phase.

[Final integrated-head manifest, build and full-suite logs](https://github.com/nikolai-vysotskyi/mlx-serve/tree/2ab09e8/research/followups/final-prefill-20260913).

Below is the complete final report, including every measurement distinction, reviewer response, upstream credit, regression investigation, cooling result, source/reproduction pointer and rejected/unimplemented hypothesis.

---

# Final Qwen3.8 Flash-Next prefill investigation — M5 Max 128 GB

The local research phase is complete. **PR [#408](https://github.com/ddalcu/mlx-serve/pull/408) is the single consolidated review target.** The measured result accepted for this phase is approximately **2300 tok/s at 64,947 input tokens**. It is an HTTP screening result, not a claim of sustained serving throughput or a >1.5× improvement over current upstream.

## Results and provenance

All new model measurements use `ddalcu/Qwen3.8-Flash-Next-MLX-Serve-mixed-4-8bit`, M5 Max 128 GB, macOS 26.5, AC power, chunk8192, ctx131072, prefix-cache entries0, KV quantization off, MTP/PLD/drafter off. The request is public repository source plus a fixed passphrase, greedy seed1234, max_tokens32; the measured responses actually contain 10 output tokens. Every listed long request has cached_tokens=0 and recalls MAGNOLIA-7731. No model conversion, requantization, layer removal or changed routing selection was used.

| Evidence | Control / candidate | Prefill tok/s | Meaning |
|---|---|---:|---|
| Frozen 087c210, 13,515 tokens, warmed 0/7/7/0 bracket | Existing QSA+HC+GDN / add bounded PLE-ahead + MoE grouping + HC-upmix | 2182.1,2163.5 / 2335.6,2330.9 | +7.38% mean; control drift -0.85% |
| Same 087c210 binary/process, 64,947 tokens, one 0/7 pair | Same arms | **2173.8 / 2347.1** | **+7.97% observed**, decode50.0/51.1 |
| Later pre-wide32 snapshot, 64,947 tokens, reversed f/7 pair | Add optional MoE MPP / previous combination | 2284.7 / 2267.6 | Only +0.75%; starting conditions differ, no strong incremental MPP claim |
| Same pre-wide32 snapshot, fans Full blast, 64,947 tokens | Combination with optional MPP; no matched Automatic control | **2299.9** | Decode48.9; independent approximately 2300 observation, not a fan-speed A/B |
| Rejected wide32 experiment, 64,947 tokens | Chunk32768, wider guards, per-layer eval, padded HC tails | **2016.9** | Regression; excluded from final runtime |

The 087c210 executable SHA256 is `cc7f4574620557b263212e009c7e933783a3e0fac4a00ff778d708467de59d48`. The later pre-wide32 snapshot executable is `283bc95a286117d72fd7f7fb5398acb058ad530fb092171545a3880e8b41bf4e`. Consolidation commit b7d9033 copies this already-built/tested runtime source without changing it; a source manifest identifies the restored snapshot and hashes. **2347.1 belongs to 087c210 without MPP; 2299.9 belongs to the later snapshot with MPP.** Packaging does not retroactively reassign measurements to a new commit or binary.

Long-prompt SHA256: `d6927da48d558a30339f665d0536288a7e07394c5f0c4daa23eb7e4246d0037e`; corpus SHA256: `1b0d4f22b5f08c471a93caac3d5f4729be34c64a04e4c0e303a3fc44bcde36a0`. Peak MLX allocations for the matched long pair were78,214,759,746/78,223,208,014 bytes. Full-blast MPP peak was78,216,733,974 bytes. These are MLX allocator observations, not total process/OS memory or a proof of admission safety under pressure.

Correction September14: the measured runtime uses MLX0.32.2 source `1f8e74e3f12f31365464a6867c6579f0e9b29d85`; the previous0.32.3 label was incorrect. Direct runtime `mlx::core::version()` returns0.32.2, and dylib SHA256`c63811e0fe9afb5463f60137b6db913da0de73634cc736c8e06771a62b1db477` matches the archived records. Do not assume a checkout's submodule label alone identifies its installed dylib. Use isolated build caches per checkout.

### Earlier llmprobe acceptance-style evidence

On reviewed core-fusion head **d23d9df**, binary `580f495dce825202ea2da7976e10f6b1c4db39158827e7d63735f60a79f0d476`, llmprobe0.6.6 contextScaling, three runs per rung, fresh process per complete arm, ctx262144, same mixed pack:

| Target | Actual OFF / ON tokens | OFF / ON prefill tok/s | OFF / ON decode tok/s |
|---:|---:|---:|---:|
|4K|4239 /4235|1434 /1489|55.1 /50.0|
|8K|8235 /8231|1599 /1623|56.2 /50.5|
|16K|16298 /16311|1531 /1735|57.2 /50.8|
|64K|65841 /65778|1517 /1761|49.6 /50.9|
|128K|131086 /131133|1447 /1771|48.5 /49.1|

Observed improvements were+16.1% at64K and+22.4% at128K. OFF decode drift was-10.9%, ON+4.8%. A standalone final64K OFF gave1564 at68,651 actual tokens; llmprobe calibrates using preceding rungs, so this is a coarse drift check, not a matched A-B-A. The newer bf38063 tail/HC3 fixes and later combined additions have separate validation; this ladder is not relabelled as a measurement of them.

The new2347 HTTP cell and old1761 llmprobe cell use different prompts/protocols. Their ratio is **not an isolated code gain**. The user's earlier1868 upstreamfa76a4b cell likewise cannot serve as a causal denominator for this newer workload. Neither2400 nor>1.5× whole-model improvement has been established by the matched data. The earlier2509.4 short-context checkpoint was not a65K result; the old1879 long result preceded the HC inject correction.

## What the consolidated code contains

| Change | Mechanism and evidence | Default / scope |
|---|---|---|
| Paired QSA | Stage common selected K/V blocks for adjacent queries; component about29.5→14.19ms. Selected-key multiset preserved. FP32 probability high/low BF16 representation retains the tested accuracy bar, but reduction order changes. | Opt-in `MLX_SERVE_QSA_PAIR=1`, NAX; approximate, not bit-identical |
| HC prefill | Fuse pending write, normalization and stream mixing; retain native inject matmul after the original fused-inject rounding defect. Actual component4bit6.44→4.43ms;8bit6.58→4.73ms. | Default on, `MLX_SERVE_HC_PREFILL=0` rollback |
| GDN prefill | Fuse cold/warm prework and output normalization/gating, preserve BF16 rounding/table sigmoid. Compiled-production fixture parity. | Default on, `MLX_SERVE_GDN_PREFILL_FUSED=0` rollback |
| Planner/shape fixes | QSA planner71,909,376→19,611,648 bytes atS8192/B1; transient billed, KV support through1,048,576 with fallback log. HC geometry derived. Coalesced S<=8703 supported in core fusions. HC3 reciprocal rounding corrected. | Included in core path |
| Bounded PLE ahead | Request-owned CPU producer copies only selected100-byte quantized rows; two ready chunks plus one preparing,26.2MB raw-row cache plus metadata. MLX calls remain on inference thread. Compare all row IDs before consuming; join on cancellation. | Opt-in `MLX_SERVE_PLE_PACKED=1 MLX_SERVE_PLE_AHEAD=1`; fixed-boundary eligible text prefill; no whole-table GPU registration |
| MoE grouping/reduction | Counting/prefix/scatter replaces two general argsorts; fuse inverse-permutation, weighted reduction with native rounding. Native upstream already sorts experts: this does not fix an unsorted upstream path. | Opt-in `MLX_SERVE_MOE_PREFILL_GROUP=1`; B1,S2048..8192,E512,K10,H2560, supported affine4/BF16 geometry |
| HC upmix | Native MPP projection epilogue performs stream mixing without materializing the large up tensor. Compiled native-reference test and two live20,971,520-element comparisons exact. | Opt-in `MLX_SERVE_HC_UPMIX=1`; NAX, guarded HC4/H2560/K320/affine8, aligned rows; tail fallback |
| MoE MPP gate/up/SwiGLU | Expert-aligned64-row tiles, gate/up weight reuse and exact-table SwiGLU in epilogue; native down/shared-expert/reduction remain. Avoids419,430,400 logical intermediate bytes perS8192 layer. | Additional opt-in `MLX_SERVE_MOE_PREFILL_MPP=1`; implies grouping. Included for reproducibility, **not needed for2347 and not a demonstrated large model gain** |

MoE MPP component on real-routing fixtures with synthetic activations/weights: grouped-native23.34–23.41→21.284ms, another layer23.24–23.32→21.420ms. Live replay using actual model weights and inputs compared OFF/ON/ON/OFF: native30.288/30.236 vs group+MPP24.855/24.698ms at the first sampled layer; native29.609/30.247 vs24.943/24.886 at the fourth. Every20,971,520-element output comparison was bit-exact. Replay synchronizes the graph; its request tok/s is **diagnostic**, not a normal-serving gain.

New experimental paths remain opt-in because their general quality, architecture fallback, admission and pressure behavior are not established. This does not represent them as proven universal default-on wins. The core HC/GDN polarity requested in review remains implemented. Optional research replay/capture/sequence controls are disabled in ordinary serving; do not enable them for production measurements except the documented arm sequence used for this screening.

## Why a faster component did not always make the model faster

1. **A real integration regression:** registering the whole~32GB PLE table with Metal gave a~36× warm standalone gather ratio but actual15,715-token model prefill fell1987.7→1075.6. Page residency/registration and graph lifetime costs were outside the component timing. That integration was removed; bounded selected-row staging replaced it.
2. **Different numerical implementations:** the old HC fused inject was faster but did not match native dense BF16 reduction. Correcting it costs time. Old1879 cannot be used as evidence for the correction. HC3 reciprocal was another real rounding bug; HC4 is unchanged by that fix, so it does not explain the reported M4 HC4 divergence.
3. **Machine-state drift:** uncooled identical-binary sequences showed roughly10% control drift, large enough to bury modest component gains. Passive matched starting conditions reduced the short control bracket drift to0.85%. Temperature alone did not explain clock changes.
4. **A separate resource collision:** later, an unrelated local training process caused model-readiness HTTP503 (only16.11GB available against~77.1GB required). The user stopped training; the retry loaded and produced2299.9. The failed load is not a speed measurement and does not prove the earlier runs were all contaminated.
5. **Larger graphs changed the inclusive costs:**32K batches helped a fixed MoE component per token, yet the full model with wider guards, per-layer evaluation and HC padded tails only reached2016.9, with84.04GB MLX peak. The experiment was reverted. The per-component ratio omitted costs that moved elsewhere in the graph.

### Cooling result, without turning it into a code speedup

The2347 pair used Automatic fans. Before each request, outside its timer, the driver required three fresh samples with GPU<=50C, GPU power<=5W and both fans<=1600RPM. It aborts after180s. The long arms started49.7/47.7C, fan pairs1346/1463 and1460/1584RPM; request-interior approximate mean frequency1482.0/1490.8MHz, peak96.6C in both. Idle gaps are excluded; including them would lower end-to-end workload throughput substantially.

A later50C MPP comparison aborted after the old-combination long arm2335.6 because the next cooldown could not reach50C in180s. No missing candidate result is invented. The reversed55C pair2284.7/2267.6 started54.9/47.5C and had mean frequencies1442.6/1466.0MHz. Retain this mismatch when judging the small delta.

After training stopped, temporary Full blast fans around5354/5770RPM yielded2299.9 at65K, peak77.0C, approximate mean1467.9MHz and67.5W GPU power. No same-setting Automatic control followed, so there is no causal fan-speed ratio. A140W adapter was connected. All frequency/temperature observations use1-second telemetry windows inside HTTP request bounds, not kernel timestamps; do not normalize tok/s by MHz. **Fans were restored to Automatic.** No clock, power or permanent cooling changes are part of the PR, and Full blast is not required to reproduce the earlier2347 result.

## Validation and remaining review gates

The measured pre-wide32 runtime snapshot was built ReleaseFast7/7 and passed the full Zig suite **2333 passed /154 skipped** on this M5 before packaging. The source manifest confirms136 tracked runtime/build inputs at consolidation commit b7d9033 match the tested research snapshot after excluding the failedwide32 changes. The binary and prior test output are identified separately; no new GPU test was run during finalization because the owner reassigned GPU/RAM. No Swift/app code changed.

Coverage includes compiled HC/GDN references, generic HC2/3/4/8 and HC3 red/green, tail integration at8296/13515/15715 tokens, QSA selected-key/float64 oracle through1M KV, PLE row duplicates/lifetime/cancellation/history mismatch, grouping/inverse reduction, padded MoE input and exact tile coverage, HC-upmix compiled parity, live actual-weight/activation comparisons and uncached HTTP recall. Finite fixtures and a passphrase are not a broad model-quality evaluation.

Still open in **#408**, not silently marked completed:

- Corrected-head **M4 HC-only/GDN-only** deterministic-output isolation and requested fresh-process M4 ladder. Original M4 run on092ce2e had-3.9% at16K,+3.0% at64K and greedy divergence; QSA did not engage there because NAX was unavailable.
- QSA broad model-quality acceptance and cooperative-layout startup validation/fallback. New HC-upmix and MoE MPP also need a layout probe/fallback; a NAX capability guard alone is insufficient, and their current unsupported-layout diagnostic can yield NaN. They stay opt-in/draft.
- Fresh final-combination llmprobe ladder and rollout/admission review for new PLE/MoE buffers. The old ladder proves the old named core head; the new HTTP screen is supporting research evidence, not a replacement.
- Productionizing the bounded PLE/MoE paths, including pressure/cancellation/architecture review and removal or relocation of research diagnostics before a general rollout. No claim that pressure benefits or every non-M5 case are validated.
- Finalization found a merge conflict and then integrated upstream0814cf3ce2. The adapter retains upstream per-slot speculative PLE capture and deferred gather, and limits the experimental packed path to non-batched slots. See the final PR comment for the separate compile/test status of this integration. No model speed is assigned to it: the measured base remainsfa76a4b, and the new main was not silently substituted into the historical comparisons.

PR remains draft for these concrete reasons. Local research is finalized; merge/readiness and cross-machine validation are separate outstanding work, assigned explicitly in the final escalation.

### Upstream work already credited and retained

The baseline already contains upstream QSA NAX/block-select/score-sheet work, including our#385 contribution integrated by beamivalice through#388 (merge680e5a56d2ca6926785b404efed06d1d85b563f3). It also contains the owned conv-state tail and cadence/admission correction recorded from#366: upstream reported about2.5GB lower peak at4K chunks and neutral prefill, plus correction of the already-fixed-size QSA raw-key ring bill. These are valuable earlier upstream improvements, not new speedups attributable to this final PR. The final source integration also retains newer upstream MTP/batching and per-request Flash-Next state changes.

## Disposition of every related issue/PR from our side

| Thread | Final disposition |
|---|---|
|[#408](https://github.com/ddalcu/mlx-serve/pull/408)|**Only open consolidated PR**: code, reproduction and final report; draft pending gates above. Reviewers asked for stable-head M4/quality review, not another investigation thread.|
|[#375](https://github.com/ddalcu/mlx-serve/pull/375)|Close as superseded for our ongoing work. A gather path/test seam was already cherry-picked upstream according to13scoobie; it is not being reverted or claimed newly landed by#408. The remaining requested automatic-thread policy, accurate engagement label and purge pressure acceptance were **not completed**.|
|[#368](https://github.com/ddalcu/mlx-serve/issues/368)|Close the tracking issue: original prefetch gate fixed in862bddf; pressure follow-up from#375 remains unproven and is archived here. Closing means research consolidated, not pressure performance proven.|
|[#366](https://github.com/ddalcu/mlx-serve/issues/366)|Close the research ledger after linking this report/#408. Original apparent MoE throughput gap was a chunk-size comparison, not an unresolved kernel defect; upstream already sorts wide MoE. New findings and all rejected ideas remain linked.|
|[#385](https://github.com/ddalcu/mlx-serve/pull/385)|Already closed, integrated via[#388](https://github.com/ddalcu/mlx-serve/pull/388) by beamivalice. Keep closed; that upstream NAX baseline is distinct from paired QSA in#408.|
|[#365](https://github.com/ddalcu/mlx-serve/issues/365)|Already closed/fixed in26.9.2: Anthropic system-message preservation. No new action or prefill speed attribution.|

The#375 requested pressure protocol remains recorded precisely: mixed pack, `MLX_SERVE_NGRAM_WARM=0`, purge before each65K request, unique salt,max_tokens1, three alternating runs per arm/boot, plus no-purge control. Acceptance requested>=3% pressure gain with lower `ple gather S=4096` time and neutral warm control. It was not run; do not infer it from the bounded-ahead results.

## Rejected and unfinished hypotheses

These are archived to avoid repeating weeks of low-yield experiments. Component times are from each cited run's own controls, not composable model ratios.

| Hypothesis | Finding | Status / next condition |
|---|---|---|
| Whole-table GPU PLE / managed mapping | Standalone large win, actual model regression; managed/storage restrictions and page-ins matter | Rejected; use bounded selected rows |
| One resident MoE gate/up/down pass | About63ms vs25ms native; register/shared-memory schedule cost | Rejected |
| Larger register tile / BM192 gate-up | Earlier28vs25ms; newer MPP BM19221.945 vs BM64~21.452 | Rejected for current shape |
| Direct-input expert kernel | Inclusive improvement insufficient versus counted/sorted native; wider variant remained slower | Archived, no large model claim |
| Exact dense expert cache | At160rows/expert9.18ms cached vs~6.04native; inclusive13.66. At1280rows cached38.15 vs41.02–41.24, before wider model costs | Reject large-weight expansion as current lever |
| INT4 codebook / shuffle dequant |8.51vs7.29–7.38ms, exact | Rejected |
| Expert-aligned staging / virtual gate-up interleave | Roughly10–16% component signal, overlap with grouping benefit | Retained only where final MPP inclusive evidence supports it; no extra multiplied gain |
| New MPP down projection | BN128 whole chain22.001 vs gate-up-only21.694/22.677; BN25629.861 | Rejected; keep native down |
| Hoist MPP weight-bank address selection |21.772 vs22.237/22.021ms, exact | Small component gain only; not integrated |
| Bigger MoE batch |8K21.465/21.970ms;16K40.328;32K75.711, exact | Per-token component improves; whole-model32K experiment regressed |
| Wider whole-model prefill |32K/QSA precision/tail checks passed;65K prefill2016.9,peak84.04GB | Reverted; patch archived, not in final runtime |
| Increase command-buffer memory budget4096MB | No material improvement | Rejected |
| Cache exactly expanded dense/HC weights | Dense projections already use upstream DQGEMM at large M; HC4.159vs3.950/4.031ms even excluding expansion | Rejected |
| QSA Q-fragment reload / packed blocks |16.951/21.007ms vs retained~14.19 | Rejected |
| QSA separate QK/PV groups / pipeline |19.04/19.69ms (addressable-union variant72.85) vs~14.19 | Rejected; reduced spilling not a measured hardware counter |
| QSA full D256 in one SIMD group |62.060ms vs13.753paired | Rejected |
| Four-query QSA packing | K/V tile count-30%, active16-head tiles+3%; bucketing fragmented | No credible inclusive large gain; no full-model run |
| One-call FP32×BF16 PV |12.413vs13.70–13.85ms, but RMSE1.93e-5 vs1.16e-6 and larger max error | Rejected quality tradeoff |
| Strict FP32 PV cooperative layout | Incorrect max difference0.204834, slower | Rejected implementation; new layout derivation required |
| Coarsened online QSA max |16.131vs13.81ms | Rejected |
| Split QK→softmax→PV | About0.83GB FP32 score scratch on S4096 fixture, several GB extra traffic | Unimplemented; first demonstrate inclusive component win and bill scratch |
| Shape-specific tail JIT | Pinned MLX passes shapes/strides as runtime buffers; actual dimensions are not kernel-name keys for same type/template class | Proposed explanation disproved by source |
| Disabled MTP running during prefill | Generation/head history gated by mtp_active; some head/rerank preparation still at load | Not a timed-prefill lever; no change |
| CPU-only parallelism for1.5× | Host tens of ms vs GPU seconds on studied normal path | Insufficient budget; pressure PLE is separate unproven case |
| GDN core alone | Roughly0.288s/chunk; not the whole GDN block budget | Need full-stage cost reduction, not recurrence-only ratio |
| More cooling guarantees speed | Lower observed peak77C, clock still varied1095–1620MHz | Thermal state matters; no proven clock root cause or permanent fan requirement |

Other future work worth reconsidering only with a new inclusive cost model: shorten QSA register lifetimes without score-spill traffic; fuse larger exact MoE segments without register overcommit; reduce full GDN projection/materialization costs; bounded PLE pressure behavior. No evidence supports promising1.5× from any one of these remaining sketches. The current research phase does not schedule more tests.

## Reproduction and evidence map

Use the final PR runtime or explicitly checkout087c210 to reproduce the published2347 candidate. Build with the stated Zig/MLX/SDK provenance, ReleaseFast, private per-checkout caches. Do not share binaries across worktrees without hashing them. Use an exclusive GPU window; do not run while another training/UT session uses GPU/RAM. The driver owns and terminates only its own server/monitor.

```sh
PLE_BENCH_TAG=cooled-short python3 research/followups/bench-prefill-combined.py \
  --sequence 070770 --cooldown-c 50 --monitor /path/to/macmon \
  --out /tmp/qwen-cooled-short --lock /path/to/gpu_lock.sh
PLE_BENCH_LONG=1 PLE_BENCH_TAG=cooled-long \
  python3 research/followups/bench-prefill-combined.py --sequence 0707 \
  --warmup-short 2 --cooldown-c 50 --monitor /path/to/macmon \
  --out /tmp/qwen-cooled-long --lock /path/to/gpu_lock.sh
```

Run from repository root. Sequence0 keeps QSA+HC+GDN;7 adds group+PLEahead+HCupmix;f adds optional MoEMPP. The first two long-mode requests are13.5K warmups. Same-process arm selection is a research control, not a user-facing serving feature. The current driver supportsf; the original087c210 driver did not yet have all cooldown options, so use the published final driver with the explicitly selected binary/source corpus.

For manual serving of the2347 feature combination, set `MLX_SERVE_PREFILL_CHUNK=8192 MLX_SERVE_QSA_PAIR=1 MLX_SERVE_HC_PREFILL=1 MLX_SERVE_GDN_PREFILL_FUSED=1 MLX_SERVE_PLE_PACKED=1 MLX_SERVE_PLE_AHEAD=1 MLX_SERVE_MOE_PREFILL_GROUP=1 MLX_SERVE_HC_UPMIX=1 MLX_SERVE_MOE_PREFILL_MPP=0`, use `--prefill-chunk 8192 --ctx-size 131072 --prefix-cache-entries 0 --kv-quant off --no-mtp --no-pld --no-drafter`, and the exact mixed-pack model. Leave all replay/capture/sequence env unset for ordinary serving. Verify engagement, actual token count, cached_tokens, prompt hash and binary hash from logs rather than assuming launch flags prove a path ran.

For the maintainer's acceptance run use `tests/bench_qwen4_prefill.py`/`tests/bench.sh` and the full named OFF/ON llmprobe ladder; its existing core-arm driver must explicitly configure the new levers before treating it as a combined-bundle comparison. Save both prefill and decode, model/tokenizer and binary hashes, launch env/flags, engagement, real token counts, warmup policy, AC/thermal state and allocator peaks. A failure to meet cooldown or load readiness is an incomplete arm, not zero tok/s. Do not retry heavy tests for a weak component hypothesis.

- [Matched cooled requests and reduced telemetry](https://github.com/nikolai-vysotskyi/mlx-serve/tree/2ab09e8/research/followups/cooled-prefill-20260913), [method report](https://github.com/nikolai-vysotskyi/mlx-serve/blob/2ab09e8/research/followups/20260913-cooled-prefill.md).
- [Final source manifest, full-suite output, live MPP and later/negative full-model artifacts](https://github.com/nikolai-vysotskyi/mlx-serve/tree/2ab09e8/research/followups/final-prefill-20260913).
- [Earlier full llmprobe ladder and engagement](https://github.com/nikolai-vysotskyi/mlx-serve/tree/2ab09e8/research/reviewed_ladder_20260913).
- [HC native inject red/green](https://github.com/nikolai-vysotskyi/mlx-serve/blob/2ab09e8/research/followups/20260912-hc-native-inject.md), [tail/HC3 validation](https://github.com/nikolai-vysotskyi/mlx-serve/tree/2ab09e8/research/followups/pr408-tail-validation), [HC-upmix actual parity](https://github.com/nikolai-vysotskyi/mlx-serve/blob/2ab09e8/research/followups/20260913-hc-upmix-fusion.md).
- [PLE residency diagnosis](https://github.com/nikolai-vysotskyi/mlx-serve/blob/2ab09e8/research/followups/20260912-ple-residency-investigation.md), [bounded ahead and grouping](https://github.com/nikolai-vysotskyi/mlx-serve/blob/2ab09e8/research/followups/20260912-prefill-ahead-and-grouping.md), [live MoE/drift](https://github.com/nikolai-vysotskyi/mlx-serve/blob/2ab09e8/research/followups/20260913-live-moe-and-thermal-drift.md).
- [Earlier component hypothesis ledger](https://github.com/nikolai-vysotskyi/mlx-serve/blob/2ab09e8/research/followups/README.md), [larger lever analysis](https://github.com/nikolai-vysotskyi/mlx-serve/blob/2ab09e8/research/followups/20260912-large-levers.md), [next-lever budget](https://github.com/nikolai-vysotskyi/mlx-serve/blob/2ab09e8/research/followups/20260913-next-levers.md).

Only source, aggregate measurements and the public-source recall workload are published. Model-derived raw routing arrays/weights/activations remain local. A workspace GPU-lock helper was also corrected: status is read-only and EPERM must not be mistaken for a dead PID; that local orchestration fix is not an mlx-serve runtime speedup.

nikolai-vysotskyi:
@ddalcu Fixed in **9644a6bd21a5c8bbfd22f920dc8cb54ceee72d9e**. The failed runner reported `[nax-probe] arch=air64_v27 macos=26.6.2 available=true`: the parser consumed compiler-target version27 as a GPU generation and enabled unsupported NAX construction. The shared runtime guard now validates the complete Apple GPU family identifier.

**Positive coverage:** supported desktop17+/phone18+, case-insensitive variants, verify-QMM/QSA dispatch, and the existing real M5 NAX parity/probe tests remain enabled.
**Negative coverage:** the exact `air64_v27` failure, other vendors, old desktop/phone families, malformed prefixes/suffixes, missing/oversized generation fields and embeddedNUL. The actual stock-gather test asserts bit-exact fallback output and no increase in NAX-kernel builds when hardware is unsupported.

Validation on M5 Max128GB/macOS26.5:

| Run | Engine passed | Skipped | Observed NAX status |
|---|---:|---:|---|
| Native full suite |2444|155|`applegpu_g17s … available=true`|
| Full `MLX_METAL_GPU_ARCH=air64_v27` rehearsal, separate cache |2430|169|`air64_v27 … available=false`|

The latter also ran20 guest-helper tests successfully (2450 total passes). It rehearses the architecture case on M5; it is not a claim to reproduce the CI machine/SDK exactly. The pure regression failed before the fix with **expected0/found27** and passes after it. ReleaseFast7/7; built-executable `--version` reports NAX on for the actual M5 and off under the CI architecture override. No CI checks or accuracy tolerances were weakened.

[Red/green logs, both full suites, build, CLI checks and manifest](https://github.com/nikolai-vysotskyi/mlx-serve/tree/9bf1928/research/followups/ci-nax-20260913). [Fresh CI run34765264300](https://github.com/ddalcu/mlx-serve/actions/runs/34765264300) **passed completely**, including the Zig suite, Swift app build and Swift tests. No full-model benchmark was rerun, and this availability fix does not close the separately documented layout/quality/M4 acceptance gates.

beamivalice:
Re-measured this PR's head `9644a6b` today on M5 Max 128 GB, Qwen3.8-Flash-Next mixed-4/8-bit, against a control built from current main plus two local MTP commits that do not touch prefill. Same box, same session, PR first then control; llmprobe 0.6.7 `--bench-only` with the planner warmed by three requests first, then cold prefills with the prefix cache off (`--prefix-cache-entries 0 --prefix-cache-disk off`), fans at max.

| Cell | PR 408 | control | gain |
|---|---:|---:|---:|
| llmprobe prefill, 2042-token prompt | 1791 (1751–1812) | 1694 (1664–1719) | +5.7% |
| llmprobe context ~16.3k, prefill | 1858 | 1761 | +5.5% |
| cold 64k prefill, two runs | 1830 / 1869 | 1747 / 1752 | +5.7% |
| cold 16k prefill | 1850 | 1753 | +5.5% |

The gain holds at every length and is close to the +8% in the description. The absolute rates on this box today are 1.75–1.87k tok/s at 64k uncached rather than the 2.17–2.35k recorded earlier, on both arms, so that difference is session state, not the PR. Decode on the PR's base is lower than the control's (86.6 vs 95.8 tok/s) only because the base predates the single-stream verify kernels the control carries.

nikolai-vysotskyi:
@beamivalice This is useful independent support for keeping the prefill work moving: your 64K means are **1849.5 vs1749.5 tok/s (+5.72%)**, with a positive result in both runs. The lower absolute number does not erase that within-session gain. @ddalcu please include this exact-head external result in the review evidence.

I found two concrete configuration-dependent paths to check before attributing the absolute gap entirely to session/thermal state:

1. **The CLI 8192 argument does not by itself pin 8192 here.** In 9644a6b, `effectivePrefillChunk` passes it through the default hd256 MoE cap of 4096; only `MLX_SERVE_PREFILL_CHUNK=8192` returns before that cap. Both published benchmark drivers set the environment pin. My manual command in the documentation omitted it; I have corrected that oversight in [92be945](https://github.com/ddalcu/mlx-serve/commit/92be945). Runtime sources in that documentation-only commit are identical to the measured/CI-green 9644a6b. [Policy](https://github.com/nikolai-vysotskyi/mlx-serve/blob/9644a6b/src/generate.zig#L151-L207). The existing pure cap test passes; this finding does not require another model run.
2. **MTP currently switches off PLE-ahead, even if PLE_AHEAD=1.** The producer eligibility guard explicitly includes `mtp_active`. MTP also selects full-hidden capture and `appendHistory` during prefill. Our2347 HTTP screen used MTP/PLD/drafter off. The 86.6/95.8 decode numbers make the exact speculative mode especially relevant, although they alone do not prove which mode ran. [Ahead guard](https://github.com/nikolai-vysotskyi/mlx-serve/blob/9644a6b/src/generate.zig#L2364), [capture/history work](https://github.com/nikolai-vysotskyi/mlx-serve/blob/9644a6b/src/generate.zig#L2455-L2515). Decode-only control patches and MTP-on versus MTP-off serving are different questions.

Our2347 HTTP screen uses the frozen source-code recall prompt. Your report explicitly names llmprobe for the short and~16.3K cells; the exact payload/protocol for the separate cold64K cell was not attached. We need its hash/config before calling those profiles equivalent. Prompt content can change expert routing, sparse-QSA union work and PLE row reuse; a thermal/session attribution is not established by matching nominal lengths.

One denominator correction: our +7.97% was **core QSA+HC+GDN vs core plus PLE-ahead/grouping/HC-upmix**, on one 087c210 binary. Yours is **PR 9644a6b vs another control tree**. Both are positive observations; matching percentages do not prove that the same feature set or incremental gain was measured. The 2,042-token cell also cannot establish paired sparse-QSA engagement by itself.

Could you attach the **existing** launch argv/relevant `MLX_SERVE_*` and `QWEN4_*` settings, exact control/two-MTP commit SHAs, checkpoint revision/config fingerprint if retained, llmprobe JSON and these log lines: `[prefill-trace]` (actual chunk widths), `[spec-stats]`/MTP history-window setting, plus QSA-pair/HC/GDN/PLE-ahead/group/upmix engagement? No new long benchmark is needed to resolve those metadata questions.

The next code lever I am investigating is retaining the bounded PLE producer with native Qwen4 MTP: the trunk capture wrapper preserves its context, while the MTP head constructs a separate context, which gives a plausible path to isolation. That is a source-backed proposal, not a validated compatibility change or new speed claim. We should keep the independently positive prefill result, repair reproduction, and target the exact path that is currently being declined. The existing M4/quality/layout gates remain distinct.

**Integration update:** #408 now includes upstream1075630 at **e4691282a37dd67456e5f2bf9440f459b165017d**; GitHub reports it mergeable again. I preserved upstream grouped PLE synchronization and joined-verifier MoE paths, with our wide-prefill grouping outside that verifier path. ReleaseFast7/7 and2512 tests passed/169 skipped onM5. The native-MTP-ahead proposal is still unintegrated, and no old throughput cell is reassigned to this new head. [Research, proposal and integration evidence](https://github.com/nikolai-vysotskyi/mlx-serve/blob/eb59279/research/followups/20260914-review-and-compatibility.md). [CI for e469128](https://github.com/ddalcu/mlx-serve/actions/runs/34829114022) passed completely.

**Reproduction package and provenance correction:** https://github.com/nikolai-vysotskyi/mlx-serve/tree/f50ca9f/research/followups/repro-64947 freezes the exact64,947-token request and source corpus. My earlier MLX0.32.3 label was incorrect: the exported runtime reports0.32.2 and its dylib SHA matches our archived records. The new harness supports a frozen corpus and records library/model-metadata hashes and raw power settings. The original2347 executable has also been recovered locally byte-for-byte. No replay has been run yet while the GPU/memory window is being coordinated.

nikolai-vysotskyi:
@beamivalice @ddalcu I audited our2300-vs1800 discrepancy and made the original profile concrete: [frozen64,947-token request, warmup, corpus and manifest](https://github.com/nikolai-vysotskyi/mlx-serve/tree/f50ca9f/research/followups/repro-64947). The exact original executable was recovered locally with matching SHA`cc7f457…59d48`, so a repeat can use the original binary.

Two clarifications from our side: **2347.1 was a warmed HTTP recall screen on087c210**, using a repeated source corpus, the8192 environment pin and MTP off. It is not a measured llmprobe rate for the current PR. On this same physical M5, the earlier d23d9df llmprobe run had1761 reported and1752.6–1803.3 in its raw server logs at~65K. Those heads/protocols differ, so this is not a causal prompt A/B, but it shows that the two headline rates do not yet describe equivalent work. The2347 observation itself checks out:27671.095ms of prefill,27.887839s total HTTP wall time,64,947 input tokens and cached_tokens0; even the wall-time rate is2328.9tok/s.

**I also corrected an erroneous dependency label:** the linked library is MLX0.32.2, not0.32.3. Its exported version() and SHA`c63811e0fe9afb5463f60137b6db913da0de73634cc736c8e06771a62b1db477` agree with the archived library records; the source pin is1f8e74e3. This corrects our documentation, not the measured timing.

Please compare the frozen profile with the retained flags/trace from your run. A same-payload replay is the useful next check; another entire ladder is unnecessary. The width/MTP/engagement metadata previously requested still matters. I am coordinating the local full-model replay window and have not assigned any new speed result to it. Your independent+5.72% result remains valuable; the absolute gap remains unattributed until conditions are aligned.

ddalcu:
Merged this with current main locally (one conflict in transformer.zig against #427's paired kernel), built ReleaseFast, ran the full zig suite, green. Then A/B'd against main on an M4 Max 128 GB, Flash-Next mixed-4/8, prefix cache off, llmprobe --bench-only, one boot per arm:

| prefill tok/s | 2k | 16k | 32k |
|---|---:|---:|---:|
| main | 747 | 723 | 728 |
| this PR | 760 | 744 | 747 |

Decode unchanged. Only two things engaged on this box: `[hc-prefill]` and `[gdn-prefill]`. That is the +2-3% here and the +5.7% beamivalice saw on M5. Nothing else in this PR ran.

That is the problem. The PR is 2,500 lines and the part that works is maybe 400 of them. The other five paths (QSA pair, PLE ahead, PLE packed, MoE group, HC upmix, MoE MPP) are all `=1` levers, default off, no number anywhere that says any of them pays, and QSA pair is lossy and never runs off NAX. Default-off is dead code. I am not merging dead code and I am not merging research/, prefill_experiment.zig, qsa_pair_precision.cpp or a 400 line doc into the tree.

Also hc_prefill.zig still hardcodes 2560/10240/HC=4. That is one pack, not an arch. Derive it from config or the next pack silently falls back.

Close this and open a new PR off a fresh main with just HC + GDN prefill fusion, default on, `=0` kill switches, literals derived, one CLAUDE.md line, one gotcha entry, one CHANGELOG line. Nothing else. If one of the other paths ever shows a win on its own, that is its own PR with its own number.

PR text: say what it does, what you measured, on what. No manifests, no thread dispositions, no hypothesis logs.

nikolai-vysotskyi:
@ddalcu @beamivalice Implemented the requested split in **#438**, on fresh main `d7452ad`: **7 files, +406/-37**, including tests. Only HC/GDN prefill fusion remains, default on with `=0` kill switches. No QSA-pair, PLE, MoE-group/MPP, HC-upmix, experiment dispatcher, research files or standalone report. HC dimensions derive from config/weights; the fixed 2560/10240/HC=4 upmix path is gone. Compiled-reference HC tests now cover counts 1–8 and hidden widths 128/1536/2560/4096, alongside GDN cold-history, batch/tail and negative cases.

**2347.1 tok/s is a recorded M5 Max 128 GB prefill result and a reproduction target.** The server processed 64,947 uncached input tokens in 27.671 seconds; client wall time was 27.888 seconds. A materially lower rate in a comparable M5 run is an unresolved reproducibility gap that needs investigation. Calling it a special benchmark does not explain the gap.

The recorded `087c210` configuration enabled QSA-pair, HC/GDN, PLE-ahead/packed, MoE-group and HC-upmix. Its paired comparison was 2173.8→2347.1 (+8%), with QSA+HC+GDN already active in both arms. Your default-path A/B and the minimal #438 use HC/GDN only. These are concrete configuration differences to isolate, not a demonstrated explanation of the entire absolute-rate gap. The original binary and request are retained for replay; we need to reproduce the result and identify what is required to deliver it in normal serving.

The replacement tree compiled ReleaseFast and passed the full Zig suite locally on M5 Max 128 GB / macOS 26.5: **2558 passed, 171 skipped, 0 failed**. #438 remains draft pending a fresh-head full-model A/B. Closing this oversized PR in favor of that focused replacement; experimental work is preserved in the existing research branch.

nikolai-vysotskyi:
@ddalcu @beamivalice **Fresh full-model reproduction on M5 Max 128 GB: 2323.423 tok/s at 64,947 uncached input tokens**, within **1.01%** of the earlier 2347.106. The 2300+ level reproduced after freeing memory; the original retained executable, libmlx and input hashes all match.

| Same-process arm | Prefill tok/s | Prefill seconds | Client wall seconds |
|---|---:|---:|---:|
| QSA-pair + HC + GDN |2163.887|30.014049|30.260480|
| Add PLE-ahead/packed + MoE-group + HC-upmix |2323.423|27.953158|28.165814|

**+7.37%** for the added combination. Both requests returned 10 output tokens, the expected recall answer and explicitly zero cached tokens. Counting the entire HTTP wall still gives **2305.9 input tok/s** for the faster arm. One new A/B pair; two 13.5K warmups excluded.

Configuration: retained source `087c210`, mixed-4/8 pack, explicit 8192 chunk pin, context 131072, KV quantization off, MTP/PLD/drafter off, MoE MPP off. All named paths logged engagement. Same passive starting-condition gate as before: GPU ≤50°C, ≤5 W, fans ≤1600 RPM; no fan/power settings changed. Cooldown is outside the request timer.

[Raw responses, matching hashes, engagement and starting telemetry](https://github.com/nikolai-vysotskyi/mlx-serve/blob/ce3371d/research/followups/replay-20260915/result.json) · [reproduction summary](https://github.com/nikolai-vysotskyi/mlx-serve/blob/ce3371d/research/followups/replay-20260915/README.md).

The lower ~1800 M5 rates therefore remain a reproducibility gap to investigate. This gives us a working 2300+ configuration to compare against: the next step is to isolate enabled paths and launch/input differences and establish what carries into normal serving. #438 remains the HC/GDN-only proposal; this replay is of the retained combined implementation, and it does not assign a separate gain to each additional path.

Issue #133, opened 2026-08-07
Process 'mlx-serve' not closed on quit


freppair:
When quitting MLX Core, either via the Quit menu button or ⌘+Q, the process 'mlx-serve' remains running, with a model still loaded/consuming large amounts of RAM.

Pressing the power button in the menubar pull-down menu quits the main GUI and closes mlx-serve.

Version 26.8.2 (26.8.2)
macOS 26.6.1

ddalcu:
Thanks for the report, fixed in 26.8.3 ... I always close it via the power button for some reason 😄

PR #411, opened 2026-09-12
feat(app): auto-restart the server after an unexpected crash

### LINKED ISSUE #265: [Feature Request] MLX Core: Auto-restart the server after an unexpected crash (with backoff)
## Summary

When `mlx-serve` dies unexpectedly, MLX Core shows a modal alert and stays stopped. There is no recovery: the server remains down until a human notices the modal and restarts it manually. On a Mac Studio serving the LAN (coding agents, other Macs), a crash takes down every client until someone is at the keyboard.

## What happens today

`app/Sources/MLXServe/Services/ServerManager.swift`:

- `handleTermination(exitCode:)` — when the process exits while `.running`, it sets `.error("Exited unexpectedly")` and calls `presentCrashAlert(...)`.
- `presentCrashAlert` — a modal `NSAlert.runModal()` with "Copy Log" / "OK". Dismissing it leaves the server stopped.
- Intentional stops are already distinguished: `stop()` nils `proc.terminationHandler` before `terminate()`, so only an unexpected exit reaches the alert path.

There is no restart anywhere in the app; the only "restart" concept is the manual "Restart Now" banner for settings changes.

The building blocks for recovery already exist:

- `lastLaunchedOptions` — the exact options from the last `start()`.
- `launch(args:options:)` — re-runnable, and already handles port conflicts via `killOrphanedServers(on:)`.
- The `/health` polling in `startHealthPolling()` can confirm the restarted process is ready.

## Request

On unexpected exit while the server was `.running`, automatically relaunch with the last launch options and confirm readiness via `/health`.

Suggested behavior (happy to discuss):

1. **Backoff + cap** — e.g. 1s, 2s, 4s … capped at 30s, max 3–5 attempts within a ~5-minute window, then fall back to today's modal. This matters because a load-OOM or a system memory-pressure SIGKILL will recur immediately; unbounded retries would spin forever.
2. **Non-blocking notification** — a menu-bar item / macOS notification while auto-restart is in progress, so the modal is only the final state when recovery is given up.
3. **Opt out of the modal entirely (headless / silent mode)** — a setting to suppress the crash dialog completely: log the crash (exit code + stderr tail) to the existing on-disk log and proceed straight to auto-restart. The stderr is already captured at `~/.mlx-serve/logs/mlx-serve-<port>.log`, so this is nearly free. It matters on unattended Macs: while the alert is up, the app is parked in `runModal()`'s nested event loop and the rest of the app is frozen until someone clicks OK.
4. **Reset on success** — clear the crash counter once the server has been up for a while (e.g. 5 min) or on a manual start.

A single "crash handling" setting with three modes (`ask` = today, `auto-restart`, `auto-restart silent`) or two independent toggles — whichever fits the Settings UI better.

Notes:

- A restart re-runs `--model`, so the model must be reloaded — tens of GB and some seconds. That is the real cost of recovery. If #214 lands (separating server start from model load), recovery becomes much cheaper: the server is back in milliseconds and the model hot-loads on the first request.
- CLI users running `mlx-serve` directly can already put it under launchd or any other supervisor; this request is specifically about the app, which owns the child process and currently has no recovery at all.
- `handleTermination` does not distinguish exit codes today (a clean exit 0 while `.running` also shows "Exited unexpectedly"); worth distinguishing while in there.

## Environment

- macOS 26.6.2, Mac Studio (M4 Max), mlx-serve 26.8.9 (MLX 0.32.0)
- MLX Core serving the LAN; on crash, all LAN clients are down until a manual restart.

Happy to test a branch.

t2tx:
## Summary

Closes #265. Adds opt-in crash recovery for the mlx-serve server process.

- **New `CrashRecoveryMode`** (Settings → Server → Crash Recovery):
  - *Ask* (default) — today's behavior, modal alert on crash
  - *Auto-restart* — exponential backoff restart + macOS notification
  - *Auto-restart (silent)* — same, no notification (headless-friendly)
- **Pure decision logic** in `CrashRecovery.swift` — fully testable, no side effects
- **Guards**: OOM → always modal (would recur); exit 0 → clean exit; startup failures → always modal; max 3 retries in 5 min window; backoff 1s→2s→4s…30s cap; counter resets after 5 min stable uptime

### Files changed
| File | Change |
|---|---|
| `Models/CrashRecovery.swift` | NEW — enum + pure logic |
| `Services/ServerManager.swift` | `handleTermination` rewrite, restart orchestration |
| `Views/SettingsView.swift` | `CrashRecoveryRow` picker |
| `Tests/CrashRecoveryTests.swift` | NEW — 13 tests |

## Test plan

- [x] Unit tests: 13/13 pass (`swift test --filter CrashRecoveryTests --enable-xctest`)
- [x] Full suite: 3330/3330 pass, 0 failures (`swift test --enable-xctest`)
- [x] Build: `swift build` + `FAST_DEV=1 bash app/build.sh`
- [x] Manual: Settings UI persists mode; `kill -9 $(pgrep -f "MacOS/mlx-serve")` triggers auto-restart when set to Auto-restart
- [x] Default `.ask` mode preserves today's crash-alert behavior

**Machine**: macOS 26, Apple Silicon

### DIFF
--- a/app/Sources/MLXServe/Models/CrashRecovery.swift
+++ b/app/Sources/MLXServe/Models/CrashRecovery.swift
@@ -21,6 +21,16 @@ enum CrashRecoveryMode: String, Codable, CaseIterable, Identifiable {
 
     static let defaultMode: CrashRecoveryMode = .ask
     static let defaultsKey = "crashRecoveryMode"
+
+    /// Clear the persisted mode when a server or all-settings reset applies.
+    static func resetIfApplicable(_ selection: SettingsSelection) {
+        switch selection {
+        case .all, .category(.server):
+            UserDefaults.standard.removeObject(forKey: defaultsKey)
+        default:
+            break
+        }
+    }
 }
 
 /// Pure decision logic for crash recovery — no side effects, fully testable.
@@ -42,7 +52,7 @@ enum CrashRecovery {
         guard wasRunning else { return false }
         guard exitCode != 0 else { return false }
         guard !isMemoryFailure else { return false }
-        guard crashCount < maxRetries else { return false }
+        guard crashCount <= maxRetries else { return false }
         return true
     }
 
--- a/app/Sources/MLXServe/Models/CrashRecovery.swift
+++ b/app/Sources/MLXServe/Models/CrashRecovery.swift
@@ -1,0 +1,59 @@
+import Foundation
+
+/// How the app handles an unexpected mlx-serve crash (#265).
+enum CrashRecoveryMode: String, Codable, CaseIterable, Identifiable {
+    /// Show a modal alert (today's behavior).
+    case ask
+    /// Auto-restart with backoff + macOS notification.
+    case autoRestart
+    /// Auto-restart with backoff, log only (headless-friendly).
+    case silent
+
+    var id: String { rawValue }
+
+    var label: String {
+        switch self {
+        case .ask:         return "Ask (show alert)"
+        case .autoRestart: return "Auto-restart"
+        case .silent:      return "Auto-restart (silent)"
+        }
+    }
+
+    static let defaultMode: CrashRecoveryMode = .ask
+    static let defaultsKey = "crashRecoveryMode"
+}
+
+/// Pure decision logic for crash recovery — no side effects, fully testable.
+enum CrashRecovery {
+    static let maxRetries = 3
+    static let crashWindow: TimeInterval = 300       // 5 minutes
+    static let stableThreshold: TimeInterval = 300   // reset counter after 5 min uptime
+
+    /// Whether the server should be auto-restarted after a crash.
+    static func shouldAutoRestart(
+        mode: CrashRecoveryMode,
+        wasRunning: Bool,
+        exitCode: Int32,
+        isMemoryFailure: Bool,
+        crashCount: Int,
+        maxRetries: Int
+    ) -> Bool {
+        guard mode == .autoRestart || mode == .silent else { return false }
+        guard wasRunning else { return false }
+        guard exitCode != 0 else { return false }
+        guard !isMemoryFailure else { return false }
+        guard crashCount < maxRetries else { return false }
+        return true
+    }
+
+    /// Exponential backoff delay in seconds: 1, 2, 4, 8 … capped at 30.
+    static func backoffDelay(attempt: Int) -> TimeInterval {
+        min(30.0, pow(2.0, Double(attempt - 1)))
+    }
+
+    /// Whether the crash window has expired (counter should reset).
+    static func crashWindowExpired(lastCrash: Date?, window: TimeInterval) -> Bool {
+        guard let last = lastCrash else { return true }
+        return Date().timeIntervalSince(last) > window
+    }
+}
--- a/app/Sources/MLXServe/Services/ServerManager.swift
+++ b/app/Sources/MLXServe/Services/ServerManager.swift
@@ -379,7 +379,6 @@ class ServerManager: ObservableObject {
     }
 
     private func handleTermination(exitCode: Int32) {
-        NSLog("[crash-recovery] handleTermination called, exitCode=\(exitCode), status=\(status.label)")
         pollSource?.cancel()
         pollSource = nil
         healthTask?.cancel()
@@ -423,8 +422,6 @@ class ServerManager: ObservableObject {
             isMemoryFailure: memFail,
             crashCount: crashCount, maxRetries: CrashRecovery.maxRetries
         )
-
-        NSLog("[crash-recovery] mode=\(mode.rawValue) exitCode=\(exitCode) memFail=\(memFail) crashCount=\(crashCount) shouldRestart=\(shouldRestart)")
 
         guard shouldRestart else {
             status = .error("Exited unexpectedly")
--- a/app/Sources/MLXServe/Services/ServerManager.swift
+++ b/app/Sources/MLXServe/Services/ServerManager.swift
@@ -1,6 +1,7 @@
 import Foundation
 import AppKit
 import Darwin
+import UserNotifications
 
 @MainActor
 class ServerManager: ObservableObject {
@@ -65,6 +66,9 @@ class ServerManager: ObservableObject {
     private var healthTimer: Timer?
     private var healthTask: Task<Void, Never>?
     private var pollSource: DispatchSourceTimer?
+    private var restartTask: Task<Void, Never>?
+    private var crashCount = 0
+    private var lastCrashDate: Date?
     let api = APIClient()
     /// True while the tray popover is on screen. Drives the live /props
     /// ticker — when the popover is closed there's nothing to render, so we
@@ -271,6 +275,8 @@ class ServerManager: ObservableObject {
     private var quitObserver: NSObjectProtocol?
 
     func stop() {
+        restartTask?.cancel()
+        restartTask = nil
         pollSource?.cancel()
         pollSource = nil
         healthTask?.cancel()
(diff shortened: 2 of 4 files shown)


ddalcu:

- `app/Sources/MLXServe/Services/ServerManager.swift:195` — `launch()` resets `crashCount = 0` / `lastCrashDate = nil`, and the auto-restart path goes through `start()`/`startHeadless()` → `launch()`, so the retry budget and the backoff are wiped on every restart.
  Failure scenario: a server that crashes reliably (bad model, a request that always segfaults) crashes → `crashCount` 0→1 → restart after 1 s → `launch()` zeroes the counter → next crash is again attempt 1/3 at 1 s. `crashCount <= maxRetries` never fails, `backoffDelay` never exceeds 1 s, and the app restarts the process forever in a ~1 s loop instead of giving up after 3 tries. `scheduleStabilityReset()` and `CrashRecovery.maxRetries` are unreachable in practice.
- `app/Sources/MLXServe/Services/ServerManager.swift:485` — `sendCrashNotification` touches `UNUserNotificationCenter.current()` without the `Bundle.main.bundleIdentifier != nil` guard that `TaskNotifier` uses for exactly this reason (`TaskNotifier.swift:25`, "crashes with no bundle identifier — e.g. under `swift test`").
  Failure scenario: any test or headless context that drives `handleTermination` in `.autoRestart` mode traps instead of returning; it also never calls `requestAuthorization`, so in the app the notification is dropped silently unless the Tasks feature happened to ask first.
- `app/Sources/MLXServe/Views/SettingsView.swift:378` — `CrashRecoveryRow` seeds `@State mode` once from `UserDefaults`, but `CrashRecoveryMode.resetIfApplicable` removes the key behind its back.
  Failure scenario: user picks "Auto-restart", hits Reset in the server section → the effective mode is `.ask` but the picker still shows "Auto-restart"; toggling any other row that re-renders is not enough to resync, and the next `onChange` writes the stale value back.

ddalcu:

Re-reviewed the increment `c1c9b4d5..d19f0544`. The notification guard is correct. The retry-budget fix is not: the reset moved out of `launch()` into the two functions the auto-restart path actually calls.

- `app/Sources/MLXServe/Services/ServerManager.swift:125` — `start()` (and `startHeadless()` at `:182`) now zero `crashCount`/`lastCrashDate`, but the auto-restart task calls exactly those two (`:465`, `:467`), not `launch()` — so the comment "Auto-restart goes through launch() directly and must NOT reset" does not describe the code, and the counter is still wiped on every restart.
  Failure scenario: unchanged from the last round — a server that crashes reliably goes `crashCount` 0→1, restarts after 1 s via `start()`, which sets it back to 0; the next crash is again attempt 1/3 at 1 s. `crashCount <= maxRetries` never fails, `backoffDelay` never exceeds 1 s, and the app respawns the process forever. The `.error("Restarting in …")` status passes `start()`'s `status != .running, status != .starting` guard, so nothing blocks the reset.
- `app/Tests/MLXCoreTests/CrashRecoveryTests.swift:132` — `testRetriesAccumulateAcrossRestartCycles` increments a local `var crashCount` and calls the pure `shouldAutoRestart`; it never constructs a `ServerManager` or drives `handleTermination`, so it cannot observe the reset it is named after.
  Failure scenario: the test is green against the code above. A red-on-revert regression test has to exercise the object that owns the counter — crash while `.running`, let the restart land, crash again, and assert the second decision saw `crashCount == 2`.
- `app/Sources/MLXServe/Views/SettingsView.swift:1308` — the new `.onReceive` sets `mode = .defaultMode`, which trips the `.onChange(of: mode)` writer two lines above and re-persists the key `resetIfApplicable` just removed.
  Failure scenario: with Settings open on a picker showing "Auto-restart", hitting Reset removes `crashRecoveryMode` and then immediately writes it back as `"ask"`. Harmless while `defaultMode == .ask`, but "never chosen" and "explicitly chose Ask" stop being distinguishable, so a future default change silently does not reach anyone who ever pressed Reset. `testResetPostsNotification` passes because no view is mounted.

ddalcu:

Re-reviewed the increment `d19f0544..54f16592`. The retry-budget defect is now genuinely fixed: `resetCrashCounter: false` at `ServerManager.swift:469`/`:471` keeps the counter across restarts, so the sequence reaches `crashCount == 4 > maxRetries` and falls back to the modal, with the backoff climbing 1→2→4 s. Two findings from the previous round are unchanged in this head.

- `app/Tests/MLXCoreTests/CrashRecoveryTests.swift:132` — `testRetriesAccumulateAcrossRestartCycles` still increments a local `var crashCount` and calls the pure `shouldAutoRestart`; it never constructs a `ServerManager`, so it did not fail against either of the two broken revisions of this fix and will not fail if the flag regresses.
  Failure scenario: delete the `resetCrashCounter:` parameter and let `start()` unconditionally zero the counter — the suite stays green and the infinite 1-second respawn loop ships again. `CLAUDE.md` asks a bug fix for a regression test that is red on revert; the reverting edit here is a default argument, and nothing observes it. The seam is already there — assert on the `ServerManager` after driving two crashes, or hoist the counter update into a pure `CrashRecovery` helper that takes `(crashCount, lastCrash, now)` and test that.
- `app/Sources/MLXServe/Views/SettingsView.swift:1308` — the `.onReceive` sets `mode = .defaultMode`, which trips the `.onChange(of: mode)` writer at `:1304` and re-persists the key `resetIfApplicable` just removed.
  Failure scenario: with Settings open on a picker showing "Auto-restart", Reset removes `crashRecoveryMode` and immediately writes it back as `"ask"`. Benign while `defaultMode == .ask`, but "never chosen" and "explicitly chose Ask" stop being distinguishable, so a later default change silently never reaches anyone who pressed Reset. `testResetPostsNotification` passes because no view is mounted. Guarding the write (`onChange` only when the new value differs from what is stored, or a flag set across the programmatic assignment) keeps the removal.

ddalcu:

Re-reviewed the increment `54f16592..1b830c35`. The retry-budget fix from the last head still holds. Both of this increment's changes are defective.

- `app/Tests/MLXCoreTests/CrashRecoveryTests.swift:157` — the scan counts the literal `resetCrashCounter: false` anywhere in `ServerManager.swift`, and the file contains three occurrences: the two call sites *and* the comment at `ServerManager.swift:123` ("Auto-restart passes resetCrashCounter: false so…"), so `hits >= 2` survives deleting a call site.
  Failure scenario: drop `, resetCrashCounter: false` from `:471` — exactly the regression that shipped two heads ago and respawned the process forever at 1 s — and the needle count is still 2, so the guard is green. Verified by simulating that edit: 3 hits → 2. A needle that matches its own explanatory comment is the "passes against its own needles" case the root `CLAUDE.md` names when it bans source-scan tests; the counter lives on `ServerManager`, so the behavioural seam is to drive two crashes and assert the second decision saw `crashCount == 2`, or to hoist the update into a pure `CrashRecovery` helper over `(crashCount, lastCrash, now)`.
- `app/Sources/MLXServe/Views/SettingsView.swift:1316` — `suppressWrite` is set before assigning `mode = .defaultMode`, but `onChange` only fires when the value actually changes, so a reset while the picker already shows "Ask" leaves the flag latched true with nothing to consume it.
  Failure scenario: the mode is already `.ask` (fresh install, or any second Reset in a row). Press Reset in the server section → `suppressWrite = true`, `mode` is unchanged so `:1307` never runs. The user then picks "Auto-restart": `onChange` fires, takes the suppress branch at `:1308`, clears the flag and returns **without writing**. The picker reads Auto-restart while `loadCrashRecoveryMode()` still answers `.ask`, so the next crash shows the modal and the setting appears not to work until the app is relaunched. Comparing against what is stored (`UserDefaults.standard.string(forKey:) != newValue.rawValue`) needs no flag and has no latch to strand.

ddalcu:

Re-reviewed the increment `1b830c35..39959c9c`. Both previous findings are properly closed: `CrashRecovery.CounterState` is a pure, injectable-clock type with real behavioural tests (`recordCrash` accumulation, `reset`, window expiry), and the picker's `loadCrashRecoveryMode() != newValue` compare has no flag to strand — a reset while already on "Ask" fires no `onChange` at all, and one that does fire finds the stored value already equal and skips the write. Calling `launch()` directly is the right shape for the counter. It also drops two things `start()`/`startHeadless()` were carrying.

- `app/Sources/MLXServe/Services/ServerManager.swift:465` — the restart task calls `launch()` with no `guard status != .running, status != .starting`, and `start()` never cancels `restartTask` (only `stop()` does, at `:282`), so a manual start during the backoff leaves a second launch armed.
  Failure scenario: the server crashes and the status reads "Restarting in 4s (1/3)". The user presses Start in the chat window or the tray — `start()`'s own guard passes because the status is `.error`, so process A launches and the counter is reset. The timer then fires, `Task.isCancelled` is false, and `launch()` runs again: `killOrphanedServers(on: port)` kills A, process B starts, and the user's start silently becomes a second full model reload. Worse, A's `terminationHandler` (`:248`) reaches `handleTermination`, which takes no process identity and unconditionally does `process = nil` plus cancels `pollSource`/`healthTask` (`:380`-ish) — so the app tears down the live B's polling and loses its handle to it. `stop()` and the `willTerminateNotification` teardown then have nothing to terminate, which is the #133 orphaned-server class. Re-checking the status inside the task before launching (or cancelling `restartTask` in `start()`/`startHeadless()`) closes it.
- `app/Sources/MLXServe/Services/ServerManager.swift:459` — the headless arm rebuilds the argv itself and omits the `modelsDir` insert that `startHeadless()` does (`if !modelsDir.isEmpty, !dirs.contains(modelsDir) { dirs.insert(modelsDir, at: 0) }`), so a restart can serve a different root set than the launch it is replacing.
  Failure scenario: unreachable today — both callers pass `ServerManager.modelsRoot`, which `scanRoots` already returns — but that insert exists precisely for a primary root outside the scanned set, and this is now a second argv construction site with no compiler tying it to the first. The same class the repo already guards for `ensureLoaded`'s cold-load `LoadRequest`. Factoring the two argv builds into one helper that both `startHeadless()` and the restart call would keep them in lockstep.

ddalcu:
🤖 **MLXServe-PRBot approved** — reviewed at head `c4f3e72d`, no issues found.

Both findings from `39959c9c` are closed. `start()`/`startHeadless()` now cancel `restartTask`, and the task re-checks `guard case .error = self.status` before launching — belt and braces, and the status cannot drift to `.running` on its own during the backoff because `handleTermination` already cancelled `pollSource`/`healthTask`. `headlessArgs` gives the two headless launches one construction site; the restart still passes `modelsDir: ""`, but `scanRoots` returns `resolvedDownloadRoot` then `builtInRoot`, so the insert is a no-op for both of the callers that exist and nothing diverges.

ddalcu:
🤖 **MLXServe-PRBot approved** — reviewed at head `9f007d2b`, no issues found.

This head is the upstream merge `9f007d2` on top of `c4f3e72` — no new crash-recovery work. I verified the merge resolved nothing in this PR's favour or against it: comparing the branch's own patch against its base before and after (`fa76a4b5...c4f3e72` vs `origin/main...9f007d2`) gives 501 changed lines on both sides and a byte-identical diff, so `CrashRecovery.swift`, `ServerManager.swift`, `SettingsView.swift` and `CrashRecoveryTests.swift` carry exactly the content approved at `c4f3e72`.

That approval stands: `start()`/`startHeadless()` cancel `restartTask`, the task re-checks `guard case .error = self.status` before launching, `headlessArgs` is the single construction site for both headless launches, and `CrashRecovery.CounterState` is a pure injectable-clock type with behavioural tests.

PR #412, opened 2026-09-12
perf(mtp): share Qwen4 drafting and verification across requests

beamivalice:
Flash Next requests speculate together. A group planner picks a draft width per request from a learned mean round cost, the MTP head drafts every active row in one graph, and one row-axis verify forward serves the group while each row keeps its own KV, recurrent state and rollback offsets. Legacy per-request MTP stays as the fallback and as the OFF arm.

**Gate.** One environment variable, `MLX_SERVE_MTP_GROUP_PLANNER`, default on, `=0` opts out. No other switches, no ablation overrides, no profiling fixtures. Round-cost persistence keeps its v3 format; group prices are runtime-only. Per-request opt-out: `enable_batch_mtp: false` in the request body.

**Measured** on M5 Max, Qwen3.8-Flash-Next mixed-4/8-bit, affine-8 KV, four concurrent streams, three paired boots, median gain over the legacy policy:

| Context | Legacy MTP tok/s | Group MTP tok/s | Median gain |
|---|---:|---:|---:|
| 4k | 102.2 | 133.0 | +30.1% |
| 16k | 110.5 | 139.6 | +26.8% |
| 64k | 105.5 | 125.0 | +18.9% |

Two streams against actively speculating legacy MTP: about +27% at all three contexts. Single-stream output is byte-identical to main with the gate off.

**Verification** on this tree: unit suite 2503 passed, 0 failed; `zig build test` 9/9 steps. The live two-stream EOS/cancellation parity gate `tests/test_mtp_batch_stops.py` (registered in `tests/CLAUDE.md`) passes on a fresh server; it records its references from a restored prefix, since on main a prompt's first run and its cached restore are not bit-exact. The fault-injected head sweep `qwen4MtpForward: every error path of the MTP head forward releases its stream` crashed with SIGSEGV on main `0814cf3` and on every earlier head of this PR; the cause was an uninitialized `queries` handle freed on an early error return in `gatedFullAttnProjected`, fixed here together with the in-flight handle ownership in `qwen4MtpForwardBatched`. The sweep now runs to its assertion and still reports 8 to 12 byte error-path leaks in helpers that predate this PR (`qmatmulBits`, `applyClosure`, `hcGroupNorm`, `moeMLP2WithRouter`); those are left for a separate sweep.

**Since the last push**, in response to review: markdown changes dropped from the diff, every per-lever switch and extra environment read removed, profiling-only fixtures removed, duplicated bodies merged (shared-expert tail, attention/GDN verify rows, rerank rescore, planner probe, finish/publish bookend, batched-head output buffer), comments cut to intent, tests named for what they prove, the bench harness kept out of the tree, the live stop gate restored and registered, every in-flight handle on the MTP head's error paths owned. `src/transformer.zig` adds 4.8k lines instead of 6.1k.


ddalcu:
Reviewed on top of #391 (now merged to main as fd9c0c3). Moving this to draft. Not mergeable in this form.

The GitHub diff says +20k, but that includes an old main folded into the commit. The real delta against 391+main is about +11.8k/-1.4k across 15 files, and most of it is not the feature.

What is wrong:

1. The suite is red on M4. `verify expert reduction preserves unsort multiply and sum` asserts a kernel that only exists on applegpu_g17 non-Ultra and errors with ExpertReduceDeclined everywhere else. The sibling test gates on verifySharedHardware(); this one does not. A kernel that only exists for one GPU generation on one machine does not ship.
2. The PR text says everything is env gated. It is not. batchedMtpHeadEnabled() and mtpVerifyMoeRowsEnabled() default true, so every qwen4 group of 2+ MTP slots takes the new batched head path by default. Solo mtpRoundVerify now goes through forwardRowAxisVerify, mtpStopPrefix truncates on EOS, mtpMaybePreDraft early-outs, mtpDetachHead truncates. None of that is byte checked against 391 on a live model.
3. 17 new MLX_SERVE_MTP_* env levers, and plannerKernelKey bakes their combination into the persisted cost key. The rule in CLAUDE.md is one lever per lossy/tradeoff path. This is one feature, one gate.
4. round_cost store version 3 -> 6 in one PR, with STORE_VERSION left at 3 and comments describing v4/v5 that never existed. That wipes every user's round-cost table for a default-off feature. Store bumps are my call, and it is one bump, not three.
5. Two cost stores layered on each other (observeGroup grid plus observeShape cells) with 256 KB stack buffers. Pick one.
6. You rewrote the CLAUDE.md rule "No KV snapshots across verify" into "Unpin KV refs before verify eval" so VerifyRollback.capture can take ctx.cache.snapshot() across the grouped verify, then reimplemented KVCacheSnapshot.releaseHandles as unpinAppendKv. Do not edit a rule to license the thing it forbids. Offset-only truncate, or reuse releaseHandles.
> Why the rule exists (gotcha "copy-on-write class"): KVCache.snapshot() refcount-shares the buffers. While any snapshot handle is alive, the verify's slice_update cannot donate and silently copies the whole KV per layer per round. At 64k on the 27B that was ~4.3 GB of copy traffic per MTP round. The rule is a memory/bandwidth invariant, not a style preference.
>When a rewrite would be fine: if grouped per-row rollback genuinely cannot be expressed as offset-only truncate + per-position SSM capture (the pattern solo verify already uses),  what breaks, and the rule gets the new WHY added, not replaced. I do not see that case made anywhere in the PR. 

7. A source-scan test got extended (@embedFile transformer.zig + a flushDeferredPleGroup needle). No source-scan tests.
8. Test-only helpers: mtpRerankForRow (returns modes[row]), mtpPredraftAllowed, mtpPlannerStop, freeMtpHeadOutputs(comptime free), releaseMtpParts(comptime free), continueOpenMtp, l10SoloVerifyPreL10 kept as an oracle. Delete them.
9. tests/test_mtp_batch_stops.py and bench_mtp_planner.py hardcode /Users/beam/llm/agent-ops/bin/agentctl and /usr/local/bin/fanctl. Nobody else can run them.
10. 11 new gotcha entries, 109 lines, mostly iteration logs ("failed with expected 2, found 0"). One story per gotcha, written once after the work lands.
11. No CHANGELOG line for the feature. zig fmt fails on five files.

The nucleus (batched qwen4 head drafting, row-axis grouped verify, the planner) is probably worth having, but it cannot be reviewed buried under this. Rebase onto current main, drop everything above, ship one gate, and post OFF/ON numbers against #391's batched MTP (not against plain decode) on the same boot with engagement lines. Your numbers are M5 Max only; on M4 the shared-hardware kernels decline, so say what is left there.


Also please change target to main, and merge latest main into this branch, since the mtp pr has been merged now.

beamivalice:
Addressed the review in squash commit 7909ca7 and retargeted this PR to `main`, including latest main `0814cf3`.

- One feature gate, enabled by default with `MLX_SERVE_MTP_GROUP_PLANNER=0` as the opt-out; restored the original solo verifier and legacy scheduler path. Global-OFF and request-OFF each matched seven main/#391 streams at N=1/2/4 byte-for-byte, including finish reasons and completion counts.
- Preserved v3 cost persistence; replaced the duplicate group stores with one runtime-only shape table. Rollback now keeps KV offsets and SSM state without pinning KV buffers; the original invariant is restored.
- Removed the individual kernel switches/bitmask, named trivial helpers, copied oracle, extended source scans and unrelated stacked plain-QSA optimization. Harnesses are portable, gotchas consolidated, CHANGELOG added, and formatting clean.
- Fixed the hardware guard on the expert-reduction test and rejected `air64_v27` as a GPU generation. M5-specific kernels decline on M4; the PR documents the generic fallbacks and makes no M4 performance claim.

Final merged-tree ReleaseFast suite: **2,503 passed, 173 gated skips, zero failures**; Swift suite green. Also passed 54 exact real-model comparisons and live EOS/cancellation, surviving-row speculation, failure restoration, cold-head resume and 48-token re-entry checks.

The updated measurement note reports same-boot native-MTP OFF/ON comparisons with engagement evidence. Four-stream median gains are +30.1% / +26.8% / +18.9% at 4k / 16k / 64k, with the remaining 64k variance disclosed. Because legacy four-stream scheduling normally demotes to ordinary batching, a separate two-stream comparison against actively speculating legacy MTP is included: approximately +27% at all three contexts.

[Full results, methodology and limitations](https://github.com/beamivalice/mlx-serve/blob/7909ca7983c1080b2bdffd0100ab9ca88010eb8d/notes/MTP-64K-TUNING.md). GitHub CI has not reported a result for the new head yet.

Default-ON behavior was additionally verified with the environment variable unset: full suite green, and live EOS/cancellation plus continued speculation for the surviving row passed.

ddalcu:

No defect found at head `7909ca79`, but this is a ~10k-line diff and the pass was **not** exhaustive — recording the scope so the next run knows what is still unread.

Reviewed closely:
- `src/mtp_group_planner.zig` in full — `History.observe`/`expected` decay and the running-min survival envelope, `consider`'s `trials[width-1] >= MIN_SAMPLES` gate (safe: `trials` is non-increasing in `j`, so a sampled tail implies sampled prefixes), `validPrice`'s mean-bound ordering vs `choose`'s `lower <= 0` refusal, `execution`'s prime/plain/speculative split, `Recovery`'s burst + `rounds >= 8` bound, `probeWidth`'s ladder.
- `src/mtp_group_cost.zig` in full — `observeShape`'s victim scan (no holes can form, so the break-on-empty is sound), the `.wall_mismatch` early-out ordering against the evict-and-reset, the EMA/`mean_weight_sq` recurrences.
- `src/scheduler.zig` planner integration — `tryPlannerTick`'s early returns and its recursion through `runDecodeTick`, the per-tick reset of `mtp_planner_width`/`probe`/`recovering` at `scheduler.zig:6358` (so a plain row cannot carry a stale width into `mtpRoundPlan`), `PlannerPrices.keyFor`/`sample`/`price` shape keying, and the two `observeShape` sites. Checked specifically whether `GroupShape.padding` — which `plannerInputSame` does *not* zero, unlike `attention`/`moe`/`head_calls` — could make a recorded round unreachable from a planner query: it cannot, because the planner's rounds go through `mtpGroupVerify` → `forwardRowAxisVerify`, which leaves `verify_len == 1 + m` per row, so `observeGroupRound` records `padding == 0`. Worth a comment there: the invariant is load-bearing, and routing a planner group through the padded merged path (`scheduler.zig:8002` sets `st.verify_len = width`) would silently orphan every sample it takes.
- `runBatchedDecodeTickInner`'s plain-price observation — `gaps`/`emitted` are only read under `live_n == N`, so `batch.len == N` and no `undefined` gap reaches the table.

Not reviewed: the 6,352 added lines in `src/transformer.zig` beyond a spot check of the new verify-row helpers' handle ownership (`attnVerifyRows`/`GdnVerifyInputs.slice` errdefer/deinit scoping looked correct, incl. the partial-`inline for` failure path), and the Metal kernel sources/parity arms.

ddalcu:

Re-reviewed at head `b864116f` (squashed onto `0814cf3`, ~1.6k net lines lighter than `7909ca79`). The trimming is sound where I checked it: the twelve per-kernel env levers are gone, `latencyMs()`'s per-row `getenv` + `parseFloat` in the decode path is now the `LATENCY_MS` constant, `planProbe` folds the two probe blocks without changing an outcome (candidate-then-price vs price-then-candidate reach the same verdicts, and the "calibration blocked" log still fires only on a fully-built candidate that lost to latency), and `verifyRowsJoined` keeps `gdnVerifyRows`' `combined_proj`/`ane_prefill` declines in the wrapper. The GDN arm's missing `defer values.deinit()` is correct, not a leak — `gatedDeltaNetProjected` takes the four handles by value and frees them through its own top-of-function defers, unlike the attention arm which passes `&values` and retains them.

- `src/transformer.zig:21976` — inside the row-slicing loop the freshly created `stream_i` (and `mixed_i`/`logits`) are live but not yet in `buf`, and `buf.built` is only advanced at `:21987`, so the `errdefer buf.deinit()` cannot see them.
  Failure scenario: any `mlx.check` failure at `:21977`, `:21980` or `:21984` — reachable since #353 made MLX errors catchable rather than `exit(-1)`, e.g. a slice issued at the Metal working-set edge — unwinds with one or two `mlx_array` handles orphaned. The `logits` row is 248320-wide, so a round that fails on row *i* leaks roughly half a megabyte of GPU memory, and the scheduler's retry re-enters the same path. This is exactly the gap the new `Qwen4MtpOutBuf` doc comment at `:21496` claims to have closed ("a failure part-way frees exactly them") — it inherits the manual `built` counter's blind spot rather than removing it. Assigning `buf.items[i]` before each fallible slice (fields default to `.ctx = null`, which `deinit` already skips) and bumping `built` first would make the claim true.

ddalcu:

The increment `b864116f..458b5efb` is deletions only: `tests/bench_mtp_planner.py` (a benchmark harness — no coverage lost) and `tests/test_mtp_batch_stops.py`. The `src/transformer.zig:21976` in-flight handle leak from the last round is unchanged at this head.

- `tests/test_mtp_batch_stops.py` — removing it leaves the PR with **no** integration coverage for an HTTP-observable feature: after this increment nothing under `tests/` mentions `enable_batch_mtp`, `MLX_SERVE_MTP_GROUP_PLANNER`, `[mtp-planner]` or the row-axis verify, and `tests/CLAUDE.md` is untouched.
  Failure scenario: `enable_batch_mtp` is a new request field on `/v1/chat/completions` and `/v1/completions` (`src/server.zig:8156`, threaded through all four stream/non-stream handlers), and grouping changes what a client receives — event sequence, finish reason, completion count. The deleted script was the only thing asserting grouped output equals solo across the two edges the PR body cites as live-validated: EOS landing inside an accepted draft block (with the surviving row required to keep speculating) and mid-round cancellation of one row of a group. It also carried the engagement assertion `[batched] row-axis mtp verify engaged (slots=2`, which is how `CLAUDE.md` says a spec arm is proven ("engagement COUNTS in tests"). The Zig unit tests that remain price plans and fold cost cells; none of them runs a request, so a regression that publishes a cancelled row or drops the survivor out of speculation now ships green. `CLAUDE.md`'s testing rule asks for the integration script precisely when the behaviour is HTTP-observable.
- `src/transformer.zig:21976` — unchanged from `b864116f`: `stream_i`/`mixed_i`/`logits` are live before `buf.items[i]` is assigned and `buf.built` is bumped at `:21987`, so `errdefer buf.deinit()` cannot free them.
  Failure scenario: an `mlx.check` failure at `:21977`/`:21980`/`:21984` orphans one or two handles, including the 248320-wide logits row.

beamivalice:
Both items from the last review are in `822e859`:

- `tests/test_mtp_batch_stops.py` is back and registered in `tests/CLAUDE.md`; it records its references from a restored prefix so the comparison shares the cache path with the grouped runs.
- `qwen4MtpForwardBatched` now owns every in-flight handle on its error paths (`errdefer` at each `mlx_array_new`, the `last_parts` and `h_parts` loops included, and one success-path leak of `mlp_inj` per grouped round). The fault-injected head sweep used to SIGSEGV on main and on this PR; the cause was an uninitialized `queries` handle freed on an early error return in `gatedFullAttnProjected`, fixed in the same commit. The sweep now runs to its assertion; its remaining 8–12 B deltas are pre-existing error-path leaks in `qmatmulBits`, `applyClosure`, `hcGroupNorm` and `moeMLP2WithRouter`, left for a separate sweep.

Unit suite 2503 passed, 0 failed; batched-head N=1/2/4 exactness tests pass.

ddalcu:
🤖 **MLXServe-PRBot approved** — reviewed at head `822e859a`, no issues found.

Both findings from `458b5efb` are closed. `tests/test_mtp_batch_stops.py` is restored byte-identical (sha256 matches the copy at `b864116f`) and now carries a `tests/CLAUDE.md` row naming both gates, the forced depth, the second-run reference rule and the `[batched] row-axis mtp verify engaged (slots=2` requirement.

The in-flight handle leak is fixed with per-site `errdefer` rather than early assignment, and I checked it does not trade the leak for the double-free class: `hcWrite` consumes only `stream` (`transformer.zig:20011`), never `inj`, so dropping `mlp_inj[i] = .{ .ctx = null }` at `:21922` hands the handle back to the function-level `defer` at `:21887` — that line was leaking it, not protecting it. The `tok2` errdefer is null-guarded and `tok2` is nulled at the free, so a later `mtpPadSeq` failure cannot re-free it; the `h_i` errdefer re-evaluates to the post-`hcWrite` handle, which is the one that needs freeing since `hcWrite` frees `stream` only after its last fallible op; and every new loop-body errdefer is discharged before `buf.built` advances, so the outer `buf.deinit()` and the inner ones never overlap. The two `errdefer`-before-`reshape` reorders and `queries` moving from `undefined` to `mlx_array_new()` close the same class on their own paths.

beamivalice:
Warm decode ladder with the pushed lever, auto planner (no forced depth, no latency override), 300 tokens per request, two runs per cell, fans at max. M5 Max 128 GB, Qwen3.8-Flash-Next mixed-4/8-bit, `--kv-quant 8 --mtp --mtp-head-kv-quant --max-concurrent 4 --ctx-size 400000`, every request served from a cached prefix at its full context. Aggregate tok/s across the streams, run 1 / run 2:

| Context | 1 stream | 2 streams | 4 streams |
|---|---:|---:|---:|
| 4k | 96.9 / 106.4 | 106.1 / 106.9 | 127.9 / 132.8 |
| 16k | 111.7 / 107.1 | 119.8 / 127.4 | 147.7 / 148.4 |
| 64k | 106.4 / 106.1 | 118.6 / 120.8 | 137.6 / 128.4 |
| 128k | 102.0 / 96.9 | 109.8 / 111.3 | 120.7 / 122.0 |

Output gap p95: 35 to 42 ms at one stream, 42 to 47 at two, 69 to 79 at four. The planner ran about 23% of tokens plain across the ladder (one `calibration blocked` line), so the four-stream cells at 4k and 64k still sit a few percent under a forced depth-2 round; that is planner policy, not this change.

Binary: this head plus a local single-stream verify-kernel change that is not in this PR; it moves only the 1-stream column.

PR #414, opened 2026-09-13
Fold a long user turn, and stop estimating the transcript

lojza3d:
## Why

A pasted file or log in the composer is the one way to hand the model a long text in a single turn. The alternative, letting it read the file with `readFile`, means several rounds through a tool that truncates its output. So a user bubble several screens tall is sometimes necessary, and today it makes the conversation around it hard to read and scroll: the reply you want is somewhere under four screens of your own paste.

## What this does

A user turn that would run past about 15 lines folds, with the last two lines fading into the bubble and a `Show more` / `Show less` control at its bottom left. Copy and edit work on the whole text regardless. A folding turn takes the full column in both states so nothing changes width when it opens. The line count is an estimate biased toward not folding, because a `Show more` that reveals nothing is worse than a long bubble. Laying out a very long `Text` takes the main thread for a moment, so a spinner stands in for the control from the click until the layout is done.

A ~22k character message screenshots below.

Before:
before

After:
after

## What it turned up

Folding a block is a large height change, and it reproduced a defect the transcript has had since its first commit: after a chat switch, a jump to the end, or a fold, the view sometimes rendered blank until the mouse moved, and rows then filled in from the bottom one layout pass at a time. The scroller thumb even changed size while merely scrolling.

The transcript was a `LazyVStack`. A lazy stack's content height is an estimate from the rows it has built, and with rows from one line to four screens that estimate is noise. A trace showed the height dropping from 6764 to 1242 and back within two milliseconds of a fold. Every anchor, offset correction and `scrollTo` aimed at those numbers could land the visible rectangle outside the rows that existed.

So, in order of weight:

- The transcript is a plain `VStack`. Exact heights cost about 60 MB across ten long chats (130 to 189 MB in Activity Monitor after clicking through them and changing column width and density); wrong heights cost every scroll decision in the file.
- A chat opens at its end by layout, `defaultScrollAnchor(.bottom, for: .initialOffset)` on a scroll view whose identity carries the session id, so `transcriptShown` no longer scrolls anywhere.
- A shrinking row (a fold, a thinking block closing, an edit field replacing the bubble) keeps the control the reader clicked where it is: `ChatScrollState` snapshots offset and height at `rowWillShrink` and drops the offset by what the row has lost on every frame. The size-change anchor cannot do this; it holds the end only when you are already there.

One condition hid behind every earlier attempt at that last rule, and it is the thing I would look at first in review: the frame after a fold has content shorter than the stale offset, so the distance from the bottom reads deeply negative. That is the rubber-band shape, and the pin rule read it as arrival at the end. Following re-engaged, and the bottom anchor plus the correction rule dragged every fold to the bottom. While a shrink is in progress, a negative distance is not the bottom. `testAFoldOvershootIsNotArrivalAtTheBottom` pins it.

Smaller things along the way, each with a one-line comment at the site: the fade is an overlay in the bubble's colour rather than a `mask` (a mask renders an unfolded turn into an offscreen layer thousands of points tall); fold state lives in the row with a `FoldStore` beside the transcript so a density change, which rebuilds every row, does not refold, and a chat switch clears it.

## Testing

`swift test` green (3335 tests). Tried by hand on messages from 20k to 110k characters, several long turns in one chat, folding and unfolding while a reply streams in, thinking blocks and tool cards above and below, chat switching, the jump-to-latest button, editing a long unfolded turn, and compact mode. Known edge: with the window at its minimum height a folded bubble can be taller than the viewport, and then "keep the control where it is" leaves its top just above the edge.

Rules and the story are in `app/CLAUDE.md` and `docs/gotchas/app.md`.


ddalcu:

- `app/Sources/MLXServe/Views/ChatView.swift:1832` — `.transcriptShown` no longer returns `.toBottom`, but the `scrollPosition` `@State` that used to be reset by that jump lives on `ChatDetailView` (reused across tabs) while the new `.id("transcript-\(sessionId)-…")` only recreates the `ScrollView` beneath it — so a programmatic offset set in one conversation is still in the binding when the next one attaches.
  Failure scenario: read history in chat A and unfold a long turn (`.toOffset` → `scrollTo(y: 3000)`), then switch to chat B. The fresh scroll view adopts the retained position instead of `.defaultScrollAnchor(.bottom, for: .initialOffset)`, so B opens mid-transcript — and since `.transcriptShown` sets `isPinnedToBottom = true` and returns `.none`, nothing corrects it and the jump-to-end button stays hidden. This is exactly the case the deleted `testSwitchingSessionsReturnsToTheBottomAndRe_engages` assertion covered; its replacement no longer checks the returned action.
- `app/Sources/MLXServe/Views/ChatView.swift:2407` — the transcript's `LazyVStack` becomes a plain `VStack` while `rows` stays uncapped (`ChatRowBuilder.rows(from: msgs)` at `:2889` builds every message).
  Failure scenario: opening a several-hundred-message conversation now builds and lays out every row — including every code block's syntax highlighting and markdown measurement — before the first frame, on the main thread. What used to cost one screen of rows now scales with conversation length, so a long chat hangs on open and on every metrics change that rebuilds the stack. The blank-transcript fix needs the real content height, but nothing here bounds what that costs.

ddalcu:

Re-reviewed the increment `6626e9c3..f8b544fc`. Both findings from the previous head are closed: `scrollPosition` is re-seeded to a fresh bottom-anchored `ScrollPosition` in `.onChange(of: sessionId)` (`ChatView.swift:2919`), so a programmatic offset can no longer cross conversations, and `TranscriptWindow` caps the eager `VStack` at 150 rows behind "Show earlier messages". The rename to `rowWillResize`/`rowDidResize` reads correctly in both directions — `max(0, s.offsetY - (s.contentHeight - h))` grows the offset by exactly what appeared above when `h > s.contentHeight` — and the reveal is correctly a no-op while pinned, where the bottom size-change anchor already holds the end.

- `app/Sources/MLXServe/Views/ChatView.swift:2900` — the "is this a different chat?" test lives in the *messages* observer and is the only writer of `windowSession`, so a switch to a session whose `messages` array compares equal never fires it and leaves `windowSession` naming the conversation the reader just left.
  Failure scenario: fork a chat of more than 150 rows at its last message — `ChatFork.prefix` returns `Array(messages[...cut])` with the boundary already satisfied and the ids kept, so the fork's `messages` is `==` the source's. Open the source, click "Show earlier messages" (`firstVisibleRow = 0`), switch to the fork: `sessionId` changes but `session?.messages` does not, so `:2898` never runs and the fork renders with the source's cut. The first streamed token then fires the observer, `windowSession` still holds the source's id, the `!=` branch at `:2900` wins, and the transcript silently collapses to its last 150 rows mid-reply. The comment at `:1878` rests on the messages observer firing first on a switch; the case it does not cover is it not firing at all. `.onChange(of: sessionId)` at `:2907` always fires and already re-seeds `scrollPosition` there — cutting the window in the same place (after `rows` is rebuilt) removes the ordering assumption entirely.

lojza3d:
Both points were worth measuring. Two commits on top:

**1. The retained position binding.** Right premise: `scrollPosition` outlives the scroll view, so a point a fold set in one conversation was what the next one attached to. In practice the follow correction in `geometryChanged` caught it (pinned + idle + far from the end = jump to the end), which is why it never showed, but a safety net is not the mechanism. 4ed3357 hands the switched conversation a fresh `ScrollPosition` beside the fresh scroll view, and the switching test asserts the action again.

**2. The cost of exact heights.** I instrumented both builds with the same checkpoint (time from the switch to the transcript's content height settling, 500 ms of quiet) over the same `chat-history.json`, on an M-series Mac. Three real chats plus a synthetic one with 500 turns (250 markdown replies, 200 thinking blocks).

| chat | main (lazy) | this branch, whole | this branch, windowed |
|---|---|---|---|
| 14 messages, two long user turns (21k + 8k characters) | 0.25 s | 0.92 s* | 0.26 s |
| 54 messages, one 130k-character user turn | 0.08 s | 0.17 s | 0.21 s |
| 569 messages, agent run (tool cards) | 0.15 s | 0.47 s | 0.21 s |
| 500 turns, 250 markdown replies | 0.05 s | 2.51 s | 0.65 s |

\* opened from the 500-turn chat; most of it is tearing that one down. From a short chat it is 0.26 s in every column.

The lazy column needs a footnote too. Its cost is not the number of messages but the number of rows it realizes: a short chat realizes all of them, so the first row pays for a 21k-character `Text` laid out whole (no fold on main). Its 0.05 s on the 500-turn chat is the time to an estimate: three height reports and then nothing, which is the blank transcript. Once the reader scrolled, the height changed 63 more times, and on the chat with the 130k-character turn 918 times in one stretch (7 to 9 s until it stopped). That churn is the scroll jank, and it is what every anchor and `scrollTo` was aiming at.

So the cost of exact heights is real and it is per rendered reply (~10 ms each; tool cards are cheap, which is why 569 messages open faster than 500 turns), and it is now bounded: f8b544f lays out the last `TranscriptWindow.rowLimit` (150) rows with a "Show earlier messages" pill above them. The cut is an index fixed when the chat opens, never "the last N", so a streaming reply appends rows without moving the oldest visible one. Revealing the rest adds height above the reader, which is the fold's problem with the opposite sign, so it rides the same bracket (`rowWillResize` / `rowDidResize`, renamed now that they carry both directions).

A metrics change still rebuilds every visible row, as before; it rebuilds 150 of them now.

S screenshot of "Show earlier messages" pill in 500-messages synthetic chat:
show_earlier_messages

lojza3d:
> `app/Sources/MLXServe/Views/ChatView.swift:2900` — the "is this a different chat?" test lives in the _messages_ observer and is the only writer of `windowSession`, so a switch to a session whose `messages` array compares equal never fires it and leaves `windowSession` naming the conversation the reader just left.
>   Failure scenario: fork a chat of more than 150 rows at its last message — `ChatFork.prefix` returns `Array(messages[...cut])` with the boundary already satisfied and the ids kept, so the fork's `messages` is `==` the source's. Open the source, click "Show earlier messages" (`firstVisibleRow = 0`), switch to the fork: `sessionId` changes but `session?.messages` does not, so `:2898` never runs and the fork renders with the source's cut. The first streamed token then fires the observer, `windowSession` still holds the source's id, the `!=` branch at `:2900` wins, and the transcript silently collapses to its last 150 rows mid-reply. The comment at `:1878` rests on the messages observer firing first on a switch; the case it does not cover is it not firing at all. `.onChange(of: sessionId)` at `:2907` always fires and already re-seeds `scrollPosition` there — cutting the window in the same place (after `rows` is rebuilt) removes the ordering assumption entirely.

Good catch, and exactly the shape: a fork keeps its source's ids, so the messages observer has nothing to fire on. The session observer cuts the window now as well, in the same place it re-seeds the position binding, so the ordering assumption is gone rather than documented.

On CI: the failing job is the Zig step (14 `transformer.zig` NAX kernel tests and one in `mlx.zig` on the `macos-26-arm64` runner, identical across both runs), and the Swift build and test steps after it never start. This PR touches nothing under `src/`.

ddalcu:
🤖 **MLXServe-PRBot approved** — reviewed at head `9b7c6f8d`, no issues found.

`cutTranscriptWindow()` from `.onChange(of: sessionId)` closes the finding: that observer always fires on a switch, so the equal-messages case (a fork keeps its source's ids, so `session?.messages` never changes and `:2899` never runs) now cuts instead of inheriting the previous conversation's window, and the second cut on an ordinary switch is idempotent — same `rows`, same count. The first-appearance path still belongs to the messages observer via `initial: true`, and a switch to a deleted session lands on `rows == []` on both paths.

One thing worth knowing rather than a defect: this now leans on the ordering the comment at `:2923` states. If the messages observer ever ran second, it would find `windowSession == sessionId` (just set) and take the `clamp` branch against the *previous* chat's cut — switching from a 10-row chat to a 500-row one would keep `firstVisibleRow == 0` and lay out all 500 rows, the hang the window exists to prevent. The previous shape was order-independent for that case and broken only for the fork case, so this is a real trade. Recomputing `rows` inside the `sessionId` handler before cutting would drop the assumption entirely.

lojza3d:
> One thing worth knowing rather than a defect: this now leans on the ordering the comment at `:2923` states. If the messages observer ever ran second, it would find `windowSession == sessionId` (just set) and take the `clamp` branch against the _previous_ chat's cut — switching from a 10-row chat to a 500-row one would keep `firstVisibleRow == 0` and lay out all 500 rows, the hang the window exists to prevent. The previous shape was order-independent for that case and broken only for the fork case, so this is a real trade. Recomputing `rows` inside the `sessionId` handler before cutting would drop the assumption entirely.

Taken: the session observer rebuilds the rows before it cuts, so the order the two observers run in no longer matters (ac39f8b).

ddalcu:
🤖 **MLXServe-PRBot approved** — reviewed at head `ac39f8be`, no issues found.

Rebuilding `rows` in the session observer before `cutTranscriptWindow()` removes the ordering assumption I flagged. Both interleavings now land on the same state: session-first sets `rows` and cuts, after which the messages observer sees `windowSession == sessionId` and its `clamp(firstRow(n), n)` is a no-op (`firstRow(n) < n` for every `n > 0`, and `n == 0` falls back to 0); messages-first cuts and the session observer then rebuilds identical rows and cuts again. The fork case still reaches only the session observer, which now has its own copy of the rows rather than the previous conversation's. `session` is read through the updated `sessionId` here, the same way `syncTogglesFromSession()` and `restoreAttachedFolderIfNeeded()` already do two lines above. The cost is one extra `ChatRowBuilder.rows` pass per switch — grouping, not layout — which is the right side of the trade against a 500-row eager render.

ddalcu:
thanks !

PR #420, opened 2026-09-14
perf(mtp): dispatch the grouped draft chain before the verify build

beamivalice:
Follow-up to #412. The grouped MTP path built its draft chain but left it lazy, so the chain ran only when the PLE flush waited on it, after the verify graph was built. One async eval over every row's draft graphs (`mtpChainDispatchBatched`) now fires before the verify build, so the chain's GPU time covers the 4 ms build. The flush wait falls from about 7 ms to 0.7 ms per round. Drafts are unchanged: a lazy sampling op binds its PRNG key at graph build, so only the dispatch time moves. No new switch; the group planner gate from #412 still covers the whole path.

**Measured** on M5 Max, Qwen3.8-Flash-Next mixed-4/8-bit, affine-8 KV, forced depth 2, 300 tokens, warm contexts, four counterbalanced boot pairs (two per implementer and reviewer), aggregate tok/s at four streams and 128k:

| Pair | Before | After | Ratio |
|---|---:|---:|---:|
| A | 132.2 | 135.7 | 1.026 |
| B | 131.4 | 135.7 | 1.033 |
| C | 133.2 | 134.6 | 1.011 |
| D | 132.9 | 133.9 | 1.008 |

16k at four streams +0.1% / +1.4%; one stream unchanged. Byte-identical to #412's head in 14 of 14 greedy cells (forced depth 2, coding, prose and counting prompts, 4k and 64k, one and four streams).

**Warm ladder** with this change, auto planner, `--kv-quant 8 --mtp --mtp-head-kv-quant --max-concurrent 4 --ctx-size 400000`, two runs per cell, every request served from a cached prefix at its full context, aggregate tok/s:

| Context | 1 stream | 2 streams | 4 streams |
|---|---:|---:|---:|
| 4k | 96.9 / 106.4 | 106.1 / 106.9 | 127.9 / 132.8 |
| 16k | 111.7 / 107.1 | 119.8 / 127.4 | 147.7 / 148.4 |
| 64k | 106.4 / 106.1 | 118.6 / 120.8 | 137.6 / 128.4 |
| 128k | 102.0 / 96.9 | 109.8 / 111.3 | 120.7 / 122.0 |

Gap p95: 35 to 42 ms at one stream, 42 to 47 at two, 69 to 79 at four. The 1-stream column was measured on a local tree that adds single-stream verify kernels not in this PR; the 2- and 4-stream columns are this PR's path.

**Meter.** The `[mtp-trace]` line gains sub-laps (`build`, `ple`, `ple_sync`, `dispatch`, `wait`, `chain`) behind `acc_idx=`, so existing consumers still parse the phase summary as one run. They are how the round was priced, and what they show is worth recording: the inference thread encodes the tape inside `mlx_async_eval` for the whole GPU window (`eval_impl` builds and runs the tape on the calling thread and throttles at `MAX_ACTIVE_TASKS`), so the post-dispatch shadow is 0 ms for a grouped round and there is no CPU-idle window left in which to prebuild the next round.

**Verification** on this tree: unit suite 2525 passed, 0 failed, 169 skipped, `zig build test` 9/9 steps; the live two-stream EOS/cancellation parity gate `tests/test_mtp_batch_stops.py` passes on a fresh server booted with forced depth 4.


ddalcu:

First review, at head `fd31d82a`. The dispatch itself is sound: `mtpChainDispatchBatched` is a strict superset-safe version of the solo `mtpChainDispatch` — `draft_arrs[0..n_drafted]` is exactly the documented valid range, `conf_arrs[0..n_conf]` + `h_chain` matches the solo `from == 0` arm, and clamping `q_probs` with `@min(to, n_qp)` is tighter than the solo `slots[from..to]`, which can reach past `n_qp`. The byte-identity claim holds structurally: `mtpChainBuildBatched` binds every sampling op's PRNG key at build, and the new call only evaluates. On the error path `n_chain` is still `n`, so the existing `defer for (chains[0..n_chain])` frees the chains. `verify_laps` is inference-thread-only so the shared field cannot race, the sub-laps are documented as non-partitioning, and `global_single_threaded.io()` matches `acestep.zig`'s existing use.

- `src/scheduler.zig:7735` — the grouped dispatch is unconditional while both solo call sites (`generate.zig:6289`, `:6354`) gate the identical behaviour on `mtpEarlyDispatchEnabled()`, so `MLX_SERVE_MTP_EARLY_DISPATCH=0` no longer turns early dispatch off everywhere.
  Failure scenario: someone chasing a suspected async-eval interaction in a grouped round sets `MLX_SERVE_MTP_EARLY_DISPATCH=0`, watches the solo path revert to late dispatch, and concludes the engine is on the late arm — while every grouped round still dispatches before the verify build. The lever silently stops covering a path, which is the class `CLAUDE.md` names twice ("a guard that shapes INIT options does not bind DISPATCH"; the batched-decode gate that read ARMED flags), and the MTP head rule asks that every spec lever be kill-switched. The PR body's "the group planner gate from #412 still covers the whole path" is true of the planner but not of this lever: `MLX_SERVE_MTP_GROUP_PLANNER=0` turns off grouping altogether rather than isolating the dispatch change, so there is no arm that keeps grouping and moves the dispatch back. Wrapping the call in the same `if (mtpEarlyDispatchEnabled())` restores it in one line and keeps the A/B the solo path already offers.

PR #421, opened 2026-09-14
perf(mtp): compile the verify kernels at load and re-validate the round-cost table

beamivalice:
Follow-up to #412 and #420. A cold boot can leave the MTP planner stuck at one draft per round for the whole process, and this closes both halves of why.

**Symptom.** Fresh server, `npx llmprobe --bench-only` as the first traffic: 91.8 tok/s decode, speculative ratio 1.21x (predictable 95 vs plain 92), and the round-cost table reading `w1:11.45/994, w2:37.74/4` for the whole run. The same binary warmed by three 300-token requests first: 115 tok/s, 2.11x, `w2:10.05/24, w3:9.40/35, w4:8.97/45`.

**Cause.** The first live round at each draft width pays that width's Metal kernel compiles inside a measured round: 139 to 148 ms against a 25 to 31 ms steady round, reproducible on every cold boot of main. `Transformer.warmup` only compiles the width-1 decode and the 8-token prefill. The table's step bound refuses a sample more than 1.5x per step above the nearest *trusted* narrower cell, but on a cold boot the first wide rounds fold before width 1 has its three samples, so the compile-laden sample becomes the cell, reaches three samples, and is trusted. Nothing re-reads a cell once its narrower neighbour matures: the implausibility sweep runs only at load, the self-spike guard compares a cell with itself, and width trials never revisit a width the planner already believes is 3x too dear. The same hole shows on a healthy boot as a stale `w3` cell at 1.8 to 2.1x `w2` for 325 samples with the drop counter at zero.

**Fix.**
- `Transformer.warmupSpecVerify`, run from the eager warm-up when the model has an MTP head and MTP is on: the solo verify at widths 1 to the depth cap (rows 2 to 7), the grouped row-axis verify at every row total the scheduler can sub-group into (2, 3 and 4 slots on the NAX lane at the token counts `mtpSubGroupSize` hands them), and the head's own step in both projections. Throw-away slots with their own KV, SSM and context; the model's state and the round-cost table see nothing. One log line with the widths and the time. No new environment variable.
- `Table.dropWiderAgainst`: when a fold takes a cell across the trusted threshold, the wider cells of that bucket are re-checked against it with the existing step bound and dropped if they fail, counted in `table_drops` `i`. Unit test, red first: four width-2 samples at 37 ms folded before width 1, then width 1 trusted at 11 ms, must clear the width-2 cell; the same at 14 ms must keep it.

**Measured**, M5 Max, Qwen3.8-Flash-Next mixed-4/8-bit, the owner's serving flags, no request before llmprobe, three base and four fixed cold boots:

| | first MTP round | decode tok/s | spec ratio | ~4.3k tok/s | ~16.3k tok/s |
|---|---:|---:|---:|---:|---:|
| main | 139 to 148 ms | 106 to 110 | 2.07 to 2.30 | 91 to 96 | 97 to 99 |
| fixed | 25 ms | 108 to 110 | 1.91 to 2.61 | 92 to 120 | 100 to 107 |

The first-round cost is the direct measurement of the compiles leaving the table's samples. The 92 tok/s pathology itself is a race the boot usually wins: it did not recur in three base boots tonight, which is why the table half exists as the backstop and is pinned by the unit test rather than a live number. Load time +1.1 to 1.4 s (`[spec-warmup] ... (1102-1359 ms)`).

**Verification.** Greedy output byte-identical to main at one stream and as a 4-stream group (forced depth 2, 4k, 300 tokens); the live two-stream EOS/cancellation gate `tests/test_mtp_batch_stops.py` passes; unit suite 2507 passed, 0 failed, 169 skipped, `zig build test` 9/9 steps.

Not warmed, deliberately: the long-context arms (QSA gather and the masked split attention at 4k+ keys) need a multi-thousand-token KV at load and still compile in the first round of their bucket.


ddalcu:

First review, at head `43b062f7`. The table half is correct: `foldInto` only ever steps `n` by one or reseeds it to 1, so `n == MIN_SAMPLES` fires exactly once per climb and the sweep cannot re-run on a matured cell; `dropWiderAgainst` clears to `.{}` so a dropped cell is untrusted for `measuredMs` and free to re-climb; and the `.implausible` early-return above `self.folded += 1` keeps a rejected sample from triggering it. The warm-up's handle discipline holds too — `rows` is fully initialised by infallible `mlx_array_new_data` calls before the first `try`, every `specWarmIds` shape matches its buffer exactly (the "COPIES shape-worth of bytes" rule), `head_in` outliving its freed parent is the same refcount pattern `qwen4MtpForwardBatched` already relies on, and the defer releases the SSM group before freeing the slots. `specWarmGroupSeq` also reproduces `mtpGroupRowCap() / size - 1` correctly for both lanes, and the inlined `naxLaneEnvEnabled() and verifyQmmNaxAvailable()` is today byte-for-byte what `dflash.wideVerifyLaneAvailable()` returns.

- `src/transformer.zig:17990` — `warmupSpecVerify` sizes its slot count from the verify lane alone and never asks how many slots the server can actually run, so a single-stream server compiles group shapes it can never dispatch.
  Failure scenario: `--max-concurrent 1`. `mtpSubGroupSize` returns `remaining` for `remaining < 2`, so every round is solo and no row-axis group is ever built — yet the warm-up still stands up 3 or 4 throw-away slots, forwards a prompt through each (allocating a KV cache plus `num_hidden_layers` conv/SSM states per slot, none of it billed by the preflight that already passed), and runs `specWarmVerify` across every group row total. That is the bulk of the reported 1.1–1.4 s and a transient the load-time estimator never saw, spent on pipelines the process cannot reach. `sch.max_concurrent` is in scope at the call site (`scheduler.zig:4009`), so clamping `max_rows` with it costs one argument.
- `src/transformer.zig:18033` — `specWarmGroupRows`/`specWarmGroupSeq` re-derive the scheduler's sub-group ladder inside the transformer, and the new test pins those copies against literals rather than against `mtpSubGroupSize`/`mtpGroupRowCap`.
  Failure scenario: change the ladder in `scheduler.zig:7870` — a new lane width, or allowing five slots at row cap 16 — and the warm-up keeps compiling 2/3/4-row shapes while the scheduler dispatches a 5-row one. `zig build test` stays green, because "spec warm-up covers exactly the group shapes the scheduler can build" asserts `specWarmGroupRows(16) == 4` against its own copy and never calls the function it claims to mirror. The first live 5-row round then pays its Metal compile inside a measured round and the round-cost table folds it as that width's price — the exact defect this PR exists to close, reintroduced for the new shape. Calling `mtpSubGroupSize`/`mtpGroupRowCap` (both already `pub`) or asserting the copies equal them would make the mirror checkable.

PR #422, opened 2026-09-14
perf(mtp): reuse exact verifier kernels for single streams

beamivalice:
On main after #421. Single-stream native MTP on Flash-Next reuses the exact verifier kernels that #412 introduced for the grouped path, so a lone stream at verify widths 2 to 6 stops paying the stock gather chain per round.

**What changes in a single-stream round** (the grouped verifier is untouched, other architectures and unsupported widths keep their paths):

1. Route packing with indexed gate/up inputs. The existing exact route kernel runs once per layer and the gate/up gathers take explicit LHS indices, so the inverse-sort setup, the expanded activation copy and the implicit index generation are gone.
2. Adjacent-expert pair reuse in the down projection plus the fused weighted reduction, with the stock bf16 reduction order preserved; the single-stream shared-expert add stays.
3. Prepared hyper-connection graphs reused across rounds, which removes host graph construction, not GPU work.

The selection gate admits Qwen4 speculative capture at batch 1 and widths 2 through 6; hardware, dtype, quantization and shape guards of the primitives stay authoritative. No new environment variable. The MTP cost prior gains a `g17_nax_qwen4_mixed_4_8_gs64` profile for this checkpoint's 8-bit head projections (the older `q4_gs64` selector correctly declined it and fell to the generic prior); its forced-depth calibration at ~8.6k context, three repetitions in two boot orders, fits a 19.0 ms floor with 4.98 ms per draft position through depth 5 and a 1.3% maximum residual.

**Exactness.** The native model gate compares logits, hidden outputs, recurrent state, PLE, QSA state and the logical speculative captures across widths 1 to 7 bit for bit at 4,101 and 65,541 prefix tokens, and asserts the five primitive counters advance only at eligible widths (48 route, 48 indexed-input, 48 down-reuse, 48 reduction and 96 prepared-HC calls per eligible width; none at widths 1 and 7, none during prefill). An HTTP comparison at forced depth 4 gave identical streamed output, reasoning content, token counts and finish reasons at 4,235, 16,431 and 65,637 prompt tokens. Unit suite on this branch: 2510 passed, 0 failed, 171 skipped, `zig build test` 9/9 steps.

**Measured**, M5 Max 128 GB, mixed-4/8-bit pack, affine-8 target and head KV, single stream, greedy, 300 generated tokens per request, warm prefix, auto planner, `MLX_SERVE_ROUND_COST_PERSIST=0`, fans at max, one boot per arm in each of two boot orders, two runs per cell:

| Context | main 1075630, four runs | this PR, four runs | median gain |
|---|---|---|---:|
| 4k | 92.2 / 95.6 / 98.4 / 99.5 | 101.1 / 102.1 / 105.0 / 104.8 | +6.6% |
| 16k | 90.4 / 94.5 / 102.8 / 101.1 | 109.4 / 102.6 / 99.7 / 105.1 | +6.2% |
| 64k | 91.6 / 94.6 / 101.2 / 98.6 | 102.0 / 104.2 / 100.4 / 98.2 | +4.8% |
| 128k | 83.3 / 84.1 / 88.0 / 80.4 | 89.9 / 86.3 / 97.7 / 90.8 | +7.9% |

Each cell lists boot order A then B, two runs each; every request restored its full context from the prefix cache. Ladder driver: 300 greedy tokens per request via `/v1/chat/completions` with `enable_mtp`; llmprobe reports for both arms are uploaded to llmprobe.deploy.dalcu.com under labels naming this PR.

Astra's original paired screen on the #412 base, three matched prompts per arm at ~4.2k input tokens: coding 92.98 → 101.20 tok/s (+8.8%), counting 118.34 → 127.59 (+7.8%). Full notes: `notes/MTP-SINGLE-VERIFY.md` in the owner's tree.


ddalcu:

First review, at head `cab7fcd8`. The gate reads correctly: `verifyFeatureEnabled` returns the old `skip_shared and qwen4 != null` answer unchanged on the grouped arm, the new single-stream arm is `spec_capture_ssm and batch == 1 and width 2..6` (so `m` 1..5, matching the new prior's `flat_max = 5`, with `m == 6` correctly falling outside), the `is_test` hook is comptime-eliminated in release, and the added `qwen4 != null` term on the HC path cannot change a non-qwen4 model since hyper-connections are qwen4-only. The prefill-tail hazard the comment names is genuinely covered — prefilling 4101 tokens in 4096-chunks leaves a 5-token width-5 batch-1 tail, and the exactness test asserts the route and HC counters do not move across it. The `Qwen4MixedGeometry` fingerprint and `qwen4MixedPackMatches` pin bits/group/dtype/shape per tensor, and the profile falls to `.generic` for any KV width other than affine-8, which the live test checks in both directions.

- `src/transformer.zig:60697` — `singleVerifyTestRound` builds `Generator` and `MtpRoundState` with `undefined` and sets three fields, but `mtpRoundVerify` already reads a fourth unconditionally: `generate.zig:6479` evaluates `self.timer.io` as a by-value argument to `SubLap.start` regardless of `tracing`.
  Failure scenario: the helper reads an uninitialised `std.Io` (0xAA-filled under the test build's safety mode) on every round. It survives today only because `tracing` is false so the value is stored and discarded without a dereference — but this is the gate that justifies shipping new kernels on the single-stream path, and the read it depends on not happening was added one PR ago (#420's sub-lap plumbing). The next unconditional field read in `mtpRoundVerify` — another lap, a counter, a cache handle — makes the exactness comparison run against garbage instead of failing loudly, and both arms would read the same garbage, so the byte-equality assertions would still pass. Initialising the two structs with the fields the call path needs (`.{ .xfm = xfm, .ctx = slot.ctx, .timer = ..., ... }`) costs nothing and makes the gate fail closed.

PR #423, opened 2026-09-14
perf(prefill): price an 8192 chunk on long-context-gated MoE instead of capping it

beamivalice:
On Flash-Next the prefill chunk is a per-request budget decision, but `boundedPrefillChunk` capped MoE hd-256 models at 4096 before the admission estimator could price the 8192 rung. That cap exists for small Macs (a gemma-26B at 99k once paid +22 GB of peak for +3%), and it stays for every architecture except the long-context-gated one (`ModelConfig.longCtxGated()`), where the 8192 rung is now offered and `chooseRequestPrefillChunk` takes it only when its bill fits, stepping down to 4096 and below exactly as it already does between the smaller rungs. The estimator already bills every chunk-scaled term at the chosen width (expert gather, GDN stream, MLP envelope, QSA sheet, KV reservation); a new test pins that and no term had to be added. One boolean threaded through `effectivePrefillChunk`, so guard and forward keep computing one chunk at all fourteen call sites. No new environment variable.

**Trade, base binary, 8192 forced through `MLX_SERVE_PREFILL_CHUNK` against the 4096 cap**, four counterbalanced boots, cold prefill, prefix cache off, M5 Max 128 GB, Qwen3.8-Flash-Next mixed-4/8-bit, owner's serving flags (`--ctx-size 400000 --kv-quant 8 --mtp --mtp-head-kv-quant ...`):

| Context | 4096 tok/s | 8192 tok/s | gain | peak active+cache, 4096 → 8192 |
|---|---:|---:|---:|---|
| 16k | 1713 | 1799 | +5.0% | 74.7 → 77.9 GB |
| 64k | 1706 | 1777 | +4.2% | 76.3 → 79.0 GB |
| 128k | 1644 | 1724 | +4.9% | 77.6 → 80.7 GB |
| 326k | 1589 | 1593 | +0.3% | 82.7 → 85.9 GB |

Recomputing the chunk-scaled terms at the geometry the pack actually ships (the unit fixture carries a wider expert, no shared expert and no GatedDeltaNet heads, so its bill is not the deployed one): at a 300k prompt, kv 8-bit, MTP on, the estimator bills **13,370 MiB at 4096 and 17,014 MiB at 8192**, a step of **3,643 MiB (3.82 GB)**; at 326,351 tokens it is 13,987 → 17,631 MiB, the same 3,643 MiB step (MTP off: 12,552 → 16,185 MiB and 13,099 → 16,732 MiB, step 3,633 MiB). So the bill for the step is about 0.5 GB **above** the widest measured active+cache step of the trade (+3.3 GB at 326k; +3.2 / +2.7 / +3.1 GB at 16k / 64k / 128k) — the rung is priced, not forced. The earlier "+3.1 GB at 300k" figure in this body was not reproducible from either config and is withdrawn.

**Margin on the rung the gate widened.** 8192 is a width only the gate offers, so the machine cap frozen at load no longer bounds that forward, and `CLAUDE.md` records that `prefillMemoryNeeded` UNDER-bills quantized prefill by about 22% on a path where a Metal OOM is uncatchable. `chooseRequestPrefillChunk` therefore takes a rung wider than its own ungated width only when `bill *| 122 / 100 <= available` (`PREFILL_WIDE_RUNG_MARGIN_PCT`, saturating); every narrower rung keeps the plain `bill <= available`, so nothing else in admission moves, and the guard still bills the width the chooser returned. On this box that puts the 8192 rung at a 300k prompt at about **20,757 MiB (21.8 GB) of post-eviction headroom**, against the ~38 GB available under the ceiling, so it is still taken here; below that threshold the request steps down to 4096, which bills 13,370 MiB (14.0 GB).

**Fixed binary against base, auto chunk on both**, four boots in both orders:

| Context | base | fixed | gain |
|---|---:|---:|---:|
| 16k | 1599 | 1834 | +14.7% |
| 64k | 1596 | 1793 | +12.3% |
| 128k | 1557 | 1728 | +11.0% |
| 350k | 1440 | 1625 | +12.8% |

Session drift on this box was large (up to 13% between a run's first and last boot); combining the two mirror-image designs above cancels the order term at **+9.7% / +8.2% / +7.9%** for 16k / 64k / 128k, so the honest figure is about +8% with a +5% to +15% bracket. A 350,464-token prompt at `--ctx-size 400000` is admitted and completes on both arms with about 23 GB left under the ceiling (`[admission] ... width=8192 verdict=admit`); at a 1M-token bill the chooser steps back to 4096.

**Correctness.** A chunk change moves GEMM shapes, so the bar is the chunking class, not byte identity: a planted needle at 40% depth is recovered by both widths at 16k and 64k; at 16k the first 30 greedy tokens are byte-identical, at 64k the arms diverge at a near-tie (top-2 gap 0.125 nats, the ≤ 0.15 bar `tests/test_mtp_equivalence.sh` uses). The QSA gather and select, the fused QSA prefill arm, the GDN kernels and the MoE gather engage at 8192 (log lines); the dequant GEMM and the blocked GDN prefill kernel, which log nothing, were proven live by kill-switch A/B at 64k (−3.4% and −5.1% when disabled). One concurrent 300-token stream during a 64k cold prefill: worst gap 2.06 s at 4096 against 2.22 s at 8192 (p95 83 → 92 ms), so halving the yield points costs +10% on the co-tenant's worst gap, not 2x.

**Verification.** Four red-first unit tests (the gated cap, the chooser's wide-rung margin at the boundary, the wider bill's chunk-scaled growth, the deployed pack's 4096 → 8192 step against the measured peak step); `zig build test` 9/9 steps, 2510 passed, 0 failed, 169 skipped.

Open: the opt-in `--ane-prefill` tile is compiled at the pinned width, which on this arch can now be 8192 while a request forwards narrower; it was already loose once the width became per-request, and it was off in every cell here.


ddalcu:

First review, at head `4f20393b`. The gating is the sanctioned one — `CLAUDE.md` lists "chunk-sizer changes" among the mechanisms `ModelConfig.longCtxGated()` owns, and the predicate is threaded rather than hand-rolled at the site. Because the new parameter is positional, every one of the call sites is compiler-enforced, so none can drift; `computeMemoryContext` still short-circuits on `total_ctx == 0` and is unaffected; `prefillAdmissionBill` passes the same flag, so the guard and the real prefill keep computing one chunk; and the rewritten `adaptivePrefillWidth` tests (`2048` → `cap / 2`) are a necessary adaptation to a cap that may now be 8192, not a weakened assertion. The "ordinary prompt buys the wide chunk a 1M session cannot" test still separates the cases (8192 vs 4096 at 1M) and keeps its monotonicity sweep.

- `src/generate.zig:171` — lifting the MoE 4096 cap for the gated arch removes the last measurement-backed ceiling on the expert-gather transient, and for exactly this arch the load-time pin does not apply, so `prefillNeededAtChunk` becomes the sole gate on an 8192-wide forward.
  Failure scenario: `chooseRequestPrefillChunk` returns `load_time_pin` only for an ungated arch (`server.zig:5444`); for a gated one it walks the ladder and takes the widest rung whose bill fits (`:5460`), which is why the test above asserts `4096 > pin` while the chooser still answers 8192. That is the documented per-request design, but it means the machine cap frozen at load no longer bounds this width — and `CLAUDE.md` records that `prefillMemoryNeeded` UNDER-bills quantized prefill by about 22%, on a serving path where Metal OOM is uncatchable. On a box whose `available` sits just above the 8192 bill, the estimator admits the wide chunk, the real peak overshoots, and the server dies instead of refusing by name through `PrefillDoesNotFit`. The PR's own step measurement does not clear this: it reports the bill for the step at +3.1 GB against a measured +3.2 GB at 300k, so the bill is already under the delta on the author's own hardware, while the prose reads it as "at least the measured step". A floor under the margin — refusing the widest rung unless its bill clears `available` by the estimator's known error, or keeping the 4096 cap for anything below a measured memory headroom — would make the trade survive a box the ladder was not measured on.

ddalcu:
🤖 **MLXServe-PRBot approved** — reviewed at head `4d759596`, no issues found.

The finding is closed on both halves. `PREFILL_WIDE_RUNG_MARGIN_PCT = 22` is applied only where `width > rungWidth(config, seq, rung, false)` — i.e. only to a rung the long-context gate widened — so an ungated arch and every narrower rung of a gated one keep the plain bill and no other admission decision moves. The arithmetic is safe: `bill *| (100 + 22) / 100` saturates rather than wraps, and `*|` and `/` share precedence left-to-right so it multiplies before dividing. The direction against `prefillAdmissionBill` is also right — the chooser is now strictly stricter than the guard that bills the chosen width, so the two cannot drift into the guard refusing what the chooser picked.

The second half answers the number I flagged: `qwen4DeployedTestConfig` pins the served pack's real MoE and GatedDeltaNet geometry (the fixture left the linear head counts at zero, and both chunk-scaled terms read them), and the new test asserts `wide - narrow >= 3.3 GB` at two long contexts with MTP on and off — so the bill for the 4096→8192 step is now pinned at or above the measured peak step rather than 0.1 GB under it. `rungWidth` is a clean extraction with the gate flag parameterised, and the boundary test was updated to expect the margin on exactly the widened rung.

PR #426, opened 2026-09-14
Give an interrupted turn a footer, and make the trash delete the turn

lojza3d:
### What the reader saw

Stop a turn while the model is thinking, or right after a tool result, and the transcript ends on a bare thinking block or a tool card. No time, no Regenerate, nothing to delete it with.

The trash under a reply removed that one message. Under an agent turn (prose, six tool rounds, prose) it left the tool rounds and their hidden results standing on their own, and the model read them back as a call with no answer.

### What changes

- `ChatTurn` (new, `Models/ChatTurn.swift`) is the one definition of a turn. `isBoundary` moves out of `ChatFork`: a message the model could be handed back, so the reader's message, or a reply with prose and no tool calls, no error card, no summary. `ChatFork` reads it from there.
- The trash sits only on a footer where the transcript can be cut: a boundary, or the last message (`footerDeletes`). It removes everything back to the previous boundary, which stays (`deletionRange`); attachments the dropped messages owned go through the same `removablePaths` sweep as `deleteMessage`. The context menu says "Delete Turn" on those rows.
- A thinking-only or tool-only row has no trash and no Delete in its menu, on purpose: cutting there would orphan its own card. The nearest trash below takes it.
- When the transcript ends on the model's side without a footer (`needsEndFooter`), `TurnEndFooter` is drawn under the last row: time, Regenerate (the existing `regenerateLastResponse`), trash. Never while the turn is in flight, its end moves.
- Deleting above the reader is bracketed like a fold (`rowWillResize` / `rowDidResize`), so the reader's place holds.

### Out of scope

- An interrupted turn in the middle of a chat (followed by a later user message) still has no footer. Same predicate at each boundary; a follow-up.
- Nudges stay user-role boundaries.

### Tests

`ChatTurnTests` (15 cases: boundaries, deletion ranges incl. an interrupted turn and prose that carries tool calls, `needsEndFooter`, `footerDeletes`, `hasOwnFooter`). `AttachmentStoreTests` now pins three removal sites. `bash app/test.sh`: 3386 tests, 0 failures. Manually: seven scenarios (stop while thinking, stop after a tool result, error card, trash under prose, trash under a tool-heavy turn, regenerate from the end footer, delete above the reader's place).

Docs: rule in `app/CLAUDE.md` (replaces "there is no turn in this app"), story in `docs/gotchas/app.md`.

### Screenshots

Before (cut after thinking block / tool call, no footer):
screenshot_3641

After (the same chat, footer available to delete th
screenshot_3642
e incomplete turn):

Before (cut turn, full footer after each prose block, no final footer):
screenshot_3644

After (cut turn, delete only available at the end):
screenshot_3645


ddalcu:

The turn model reads correctly. `deletionRange` walks back only while the row *above* is not a boundary, so the reader's question always survives and a later segment stops under the previous reply; `deleteTurn` captures `dropped` before `removeSubrange` and asks `removablePaths(orphanedBy:)` after, which is the order that makes "is this path still named anywhere" true; and the removal-site pin in `AttachmentStoreTests` was raised from two to three with the third named rather than loosened. `deleteTurn` is unreachable from a user row — those keep `deleteMessage` on both the footer and the menu. An error-notice message carries `content: ""` (`ChatTurnEngine.swift:360`), so `hasOwnFooter` is false for it and `needsEndFooter` does draw the footer under an error card, which is the case the description promises.

- `app/Sources/MLXServe/Models/ChatTurn.swift:27` — `isBoundary` is described as moved out of `ChatFork`, but the assistant arm gained a `!message.content.isEmpty` conjunct the original did not have, and `ChatFork.prefix` now reads the widened version.
  Failure scenario: an assistant row with empty content that is otherwise clean — a round stopped while thinking, or the media row the agent loop appends for a generated image (`ChatTurnEngine.swift:1339` constructs it as `content: ""` with only `media` set) — used to be a fork boundary and is no longer one. "Branch Chat From Here" on a generated image now trims that row away and cuts at the reply above, so the branch silently loses the picture the reader forked on; the same click before this change kept it. Whether that is the better rule is arguable, but it is a change to a different feature made inside a move: `ChatForkTests.swift` is untouched by this PR and its only empty-content case (`:87`) is a `failedRetry` failure the old predicate already excluded, and the new `ChatTurnTests` never mentions `ChatFork` — so nothing covers the difference in either direction. Either keep `ChatFork` on the old predicate, or state the widening and pin the fork cases it moves.

lojza3d:
> `app/Sources/MLXServe/Models/ChatTurn.swift:27` — `isBoundary` is described as moved out of `ChatFork`, but the assistant arm gained a `!message.content.isEmpty` conjunct the original did not have, and `ChatFork.prefix` now reads the widened version.
>   Failure scenario: an assistant row with empty content that is otherwise clean — a round stopped while thinking, or the media row the agent loop appends for a generated image (`ChatTurnEngine.swift:1339` constructs it as `content: ""` with only `media` set) — used to be a fork boundary and is no longer one. "Branch Chat From Here" on a generated image now trims that row away and cuts at the reply above, so the branch silently loses the picture the reader forked on; the same click before this change kept it. Whether that is the better rule is arguable, but it is a change to a different feature made inside a move: `ChatForkTests.swift` is untouched by this PR and its only empty-content case (`:87`) is a `failedRetry` failure the old predicate already excluded, and the new `ChatTurnTests` never mentions `ChatFork` — so nothing covers the difference in either direction. Either keep `ChatFork` on the old predicate, or state the widening and pin the fork cases it moves.

Right, and it was a real regression: the agent loop's generated-image row is `content: ""` too, so a fork made on the picture lost the picture and its whole round. Fixed in 9599a7e: a row carrying media is a boundary (the picture is that round's answer, delivered by a file). The narrowing for thinking-only rows stays, stated now: empty content never reaches the model (`buildAgentHistory` drops it), and a thinking-only row does occur mid-turn when the loop nudges the model on, so as a boundary it would strand a thinking block above every deletion below it. Both directions are pinned in `ChatForkTests` (fork on a generated image keeps the row and its round; fork on a thinking-only row trims to the reply above), the media case in `ChatTurnTests`, and the rule in `app/CLAUDE.md` names the narrowing.

ddalcu:
🤖 **MLXServe-PRBot approved** — reviewed at head `9599a7ee`, no issues found.

The finding is closed the right way round: a media-carrying row is a boundary again (`saidSomething` admits `!(media ?? []).isEmpty`, which also handles the nil case), so forking on a generated picture keeps it and its round; the thinking-only narrowing stays but is now stated in both `app/CLAUDE.md` and the gotcha rather than hiding inside a move; and both directions are pinned in `ChatForkTests` — the consumer that actually changed — not only in `ChatTurnTests`.

I checked the two predicates cannot now disagree visibly. `hasOwnFooter` still requires non-empty content, so a media row is a boundary that draws no footer of its own — and because `needsEndFooter` is exactly `!hasOwnFooter(last)`, a media row at the end gets the `TurnEndFooter` and never a second one beside it. Away from the end it simply has no footer, so `footerDeletes` is never consulted there, while the context menu it does carry offers a coherent "Delete Turn" that stops at the boundary above. `deletionRange` stopping at a media row is the same rule applied consistently: a picture ends a segment, so the trash under a later reply no longer reaches back past it.

lojza3d:
Follow-up in https://github.com/ddalcu/mlx-serve/pull/426/commits/4773f4495c29b68136eba2f491a3455915a7056e: the approve above noted that a media row is a boundary that draws no footer, which as a reader means a trash you cannot see (mid-history the picture had Delete Turn only in the context menu). The footer rule now reads the same "said something" as the boundary rule, so a generated picture gets time, Regenerate and the trash like prose does; Copy and Edit hide wherever there is no text to act on. Tested live with a generated image: delete under the caption keeps the picture, delete under the picture takes its tool round back to the question, fork on the picture keeps it. The generated file itself stays on disk, which is by design (removablePaths only ever removes user attachments).

screenshot_3646

ddalcu:
🤖 **MLXServe-PRBot approved** — reviewed at head `4773f449`, no issues found.

This closes the asymmetry I noted in passing at `9599a7ee`: `hasOwnFooter` now shares `saidSomething` with `isBoundary`, so a generated-picture row is a boundary that also draws its own footer instead of being a boundary with no controls on it. The two predicates stay mutually exclusive where it matters — `needsEndFooter` is still `!hasOwnFooter(last)`, so a picture at the end gets its own footer and no `TurnEndFooter` beside it, which the new `testAPictureAtTheEndIsItsOwnEnd` pins directly.

Hiding Copy and Edit on empty content is exactly scoped: `hasOwnFooter && content.isEmpty` holds only when `media` is non-empty, since `saidSomething` is the disjunction of the two, so no prose row loses a control. It also retires two dead controls that predate this PR — Copy on a thinking-only row and on an attachment-only user message both copied the empty string — which is the right side of the dead-control rule. The footer and the context menu got the same guard, so the two surfaces cannot disagree, and `app/CLAUDE.md` now states what a picture row's footer carries and what it deliberately does not.

Issue #419, opened 2026-09-14
Qwen3.8 Flash Next: port MTPLX acceptance work (typical, cascade, block verify, seeded correction) — engine diff and plan


davidtai:
### Before opening

- [x] I searched open and closed issues, open PRs and recent commits for this, and it is not already reported or fixed. Nearest threads: #338 (Qwen4 MTP sidecar loading, relevant only to the Optimized-Speed pack, not to this proposal), #129 (attribution for the MTPLX verify kernels already in the tree), #111.

### Problem

MTPLX's Qwen3.8 Flash-Next serving stack (youssofal/MTPLX PR 391, now on MTPLX main, plus the open PRs youssofal/MTPLX#475, #478, #482, #485, #488) reaches 82.8 tok/s decode exact, 104.7 with typical acceptance and 107.1 with speculative-cascade acceptance at a 16,384-token-prompt / 1,024-token cell on an M5 Max (temperature 1, top-p 0.95, top-k 20, Optimized-Speed pack). I want the transferable parts of that work in mlx-serve, serving `ddalcu/Qwen3.8-Flash-Next-MLX-Serve-mixed-4-8bit`.

Before writing anything I did an exhaustive engine-versus-engine comparison of the `qwen4_exp` path (kernels, verify round, acceptance, draft head, long context, prefill, memory) so the port list is only what mlx-serve does not already have. This issue is that comparison and the resulting plan. Nothing here has been built or run yet; the analysis was done by Claude Code (Fable 5.1) working with me from the two source trees, and I will drive the measurements myself.

### What mlx-serve already has (no port needed)

Of the 14 distinct mechanisms in the five MTPLX PRs, nine are already present or superseded in mlx-serve at 6d3cb6d, and three are same-purpose-different-design items:

| MTPLX mechanism | mlx-serve equivalent | Verdict |
|---|---|---|
| Hyper-connection read at verify width 4 (#475; fused norm/down/up over the bf16 mixers) | `hcReadFused` N/D/U kernels, row-batched to 16 rows, deferred write (`HcPending`), 8-bit prepared verifier graphs (`hcReadPrepared`, `hcReadJoined`) | present |
| QSA split-K sparse decode (#475; native two-pass flash over the 2,051 selected keys, dense bf16 K/V) | `msv_attn_qsa256_q` split-K under kv-quant, `qsaVerifyGatherAttn` union gather on dense KV, `qsaDecodeGatherAttn` at S=1 | present |
[...]

Issue #415, opened 2026-09-13
GDN recurrent state saturates to the bf16 ceiling and NaN at long context (qwen4_exp)


titan550:
### Before opening

- [x] I searched open and closed issues, open PRs and recent commits for this, and it is not already reported or fixed

### Priority

high (blocks my use of mlx-serve)

### Area

Server / API (chat, completions, messages, responses, embeddings)

### mlx-serve version

26.9.2-dev (build from commit ecf8d35)

### Hardware + macOS

M5 Max 128 GB, macOS 27.0

### Model

ddalcu/Qwen3.8-Flash-Next-MLX-Serve-mixed-4-8bit (qwen4_exp), MTP on (default), KV default, context default (no per-model settings were applied)

### What happened vs what you expected

The model stores its memory in one state buffer per GDN layer (48 heads × 128 × 128 values, bf16). This buffer is the KV-cache equivalent for the 36 linear layers. The disk cache tier saves copies of it to `~/.mlx-serve/kv-cache/<entry>/e<n>/s<pos>.safetensors`.

In one long session, the saved buffers stop being valid numbers after roughly 140k tokens:

| files | position in session | max finite value | values > 1e30 | NaN values |
|---|---|---|---|---|
| 7 files | 12.7k–25.8k tokens | 22–28 | 0 | 0 |
| 12 files | 144.7k–156.4k tokens | 2.565e38 | 2.31M of 28.3M (8.2%) | ~80,200 |

bf16 has a maximum of 3.39e38. The values climb to its ceiling and freeze. I measured these files exactly as the server wrote them (sha256 confirmed on my copies).

Four conclusions:

1. The live model computed these values. A file copy cannot create NaN. The writer path performs no arithmetic.
2. The damage is permanent in the model. The recurrence reuses its own state. Once one NaN exists, that state element never becomes a number again. All 12 later saves hold the same pattern.
3. The server then reads the damage back. A cache restore loads these files into the live model. Every answer after that point uses damaged memory.
4. Nothing visible marks the moment. No log line, no error, no metric. The server also holds no `inf` values, so the growth crossed the range and turned to NaN between saves.

**Reference deviation (fact):** mlx-lm keeps this state in float32 (`gated_delta.py`: `state = mx.zeros(..., dtype=mx.float32)`). mlx-serve declares the state output bf16 (`src/transformer.zig:22438-22447`). This dates from the first GDN port (f7c8c96). No commit since changes it, and no code checks the state for range.

[...]

PR #427, opened 2026-09-14
Port sampled MTP verifiers and paired Qwen3.8 routed gate/up

davidtai:
Qwen3.8 Flash-Next already has MTP and a large mlx-serve-specific kernel stack, but sampled creative requests could only use exact speculative acceptance. A matched llmprobe creative/thinking-off sanity run also left mlx-serve at 94.7 decode tok/s versus 101.3 for MTPLX PR #475 at the approximately 16K rung. That comparison used different packs, KV formats, and sampler filter orders, so it established a gap without assigning its cause.

Issue #419 started with a source-level inventory instead of a blind kernel transplant. Nine of the fourteen MTPLX mechanisms examined were already present in mlx-serve, and three more had the same purpose with a different or more general design. The missing, testable pieces were the two sampled verifier policies and complete request-local RNG handling. Of the kernel differences, only the paired routed gate/up producer survived an in-situ A/B.

This PR adds:

- opt-in Typical acceptance at <code>--mtp-typical 0.2</code>;
- opt-in TokenV3 acceptance at <code>--mtp-tokenv3 0.95</code> (<code>--mtp-cascade</code> is an alias);
- request-local keys for every sampled MTP draft, acceptance/correction draw, and full-accept bonus; and
- an opt-in physical-S=4 q4/group-64 paired routed gate/up kernel for the exact Qwen3.8 mixed pack.

Exact acceptance remains the default. The two relaxed verifiers are mutually exclusive and are explicitly not distribution-exact.

## Suggested review order

1. <code>src/mtp_acceptance.zig</code> contains verifier configuration and small-distribution arithmetic.
2. <code>src/main.zig</code> parses and validates the CLI modes; <code>src/scheduler.zig</code> carries the selected mode into request construction.
3. <code>src/generate.zig</code> installs one verifier graph per request and owns request-local sampled-MTP draw keys.
4. <code>src/transformer.zig</code> contains the paired gate/up Metal kernel, its exact pack contract, all-layer installation self-check, and fixed S=4 route.
5. <code>scripts/patch_llmprobe_creative_off.py</code> and its Python tests describe the disposable llmprobe audit patch used to force thinking off and stable timed request seeds. It does not alter serving behavior.
6. <code>docs/mtp-acceptance-port.md</code>, <code>NOTICE</code>, and the [benchmark record](https://github.com/davidtai/mlx-serve/tree/e40b64aa04716cfd5bf753e6f014251cfbdb4114/docs/benchmarks/qwen38-mtp-port) give the algorithm provenance and all plotted values.

## Verifier behavior

**Typical 0.2** computes the target entropy H(p) for each drafted token x and accepts when <code>p(x) &gt; min(1, 0.2 exp(-H(p)))</code>. It consumes no acceptance uniform. At the first rejection it samples from that target row; after a fully accepted draft it samples the bonus from the final target row.

**TokenV3 0.95** constructs <code>Top = {v: p(v) &gt;= max(p)(1-0.95)}</code>, <code>eta = sum(q(v), v outside Top)</code>, and <code>pi(v) = q(v) 1[v in Top] + p(v) eta</code>.

A proposal in Top is retained directly. Otherwise it uses the deferred <code>min(1, pi(x)/q(x))</code> coin and samples a rejection correction from <code>normalize(max(pi-q, 0))</code>. The implementation uses the full sampled draft distribution; substituting only p(x) or using p for the deferred coin would implement a different algorithm.

Both modes build their acceptance values and possible corrections in one lazy batched graph and perform one asynchronous evaluation per MTP round. Selection happens at construction. An enabled relaxed lane cannot silently fall back to exact acceptance, and greedy temperature continues through the existing argmax verifier.

The RNG change is part of this PR because sampled proposals and correction draws previously consumed process-global MLX state in several paths. A request seed now supplies a fresh key for each draw. The unchanged benchmark control predates this fix, so matching wire seeds between control and candidate makes requests repeatable but does not make their generated streams identical. The three-arm speed and quality results below measure the complete PR configuration.

## Paired gate/up kernel and install contract

<code>MLX_SERVE_MOE_VERIFY_PAIRED_GU=1</code> requests the new kernel at model load. It installs only when all of these invariants are true:

- Qwen3.8 Flash-Next with 48 MoE layers;
- hidden width 2560, expert width 640, 512 experts, and top-10 routing;
- separate contiguous affine q4/group-64 gate and up banks; and
- the physical verify width is S=4.

The kernel keeps mlx-serve's sorted row/expert indices and BF16 sigmoid table while pairing each gate output row with its up row so the hidden tile is loaded once. Before generation, every routed layer is compared bit-for-bit with the existing sorted gathers plus fused SwiGLU. A contract or self-check failure stops installation before a measured request; there is no enabled-path eligibility check or custom-to-stock fallback.

The same-binary kernel-only test used exact acceptance, KV8, sampled depth-3 drafts, temperature 1, top-p 0.95, top-k 20, thinking xhigh, three 16K prompts with up to 1,024 output tokens, and both boot orders:

| Arm | Mean decode | Total wall time |
| --- | ---: | ---: |
| Existing sorted gate/up gathers | 70.377 tok/s | 109.077 s |
| Paired gate/up kernel | 70.914 tok/s | 108.675 s |
| Difference | **+0.76%** | **-0.37%** |

All six request pairs matched completion length, final-answer digest, and reasoning digest. This isolates a small repeatable kernel win. It does not explain the larger whole-PR gains below.

## Final control and methodology

Several early runs were useful only for eliminating configuration mismatches. The final matrix uses no greedy drafts, no MTPLX AR samples, and no thinking-enabled requests:

- **Control:** unchanged upstream-main source tree <code>c5c57f9c0fa939c36f0c91d3eb7c1591e943964e</code>, exact acceptance.
- **Candidate A:** PR #427 tree <code>52d50c57a66839991c09808e300b45a572e1f658</code>, Typical 0.2, paired kernel installed on all 48 layers.
- **Candidate B:** the same PR tree and kernel, TokenV3 0.95.
- **Common setup:** <code>ddalcu/Qwen3.8-Flash-Next-MLX-Serve-mixed-4-8bit</code>, KV8, MTP forced to depth 3, sampled drafts, temperature 1, top-p 0.95, top-k 20, thinking off, prefix cache off, max fans, and zero cached input tokens.

The model config SHA-256 was <code>fe0b5952857299b31d75bedf1d8897faea14b4116a04c73a73152361e7788f59</code>. The audited llmprobe 0.6.7 bundle kept prompt bodies, request sequence, token caps, and wire seeds matched across arms. Each 1K-64K point is the median of three coding requests. Each 128K-1M point is one request, intended as a timing and memory scale check rather than a repeatability claim.

## Quality was screened before the long-context run

Startup logs first confirmed the requested verifier and all 48 paired-kernel installations, after algorithm tests and kernel self-checks passed. We then ran one llmprobe creative/thinking-off, concurrency-1 JavaScript code pass before spending the long-context GPU budget:

| Arm | HumanEval-JS | MBPP-JS | Total pass@1 | Token-cap stops | Errors |
| --- | ---: | ---: | ---: | ---: | ---: |
| Unchanged exact control | 45/50 | 34/50 | **79/100** | 4 | 0 |
| Typical 0.2 + kernel | 45/50 | 33/50 | **78/100** | 7 | 0 |
| TokenV3 0.95 + kernel | 45/50 | 32/50 | **77/100** | 5 | 0 |

All arms completed the same case IDs and reported the same 16,009 input tokens. The 1-2 task spread is a sanity result from one sampled pass@1 run. It is not a quality-equivalence claim and it is not the separate 164-task Python HumanEval suite. An earlier same-binary screen scored 76/100 exact, 78/100 Typical, and 77/100 TokenV3; the table above is the final screen against the unchanged control.

## Full 1K-1M sweep

Combined decode, gain, prefill, TTFT, allocator, and process-memory charts

| Requested context | Actual input | Control | Typical + kernel | Gain | TokenV3 + kernel | Gain |
| ---: | ---: | ---: | ---: | ---: | ---: | ---: |
| 1K | 1,035 | 98.5 | 114.5 | +16.2% | 116.4 | +18.2% |
| 2K | 2,048 | 101.5 | 111.3 | +9.7% | 111.1 | +9.5% |
| 4K | 4,088 | 101.1 | 107.8 | +6.6% | 109.3 | +8.1% |
| 8K | 8,258 | 88.7 | 103.4 | +16.6% | 104.0 | +17.2% |
| 16K | 16,295 | 94.3 | 102.4 | +8.6% | 104.2 | +10.5% |
| 32K | 32,885 | 90.9 | 108.1 | +18.9% | 111.9 | +23.1% |
| 64K | 65,446 | 100.6 | 109.6 | +8.9% | 109.2 | +8.5% |
| 128K | 136,358 | 98.0 | 104.6 | +6.7% | 106.0 | +8.2% |
| 256K | 273,113 | 86.6 | 96.5 | +11.4% | 98.9 | +14.2% |
| 512K | 546,701 | 73.4 | 83.4 | +13.6% | 86.9 | +18.4% |
| 1M | 1,043,375 | 54.1 | 65.2 | +20.5% | 68.7 | +27.0% |

The improvement is in decode. Prefill is effectively unchanged between arms:

Prefill throughput

At 16K, all nine timed coding requests generated 192 tokens. Median full-request wall time was 11.332 seconds for control, 11.190 for Typical, and 11.157 for TokenV3. Roughly 9.3 seconds was prefill, so the 8.6-10.5% decode gain reduced total wall time by only 1.3-1.5% for that short output.

At 1M, llmprobe tokenized the nominal rung to 1,043,375 input tokens and produced 192 output tokens in every arm. TTFT was 779.6, 777.5, and 777.8 seconds. The measured decode rates were 54.1, 65.2, and 68.7 tok/s, but prefill still dominates the roughly 780-second request.

<details>
<summary>Individual decode-gain and memory charts</summary>

Decode gain versus control

MLX allocator peak

Process physical-footprint peak

</details>

At 1M, the MLX allocator peak was 96.13 GiB and process <code>phys_footprint_peak</code> was approximately 101.6 GiB in each arm. These are different scopes and must not be added. The operator raised the macOS GPU wired limit to 116 GiB; the admission gate used an 8 GiB reserve against a 106 GiB empirical projection.

The model's native context limit is 262,144 tokens. The 128K point uses native RoPE; 256K, 512K, and 1M use an explicit YaRN factor-4 override. The high-context points are timing and memory statistics only. They do not establish long-context answer quality or GDN reference parity.

Issue #415 was investigated before the high-context statistics run. The apparent late-context NaNs came from seeking SafeTensors <code>data_offsets</code> from byte zero instead of from <code>8 + header_len</code>; corrected readers found finite state in the reported files, and a separate M5/HEAD reproduction stayed finite through 229K with cold restores. This PR does not change GDN state arithmetic. The qualification above remains because these 128K-1M requests were performance probes rather than a reference-state or long-context quality suite.

## What the results support

- The complete PR configuration improves sampled creative decode at every measured context rung.
- The paired gate/up kernel accounts for about +0.76% at its isolated 16K cell; the larger gains belong to the relaxed verifier/RNG/kernel configuration as a whole.
- Typical and TokenV3 leave prefill unchanged and trade distribution exactness for more accepted work per MTP cycle.
- The one-pass quality screen found no request failures and a 1-2 task pass@1 spread, which is enough for a sanity gate but not an equivalence claim.
- The 128K-1M results establish that the three configurations complete within the measured memory envelope. Their n=1 design does not establish variance or long-context quality.

## Validation completed

- pure toy-distribution tests for Typical and TokenV3;
- GPU graph tests for one-hot and sampled proposals;
- a full 248,320-vocabulary depth-3 TokenV3 finiteness test;
- seeded replay tests showing unrelated MLX draws cannot perturb a request;
- unchanged exact-verifier tests;
- paired-kernel small-shape GPU parity and all-layer load-time self-check;
- same-binary, two-order kernel serving A/B;
- full <code>zig build test</code> test-binary compilation and ReleaseFast server build;
- Python tests for the audited llmprobe patch;
- fresh three-arm quality screen;
- full 1K-1M timing/memory matrix under the exclusive GPU/service guard; and
- post-run restoration of the original production model, free GPU lock, candidate teardown, and automatic fan mode.

## Deliberately excluded

- The block verifier is not in this PR.
- The q3 sampled-head candidate is not in this PR.
- The paired down/reduce candidate is not in this PR.
- A dense-BF16 split-K QSA verifier kernel is not useful to the fastest control here: the measured KV8 arm was faster than BF16, and mlx-serve already has a quantized-KV QSA route.
- No MTPLX autoregressive optimization or sample is part of this comparison.
- No pack conversion is required; the existing mixed q8-dense/q4-expert mlx-serve checkpoint remains the serving artifact.

Refs #419 and #415.


ddalcu:

First review, at head `ec13a3ae`. The acceptance work holds up. `exactPrefix` (`mtp_acceptance.zig:68`) is a faithful extraction of the loop it replaces — same order, one `float(f32)` per position, same `>=` break — so the default route draws the same coins in the same sequence; the greedy `!stochastic` arm is untouched, and `accepted` is freshly zeroed above the call, so nothing is clobbered. `tokenV3Target` checks out arithmetically (the test's `pi` sums to exactly 1 and the deferred set is the `p < peak*(1-alpha)` tail), and the "keep-path consumes no coin" test is the right thing to pin for reproducibility. The `sampleFromProbsLazy` seed change is a fix rather than a regression and is wired correctly: both production callers pass `mtpSamplingDraw(...)`, which hands over the current `draw` and increments it, so each draft position gets a fresh key — and `seedKey` still returns an empty array when `seed == null`, leaving unseeded requests on the previous path.

- `src/transformer.zig:15380` — `MLX_SERVE_MOE_VERIFY_PAIRED_GU=1` makes `validatePairedGateUpPack` and `selfCheckPairedGateUpPack` `try` inside `Transformer.init`, so a model that does not match the one validated pack fails to LOAD instead of declining the kernel.
  Failure scenario: the env var is process-wide and `validatePairedGateUpPack` returns `error.PairedGateUpUnsupportedModel` on the very first conjunct for anything that is not `isQwen4()` with hidden 2560 / 640 experts-wide / 512 experts / top-k 10 (`:36940`), and `PairedGateUpMissingLayers` for any non-MoE model. Turn it on to A/B the kernel on a `--model-dir` server and every other checkpoint in the library — a Gemma, an LFM2, the same pack at a different quant — becomes unloadable, answering a named 503 from `/v1/load-model` for a kernel it was never going to use. The PR body reads "installs … only for the validated 48-layer Qwen3.8 pack", which is decline-and-continue; the code is fail-the-load. Every other shape-gated kernel in the tree declines and lets the stock path run ("other-shape kernels DECLINE"), and the same works here: fold the validator into the `break :blk false` arm so a non-matching model logs the decline and loads. Keeping the hard error for the *self-check* alone would still catch a matching pack whose kernel disagrees with stock, which is the case worth failing on.

ddalcu:

The increment `ec13a3ae..e40b64aa` is documentation only — a `docs/benchmarks/qwen38-mtp-port/` directory of measurement artifacts plus six lines in `docs/mtp-acceptance-port.md`. No file under `src/` changed, so the finding from the previous head stands unaltered; I re-read the block to confirm.

- `src/transformer.zig:15380` — `MLX_SERVE_MOE_VERIFY_PAIRED_GU=1` still makes `validatePairedGateUpPack` and `selfCheckPairedGateUpPack` `try` inside `Transformer.init`, so a model that does not match the one validated pack fails to LOAD instead of declining the kernel.
  Failure scenario: unchanged. The env var is process-wide and the validator returns `error.PairedGateUpUnsupportedModel` on its first conjunct for anything that is not `isQwen4()` with hidden 2560 / 640-wide experts / 512 experts / top-k 10, or `PairedGateUpMissingLayers` for a non-MoE model. Turn it on to A/B the kernel on a `--model-dir` server and every other checkpoint in the library answers a named 503 from `/v1/load-model` for a kernel it was never going to use. Folding the validator into the `break :blk false` arm keeps the hard error where it earns its keep — the self-check, which catches a matching pack whose kernel disagrees with stock — while a non-matching model logs the decline and loads.

One note, not a finding: `CLAUDE.md:100` puts the bench record at "`benchmarks.md` + `~/claude-tmp/bench-<tag>/`", and the 34 PNGs tracked on `main` today are website screenshots, app resources and one test fixture rather than measurement output — so ~600 KB of benchmark plots under `docs/` is a new kind of thing in the tree, permanent in history once merged. Worth a deliberate decision rather than a side effect of the rebase.

ddalcu:
🤖 **MLXServe-PRBot approved** — reviewed at head `c1222a1e`, no issues found.

The finding is closed with exactly the right split. `validatePairedGateUpPack` now declines — one `[mtp-verify] paired routed gate/up declined: <error>` line and `break :blk false` — so a pack or chip outside the contract loads on the stock route instead of answering a named 503, which is what every other shape-gated kernel in the tree does. `selfCheckPairedGateUpPack` keeps its `try`, which is the case that deserves a hard failure: a pack that DOES match the contract but whose kernel disagrees with the stock sorted gather is a wrong-answer risk, not a decline. The ordering is also safe — the self-check's `moe_layers.?` is only reached after validation passed, and validation is what rejects a null layer list. The prose in `docs/mtp-acceptance-port.md` was updated to describe the new behaviour rather than left claiming the old one.

Comparing the branch's own patch against its base before and after the rebase (`7e951144...e40b64aa` vs `ef7d2977...c1222a1e`) shows those two hunks as the only content change — 1998 lines before, 2005 after — so nothing else moved under cover of the rebase. Everything verified at `ec13a3ae` still holds: `exactPrefix` draws the same coins in the same order as the loop it replaces, the greedy arm is untouched, `tokenV3Target`'s π sums to 1, and both `sampleFromProbsLazy` callers pass `mtpSamplingDraw` so each draft position gets a fresh key.

PR #428, opened 2026-09-14
fix(registry): a status poll must not reset the idle clock

latent-variable:
## Summary

`--idle-evict-secs` never fires for anyone running a client that polls. `/props`, `/api/tags` and `/api/show` share `handleConnection`'s `ensureLoaded` / `defer release(lm)` pair with the generation routes, and `release()` stamps `last_used_ms`. Every poll therefore looks like use and resets the idle clock. The app's tray polls `/props` every 3s, so in practice the sweep sits armed and never picks a victim.

This is the other half of #398. That PR stopped `/props` cold-loading an **unloaded** model; the same endpoint against a **resident** one was never covered, and that is the direction that matters once eviction is on.

`releaseStatus()` returns the refcount without touching either recency stamp. The borrow protocol is unchanged, so the pointer is still protected for as long as the handler holds it. `POST /v1/load-model` keeps stamping — an explicit load is use.

## Changes

- `ModelRegistry.releaseStatus` — refcount-only release; it and `release` share `releaseRefLocked`.
- `Scheduler.releaseStatus` passthrough.
- `isStatusRoute(method, path)` — `GET /props`, `GET /api/tags`, `POST /api/show`; `handleConnection` picks the release accordingly.

## Test plan

Same server, 30s window, one variable:

| | result |
|---|---|
| idle, nothing polling | evicted at t+35s |
| `GET /props` every 5s | still resident at t+90s (3× the window) |

- `tests/test_idle_evict.sh` gains an arm that polls `/props` and `/api/show` against a **resident** model and requires it to evict anyway.
- Unit tests for `releaseStatus` and `isStatusRoute`.
- Controls run, not assumed: with the fix removed the unit test fails on its stamp assertion and the e2e arm fails with `/props polling pinned the model for 20s against a 2s window`. Nothing else changes state in either direction.
- `zig build test` 9/9 steps.
- On hardware, Qwen3.8-Flash-Next at 524k ctx, 120s window, a real status client polling `/props` once a second throughout: evicted at 145s idle, 70.13 GB released, RSS 20.1 → 0.2 GB, next request reloaded and answered normally.

https://claude.ai/code/session_01Cp9y54EtmaBkrZWSyxWtHa


latent-variable:
Rebased onto `7117d17`. The CHANGELOG conflict was your rewritten Fixes list — kept yours, added one line to it.

**Ran your `ci.yml` locally, every step, green:** ReleaseFast build, `zig build vz-agent`, `--version` / `--help` smoke, `tests/test_mlx_staged_nax.sh`, `zig build test --summary all` (9/9 steps, 2514/2685, 171 skipped, 0 failed), then `swift package resolve` → swatex patch → `swift build -c release` → `swift test`. Only the runner-provisioning steps were skipped (zig fetch, brew bundle, mlx build, llama fetch) — this checkout already has them.

**The app across an eviction.** `--idle-evict-secs` is off by default, so nothing exercises the state this changes. Worth calling out: before this fix the app could never see a model evict, because its own 3s tray poll is what prevented it. With the flag on, it now can. Drove every endpoint the app calls, on its 3s cadence, through an eviction:

```
setting ON: idle eviction ON — unloading models idle for 5s
phase 1  model resident, app polling every 3s     resident=1
phase 2  evicted at t+6s WHILE the app polled
phase 3  app keeps polling, nothing resident      polling did not resurrect it
phase 4  user sends a message                     reloaded and answered
phase 5  explicit unload button                   OK
```

Any 5xx or non-JSON body from `/health`, `/props`, `/v1/models`, `/metrics.json` or `/v1/providers` would have failed it; none occurred in any phase.

**e2e on this exact head:** `tests/test_idle_evict.sh` PASS, new polling arm included. Control still holds — with the fix removed that arm fails with `/props polling pinned the model for 20s against a 2s window`.

ddalcu:

The mechanism is right. `releaseStatus` shares `releaseRefLocked` with `release`, so the refcount protocol and the evictor broadcast are untouched and only the two recency stamps are skipped — and skipping `lru_clock`/`last_used_ns` as well is correct, since a poll should not move a model up the resident-cap LRU either. `path` has its query string stripped at `server.zig:2005`, so `/props?x=1` still matches. The registry test is built the right way round: it backdates to an absolute `stale` rather than reading the stamp under test, and then contrasts with a real `release` that does restamp, so neither half can pass vacuously.

- `src/server.zig:6054` — `isStatusRoute` lists `/api/tags` but not `/api/ps`, its immediate sibling in the same dispatch block, served by the same handler, and the Ollama endpoint whose entire purpose is reporting which models are resident.
  Failure scenario: `/api/ps` at `:2531` is `handleOllamaTags(allocator, stream, true)` — five lines below the `/api/tags` arm at `:2526`, both after the `defer` at `:2353` — so it reaches `scheduler.release(lm)` and stamps `last_used_ms`. An Ollama-compatible client polling `/api/ps` to render "running models" is the same shape as the tray polling `/props`, and it pins every resident model against `--idle-evict-secs` exactly as before this PR; the new `tests/test_idle_evict.sh` arm polls `/props` and `/api/show`, so it passes while the gap stands. `GET /api/version` at `:2519` is in the same block and even more plainly not use. Adding `/api/ps` (and `/api/version`) to the `GET` arm is a one-line change, and the unit test already has the right shape to pin them.

latent-variable:
Thanks — the `/api/ps` finding doesn't hold, but chasing it found a real problem with my patch, so it was worth raising.

**Why it doesn't hold.** `/api/ps:2531` and `/api/tags:2526` are inside `handleOllamaEarly`, which `handleConnection` calls at **`:2174` and returns from on a hit** — above `ensureLoaded` at `:2302` and the defer at `:2355`. File order isn't execution order; those routes never reach the release.

Measured rather than argued. 5s window, resident model, each endpoint polled every 0.7s for 17s:

```
GET  /api/ps       -> evicted anyway (does NOT pin)
GET  /api/tags     -> evicted anyway (does NOT pin)
GET  /api/version  -> evicted anyway (does NOT pin)
POST /api/show     -> evicted anyway (does NOT pin)
control: polling /v1/chat/completions -> PINNED
```

The control matters: without it, four "does not pin" results are indistinguishable from a probe that can't detect pinning at all.

**What it did find.** By the same reasoning, the `/api/tags` and `/api/show` entries in my own `isStatusRoute` were unreachable — dead code that reads as coverage. Cut in `f29b72f`; the predicate is now `/props` alone, which is genuinely the only status route falling through to the shared pair. The test now asserts `/api/ps`, `/api/tags`, `/api/version` and `/api/show` are **not** status routes, with the reason, so nobody re-adds them as a gap. Dropped `/api/show` from the e2e arm too — it was decoration.

Re-verified on `f29b72f`: `zig build test` 9/9 (2514/2685, 171 skipped, 0 failed), `tests/test_idle_evict.sh` PASS, and the arm still fails with the fix removed (`/props polling pinned the model for 20s against a 2s window`) — re-run after the edit, since a changed test doesn't inherit its old control.

ddalcu:
🤖 **MLXServe-PRBot approved** — reviewed at head `f29b72f8`, no issues found.

My `/api/ps` finding on `d54459eb` was wrong and this head is right to narrow the predicate rather than widen it. `handleOllamaEarly` is *called* at `server.zig:2174`, unconditionally for any `/api/` path and well above `ensureLoaded`, and it answers `/api/version`, `/api/tags`, `/api/ps`, `/api/show` and `/api/pull` with `return true`. I had read the comparisons at `:2516`+ as inline dispatch below the release `defer`; they are that function's body, which merely sits later in the file. None of those routes ever reaches `ensureLoaded` or the release pair, so listing them in `isStatusRoute` was dead code that read as coverage — exactly as the new comment says.

`GET /props` really is the only survivor: `/health`, `/metrics`, `/metrics.json` and `/v1/models` return above the resolution block, `/api/chat`, `/api/generate` and `/api/embed(dings)` deliberately fall through as use, and `/v1/load-model` is a load. The test now pins the Ollama paths as negatives with the reason, which is the honest shape, and dropping the `/api/show` poll from the `test_idle_evict.sh` arm keeps that arm describing what it actually exercises. The mechanism I checked at the previous head is unchanged and still correct: `releaseStatus` shares `releaseRefLocked`, leaves both recency stamps alone, and the registry test backdates to an absolute value with a restamping `release` as its contrast.

Issue #429, opened 2026-09-14
Attached videos are dropped on send: the chip says "16 frames", the model gets none


lojza3d:
### Before opening

- [x] I searched open and closed issues, open PRs and recent commits for this, and it is not already reported or fixed

### Priority

medium (workaround exists)

### Area

Server / API (chat, completions, messages, responses, embeddings)

### mlx-serve version

Version 26.9.3-dev (26.9.3-dev), first seen on 26.8.11

### Hardware + macOS

M5 Max 128 GB, macOS 26.6.2

### What happened vs what you expected

Attaching an .mp4 to a chat message does nothing. The composer accepts it, the chip reads `Video · 16 frames`, and on send the frames are silently discarded. The model answers about a video it never received, so it invents one.

Expected: the sampled frames reach the model, the way images and audio clips already do.

The frames never reach a `ChatMessage` either, so they are not in `chat-history.json` and there is nothing in the transcript to show. Nothing reports a failure anywhere.

## Where it goes

`ChatView.swift:3741` (`proceedSend`) takes the pending videos, and the guard uses them to decide there is something to send:

```swift
let attachedVideos = consumePendingVideos()
let attachedAudio = consumePendingAudio()
let pdfText = consumePendingPDFsAsText()
guard !text.isEmpty || attachedImages != nil || attachedVideos != nil || ...
```

Then the send passes everything except them:

```swift
chatEngine.runTurn(sessionId: sessionId, userText: text,
                   images: attachedImages, audio: attachedAudio,
                   config: buildTurnConfig(),
                   approval: { await requestToolApproval($0) })
```

`ChatTurnEngine.regenerate` (`ChatTurnEngine.swift:512`) has the same shape, so regenerating a turn would drop the video too. Edit & Resend (`ChatView.swift:3862`) is the one path that does pass `videos:`, which shows the intended shape and that the other two are an omission. It is not a workaround: it reads them back from the stored message, where the first send never put them, so it passes `nil` explicitly. No path in the app gets a video to the model today.

Everything downstream is in place: `ChatTurnEngine` puts `msg.videos` into the request for the last user message, and `MultimodalContent.build` emits the `video_url` block. Both are simply handed nothing.

## Why it compiles

The protocol requirement declares no default and says so:

```swift
// ChatTurnEngine.swift:20-28
func runTurn(sessionId: UUID,
             userText: String,
             images: [ChatImage]?,
             videos: [ChatVideo]?,
[...]

PR #430, opened 2026-09-14
Audio attachments move to disk, as a 16 kHz WAV

lojza3d:
### What changed

The second half of #288. Pictures moved to `~/.mlx-serve/attachments/` in #313; audio clips still rode `chat-history.json` as base64 float32 (64 kB per second of speech, re-written on every save, loaded whole at launch).

- `AudioClipFile` (new): a clip on disk is a canonical 16 kHz mono 16-bit WAV, written through the app's existing pure WAV writer (`AudioReference.wavData`) and read back by a small parser that accepts only that shape. Anything else answers nil rather than samples that would be a guess handed to the model.
- `ChatAudio` persists `id`, `name`, `path` (the `ChatImage` shape). `pcm` is decoded from the file; a history from before this change or a file since deleted decodes with empty `pcm`, never a throw (one throw is `loadChatHistory`'s `?? []`, every conversation).
- The 16-bit round trip is taken once, on send (`AudioClipFile.stored`, a pure helper with its own test), and the in-memory `pcm` is re-read from that encoding, so the first turn and a regenerate after a relaunch hand the model the same bytes.
- A sample-less clip is filtered off the wire (`MultimodalContent`) like a byte-less image, and its chip in the transcript says the file is gone.
- Deletion: `AttachmentStore.removablePaths` reads `audio` beside `images`, same sparing rule for forks, same two removal sites.
- A mic recording is named `Recording`; the chips already print the duration, and the old `Recording · 6s` name read as `Recording · 6s · 6.3s` and landed in the filename.

Telegram voice notes keep `path` nil: their session is never persisted.

### Found on the way, not fixed here

With tools on, an audio-only turn never reaches the model: `AgentEngine.buildAgentHistory` builds multimodal content from images only and drops a user message whose text is empty (the server log shows no `Multimodal:` line and the history ends on the assistant's own reply). Same place #429 will need for video; I will send it with that.

### Tests

`AudioAttachmentTests` (12): WAV round trip within 16-bit precision, canonical header, foreign files decode to nothing, a `Data` slice decodes like the whole file, the send-time store carries the file's samples and its path, history carries the path and not the samples, decode reads the file back, legacy and missing-file records both decode, sample-less clips are not sent, message and session deletion remove the clip and spare a fork's copy, filename scheme. `bash app/test.sh`: 3382 tests, 0 failures.

Live on Gemma 4 12B unified (the one audio model): send (file lands, server logs `Multimodal: 1 audio clip(s)`, model answers the clip), relaunch (chip keeps its duration), regenerate after relaunch (model still hears it), fork, delete message and delete chat remove the WAV, a WAV removed by hand shows as gone and is not sent.

Docs: rule in `app/CLAUDE.md`, story in `docs/gotchas/app.md` (incl. how to test audio live: the model reads a clip as typed text and needs about four seconds of speech).


ddalcu:
🤖 **MLXServe-PRBot approved** — reviewed at head `6f9d1d3e`, no issues found.

I walked the hand-rolled WAV parser for the usual hazards and it is bounded everywhere: the chunk loop's `guard body + size <= wav.count` covers both the `fmt` reads (`size >= 16` puts `body + 14 + 2` inside it) and the `data` subdata; `offset = body + size + (size & 1)` advances by at least 8 on a zero-size chunk so it cannot spin; `Int(u32(...))` cannot overflow on a 64-bit `Int` and a 0xFFFFFFFF size simply fails the bound; the per-sample loop reads `i * 2` for `i < count / 2`, which stays in range for an odd byte count too; and `tag` going through `String(decoding:as: UTF8.self)` cannot throw on arbitrary bytes, it just fails the comparison. The slice normalisation at the top is what makes the absolute offsets safe, and it has its own test.

The ownership rules line up with the picture precedent they cite. `paths(in messages:)` is now the single collector, so `removablePaths(orphanedBy:)` and `removablePaths(deleting:)` see audio on BOTH the doomed and the surviving side — which is what keeps the fork-sparing rule true for a clip two sessions share — and both entry points are pinned by tests. No third removal site is added, so the `AttachmentStoreTests` count stands. `consumePendingAudio` writes on SEND like the images, and regenerate/edit-and-resend never re-enter it (they rebuild from the stored `ChatAudio`, and `pendingAudio` is empty), so `truncateMessages` still touches no file; even if it did, `filename(id:)` is stable and the float→int16→float round trip is idempotent after the first pass. `ChatAudio.init(from:)` mirrors `ChatImage`'s exactly — same eager `contents(atPath:)`, same `?? Data()` instead of a throw — so one unreadable clip cannot empty every conversation, and a sample-less clip is filtered off the wire and says so in the transcript rather than claiming it was sent.

PR #431, opened 2026-09-15
fix(prefix-cache): cold-boot SSD restore recovers the highest checkpoint

beamivalice:
## Defect

On a hybrid arch (Qwen3.8-Flash-Next), a **text-only** conversation restored almost nothing from the SSD tier. A 73k-token chat, restarted and replayed byte-for-byte, came back with 16,384 of 73,398 tokens and spent **34.2 s** re-prefilling what it already had on disk — while the RAM tier, given the identical request, matched 73,293 in 2.0 s. The same cap hit every mid-session disk restore in that session, at prompt lengths from 61k to 229k, always landing on the entry's lowest checkpoint.

## Cause

`scheduler.firstMediaPlaceholder` scanned every prompt for `image_token_id` / `audio_token_id` / `video_token_id` unconditionally. **Those are ordinary vocabulary entries** — this conversation pasted repo sources and held `248056` (`image_token_id`) at index 18,338 — so a request carrying no media at all got `media_start = 18338`.

That value is the disk hybrid lookup's `limit` (`disk_limit = media_start orelse prompt_ids.len`). The trace shows the consequence exactly: the donor that covered the whole prefix (`shared=73390`, checkpoints 49,152…73,312) has **no checkpoint below 18,338**, so `highestSsmPosAtOrBelow` returned null and it was skipped outright; a stale entry whose lowest checkpoint is 16,384 won instead.

```
[disk-trace] e1304 shared=73390 kv_len=73390 limit=18338 cps=5 lo=49152 hi=73312 usable=18338 cp=null
[disk-trace] e1303 shared=49084 kv_len=49086 limit=18338 cps=6 lo=16384 hi=48994 usable=18338 cp=16384
[disk-trace] chose e1303 usable=18338 cp=16384 (of 74 entries)
[disk-cache] restored 16384/73398 tokens from SSD in 856ms (ssm@16384)
```

The tier itself was innocent: replayed offline against those very files (real `meta.json` and `tokens.bin`, sparse placeholders at the recorded sizes) `bestHybridMatch` picks the right entry. The RAM path was unaffected because it does not take the boundary as a hard cap.

## Fix

`firstMediaPlaceholder` takes `has_media` (`params.vision_embeddings != null`, the one signal that the request has media rows to splice — image, audio and video all ride it) and answers `null` without it. A media boundary exists only where media rows do.

The same value keys `commitWithMediaState`, so a text entry no longer carries a bogus boundary into checkpoint inheritance and retention thinning either.

## Guard

`firstMediaPlaceholder: a placeholder id in ORDINARY TEXT is not a media boundary` — red first (`expected null, found 2`), green after. The existing "finds every dynamic media kind" test keeps the positive cases and now passes the gate explicitly.

Also added `DiskTier: a fresh tier over the same root ranks by the HIGHEST restorable checkpoint`: it commits an entry with 8 checkpoints, opens a **fresh** tier over the same directory (the restart) and asserts the lookup takes the highest checkpoint at or below the match. It passed before this change too — that is the point; it pins the contract the symptom appeared to violate, so the next investigation starts one level up.

## Verified

- `zig build test` — exit 0, 9/9 build steps, no failures.
- `bash tests/test_prefix_cache_disk.sh <qwen3.8-flash-next> 8096` — **ALL PASS** (section 6, the hybrid Qwen3.5-0.8B arm, SKIPped: that model is not on this box).
- `zig build -Doptimize=ReleaseFast` clean.
- **Live, same conversation and flags, before vs after:**

  | | restore after restart | wall |
  |---|---|---|
  | before | `16384/73398` | 34.2 s |
  | after | `73293/73375` | **1.6 s** |

  Mid-session restores went deep too (`12585/12616`, `24907/24938`, `73293/73375`) where before every one capped at 16,384.

## Not verified

- The hybrid arm of `test_prefix_cache_disk.sh` (no Qwen3.5-0.8B pack locally).
- Only `qwen4_exp` was exercised live; the change is arch-independent (it is the media-boundary predicate, not an arch path).
- The investigation used a temporary `MLX_SERVE_DISK_LOOKUP_TRACE` trace in `bestHybridMatch` to capture the numbers above; it is **not** part of this diff.


ddalcu:
🤖 **MLXServe-PRBot approved** — reviewed at head `c0b19485`, no issues found.

The gate is the right predicate and covers more than its name suggests: `vision_embeddings` is the single splice channel for vision AND audio (`scheduler.zig:11`, and the Gemma-4 unified path routes both through it), so there is no audio-only request that keeps placeholders while the field is null. It is also the only `firstMediaPlaceholder` call site — the other `params.vision_embeddings != null` at `:697` is the pre-existing drafter gate.

The consistency that makes it safe holds on the producing side too: a vision encode that fails leaves `local_ve` null AND skips `insertMultimodalTokens`, so a null embedding never coexists with inserted placeholder rows. And the cache-side invariant is unaffected — an entry committed with media keeps its own `media_start`, which is what bounds a cross-key match, so losing the boundary on a text-only request cannot let a match run past somebody else's pixels.

The `kv_disk_cache.zig` addition is a characterization test with no production change beside it, which is the honest way to pin the second half of the symptom: it asserts a tier reopened over the same root still carries all eight `ssm_positions` and that `bestHybridMatch` ranks by the highest restorable checkpoint cold as well as warm.

PR #432, opened 2026-09-15
perf(sampling): exact shortlist top_p/top_k replaces the full-vocab rank

beamivalice:
Rebased onto `bdcf5a1d` and ported to the rank contract that commit established.

## Defect

Every sampled token ranked the whole vocabulary. `applyTopP` argsorted the row and `applyTopK` paid a second full pass — because on Metal `Partition::eval_gpu` and `ArgPartition::eval_gpu` both say *"we direct partition to sort for now"*, so `mlx_topk` and `mlx_argpartition` **are** that same multi-block merge sort. A shortlist built from `argpartition` buys nothing here, which is worth writing down before someone tries it. `bdcf5a1d` did not change this: it still argpartitions for top-k and argsorts for top-p, plus a gather and a scatter.

## Algorithm

`topRanksDescending` takes the exact top `m` values **and their column ids** with no full-row rank. Every element of a chunk is at most that chunk's max, so a chunk whose max is not among the `m` largest chunk maxima holds no top-`m` element: at most `m` chunks can, so one small sort of `V / chunk` chunk maxima plus one of `m * chunk` candidates is exact, balanced at `chunk = sqrt(V / m)`.

- **`applyTopK`** scatters the shortlist's ids through `keepByIndex`. Exactly `k` columns kept.
- **`applyTopP`** reads its cutoff off the k-shortlist, because nothing below the row's top `k` can be the cutoff once top-k has run. Pure `top_p` has no bound on the nucleus and still ranks the row; `sampler_shortlist_draws` / `sampler_fullsort_draws` count which route a request took, decided while the graph is built so no host read is needed.

Everything stays lazy and on device: no `mlx_eval`, no host read inside the sampler.

## Exactness under the rank contract

Cutting by rank makes exactness a question about **order**, and three things have to line up:

- **One tie rule, shared by both routes.** `ranksDescending` is a stable ascending argsort of the *negated* row: value descending, ties by **lowest column id** — the `msv_qsa_select` rule. mlx's merge sort is stable (`ThreadSort` swaps only on a strict less-than; the merge takes from B only when `b < a`, and A holds the lower ids). That is an undocumented property of a pinned dependency, so a test asserts the order directly and will fail loudly on an mlx bump.
- **The chunk pick takes the LOWEST ids among equal maxima.** A tie group's contributing chunks are a *prefix* of its ascending-id list, since a lower chunk id holds a lower column id, which outranks. Taking the last `m` of an ascending argsort gets this exactly backwards; a 248k all-equal row then keeps the wrong columns.
- **Candidates are laid out in ascending column order** (the winning chunk ids are sorted before the gather), so the stable rank over `m * chunk` candidates breaks ties the way the whole row would.

The nucleus is the mass **strictly above** each rank — an exclusive scan down the ranking — which is a function of the shortlist alone, because everything outranking a top-`k` column is itself top-`k`. Rank 0 sees zero mass, so the argmax is always kept and `top_p → 0` stays greedy. The normalizer is the **whole row's** `logsumexp` (mlx's logsumexp accumulates in f32 whatever the input dtype).

The scan runs in **f32**. mlx instantiates cumsum at the input dtype (`sum_bfloat16_bfloat16`) and logits off a quantized `lm_head` are bf16, so the old accumulator dropped **148 of 3720** nucleus columns on one of six probe rows, and made the running sum depend on how many entries the scan walked — the one thing a shortlist changes. This is a small behaviour change for bf16 models at `top_p < 1`: the nucleus is now the exact one.

## Tests

- `sampler ranks break ties by the lowest original index` — the shared ordering contract, asserted directly.
- `topRanksDescending returns the exact top m values and column ids` — V ∈ {8192, 248320}, m ∈ {1, 20, 257}, 3 rows incl. a 0.25-step tie row, against the whole row's ranking.
- `applyTopK keeps exactly k columns, the highest ranked ones` — k ∈ {1, 20, 64}, incl. a 1.0-step tie row where a value cutoff would keep thousands.
- `shortlist top-p equals the whole row's ranking, ties and 248320-wide rows` — **byte-identical** filtered logits over six `(top_p, top_k)` settings, a 0.25-step tie row and a row where **every column is identical**, plus the same token drawn under 8 keys.
- `shortlist top-p holds on rank-3 [B, L, V] blocks` — the MTP stochastic-verify shape, per-row cutoff.
- `applyTopP keeps the exact nucleus on a bf16 vocab row` — six 248,320-wide rows against an f64 nucleus. The reference is allowed to differ by one column: the boundary column's own probability is ~1e-5 while the f32 scan carries ~1e-7 of error, so the exclusive mass can cross `top_p` one column early. The defect it pins was 148 columns wide.
- `sampler route census` — greedy returns before either filter (counters stay 0); routes taken as expected.

Mutation-checked: dropping the f32 promotion fails the nucleus test; picking the highest chunk ids among tied maxima fails both ranking tests (`mismatch at 59: 244922 vs 20730`).

`zig build test -Doptimize=ReleaseFast`: **2531 passed, 171 skipped, 0 failed.**

## Live A/B — pending

Against the **pre-rank** baseline (`10756309`), M4 Max / Qwen3.8-Flash-Next mixed-4-8bit / `--no-mtp` / 256 tokens / 3 runs, medians of the server's `timings`: the `top_p 0.95 + top_k 20` filter cost **1.13 → 0.51 ms/token** (60.5 → 64.0 tok/s, against a greedy control of 65.0 → 66.1 that moved 1.8% between boots, so read the per-token filter cost, not the absolute delta). Pure `top_p` was unchanged by design.

Those numbers predate `bdcf5a1d`. A fresh A/B against a `bdcf5a1d` baseline binary is queued for the next free GPU window and will be posted here; the box is busy with two other measurements.


ddalcu:
🤖 **MLXServe-PRBot approved** — reviewed at head `e0063bc4`, no issues found.

The selection argument holds, including at the boundary: if a chunk holding a top-`m` element `x` were not among the `m` largest chunk maxima, `m` other chunks would each contribute an element ≥ `x` at a distinct position, putting `x` at rank ≥ m+1. Ties do not break it because `topValuesAscending` returns VALUES, not indices, so a tie-broken argsort still yields the same multiset. The arithmetic is safe too — `groups - want` cannot underflow because the `want * chunk >= v` early return guarantees `groups > want`, and `shortlistChunkWidth` clamps to ≥ 16 then `@min(c, v)`, so `chunk >= 1` and the ceiling division has no zero divisor.

Per-row correctness survives the rewrite, which is the class `CLAUDE.md` names for these helpers: every op is `_axis`, the chunk reshape carries the leading dims through, and the cutoff stays shape `[..., 1]` broadcasting against the row — so a `[m, V]` draft block cannot have one row's cutoff mask the others. `applyTopK`'s output is unchanged from main: the old `topk_axis` → `min_axis` produced the same k-th-largest value and the same `>= cutoff` mask, ties included. `applyTopP`'s criterion is an algebraic restatement — old `inclusive prefix > 1 - p` over an ascending row is exactly new `exclusive suffix < p` — and the shortlist normalizer reasoning is right, since anything strictly above a candidate is itself larger and therefore also a candidate, so the exclusive suffix over candidates equals the true mass above.

The precondition the bound rests on is met at all six production call sites: `applyTopK` runs first on the same array and `sampling.top_k` is what is handed to `applyTopP`, so nothing below the k-th largest can be the cutoff. The bounded-vs-unbounded equivalence test is the right guard — bit-for-bit across five `(p, k)` configs including `k = 1` and `k > v`, plus the drawn token for 16 seeds. The f32 filter cast changes sampled output against main for bf16 logits, but it is commented, covered by its own bf16 test, and cannot move a greedy draw, since the argmax is always inside the nucleus.

ddalcu:

The move from a value cutoff to a rank cut is the right fix and the exactness argument survives it. I re-derived the chunk-max lemma with the id tie-break: if a top-`want` column sat in a chunk outside the `want` highest-ranked chunk maxima, each outranking chunk would contribute a column that beats it — strictly on value, or on value-equal-and-lower-id, since a lower chunk id spans strictly lower columns — so it could not have been top-`want`. Ranking the chunk maxima with `ranksDescending` is what supplies that (max desc, chunk id asc), and sorting the picked chunk ids ascending before the gather lays candidates out in ascending column order so the stable rank over them breaks ties the way the whole row would. `applyTopK` keeps exactly k: padding can never be selected (at most `chunk - 1` padded entries against `want * chunk` candidates, and they sort last among equal `-inf` because they hold the highest columns), so the `v - 1` clamp cannot collide two ids into one `put_along_axis` slot. The scan direction is right for the new order too — descending ranking with `reverse=false, inclusive=false` is still mass strictly above — and leaning on mlx's stable merge sort is acknowledged as an undocumented property with a test asserting the order directly.

- `src/generate.zig:12265` — `above < top_p` excludes rank 0 when `top_p` is exactly 0, so the row comes back all `-inf` and the comment two lines up ("Rank 0 sees 0, so the argmax is always kept") does not hold at the boundary.
  Failure scenario: `"top_p": 0` is inside the accepted range — `parseJsonFloatOpt(root, "top_p", 0.0, 1.0)` at `server.zig:7911` and its three siblings admit it, and the caller runs the filter for any `top_p < 1.0` — so the request reaches `applyTopP` with `threshold = 0`. Rank 0's exclusive mass is 0, `0 < 0` is false, every rank is excluded, and the mask is empty: `softmax` over an all-`-inf` row is NaN and `mlx_random_categorical` then draws off NaN logits. `top_k: 1` does not save it, since top-p runs after and masks the one survivor. The old code was fp-dependent here (`cumsum <= 1.0` at the last element sometimes left the top token alive); this makes it deterministic, and it is the same boundary the CHANGELOG entry is about. Keeping rank 0 unconditionally — or treating `top_p <= 0` as greedy at the caller — restores the invariant the comment states.

PR #433, opened 2026-09-15
perf(mtp): sampled requests draft from the exact re-scored shortlist

beamivalice:
Based on origin/main `ef7d2977`. Not stacked on #432 — that PR is merged (`ff3d7574`) and this branch sits on top of it plus `ef7d2977`, so the 32-row draft filter goes through the merged rank-based `applyTopK`/`applyTopP` and `topRanksDescending` as they exist on main. No second selection primitive: after `top_k` the row's finite entries are the shortlist, which is exactly the normalizer `ef7d2977` relies on, and a 32-entry row takes `topRanksDescending`'s full-rank branch.

## Defect

At temperature the MTP draft is the argmax and acceptance is `min(1, p_target[argmax])` against a one-hot q — i.e. `p(argmax)`, which collapses on a flat row. At `MLX_SERVE_MTP_FORCE_DEPTH=3`, per-index acceptance reads **0.50/0.22/0.09** on thinking prose against **0.97/0.81/0.63** on code: the third draft index essentially never accepts, so the planner narrows the width and, with two concurrent chats, stops speculating altogether — **19%** of tokens from MTP rounds against greedy's 96%.

The sampled proposal that already existed (`MLX_SERVE_MTP_DRAFT_GREEDY=0`) lifted per-draft acceptance 7–12 points and netted zero.

## Mechanism

`q_probs != null` set `need_logits`, so every draft step dropped the coarse 3-bit shortlist and paid the full 8-bit 675 MB lm_head readout: `corr` 0.53 → 2.6 ms, `predraft` 1.75 → 2.9 ms, +3.2 ms of a +4.2 ms round, with the trunk `eval` term unmoved. In a group round a second cost dominated: the sampled arm looped rows calling `probsAtLastPos` + `sampleFromProbsLazy` over `[1, V]` per row per step while the rerank arm stacked rows into one `[N,1,H]` — the group audit measured `chain` 3.82 → 9.07 ms/round at N=4, 85% of that arm's whole penalty. These kernels are launch-bound, not bytes-bound: a full-vocab sort of 1 row and of 6 rows differ by 40%.

## Fix

- `mtp.rerankShortlist` stops before the argmax and returns the 32 candidate ids plus their exact trunk-head logits. `rerankSelect` is that plus `shortlistArgmax`, op for op what it was.
- `shortlistProposal` filters those 32 rows through the **shared** sampler (`filteredProbsRows`, so a draft can never keep a token the target's filter would have cut), draws the draft from that row, and scatters it into a dense `[1, V]` q. Zero off the shortlist costs nothing: Leviathan is exact for any q and the residual `max(p−q,0)` is what reaches the rest of the vocabulary. Filtering *inside* the shortlist can only move the acceptance rate, never the output distribution.
- **The group keeps its batched shape.** A sampled row now joins the rerank group (`mtpRowRerankMode` keys on `!= .full_logits`, so a mixed group still does one coarse readout and one head forward per step) and `shortlistProposalRows` runs ONE filtered block plus ONE categorical over the sampled rows' `[K, 32]`. Gated on `sampleRowsHomogeneous` and no row carrying a seed — one categorical cannot carry a per-row key — falling back to the per-row proposal otherwise. The correction and bonus draws are untouched (they still use MLX's global RNG; that is a separate finding).
- **The proposal is per REQUEST.** `mtpDraftGreedyFor` keeps the argmax at `temperature <= MTP_SAMPLED_PROPOSAL_MIN_TEMP` or `top_k == 1`, so a greedy request is untouched and `MLX_SERVE_MTP_DRAFT_GREEDY` remains the A/B override (1 greedy, 0 sampled, absent = the rule).
- **`MTP_SAMPLED_PROPOSAL_MIN_TEMP = 0.5` is measured, not derived**: at forced depth 3 greedy took temperature 0.3 (per-draft 69.9%/48.7% code/prose vs the sampled 60.8%/43.8%) and lost temperature 0.7 (55.6%/34.8% vs 62.2%/42.9%). 0.5 is the conservative point in that bracket.
- **Draft temperature is per family** (`mtpDraftTempFor`): 0.6 was swept on the dense 27B oQ4e sidecar head and stays there; on Flash Next the target's own temperature drafts better (six cells, mean per-draft 53.4% vs 51.1%), which agrees with MTPLX's own 3.8 sweep. `MLX_SERVE_MTP_DRAFT_TEMP` is the sweep lever for the next head.

## Tests

- Toy-vocab exactness for a **shortlist** q (zero off the shortlist): the marginal is still the target's p to 0.01 over 400k draws, every draft has `q > 0`, and the two tokens the shortlist never sees are reached — only the residual can produce them.
- GPU tests of the proposal builder: q is nonzero only on shortlist ids, sums to 1, peaks on the highest exact logit, and every draw is a candidate — over one row, and over **three rows with disjoint candidate sets** for the batched block (the `applyTopK`-on-`[m, V]` axis-bug class).
- `mtpDraftStepPath` (a sampled step keeps the rerank; only the chunk-A confidence gate needs the whole head) and `mtpDraftGreedyFor` (greedy at temperature 0, 0.005, 0.3 and `top_k == 1`; every override).
- `zig build test -Doptimize=ReleaseFast`: 9/9 steps, 2537/2708 passed, 171 skipped.
- `MTP_FORCE_ENABLE=1 MTP_TEST_MODEL=<Flash Next> ./tests/test_mtp_equivalence.sh`: **11 passed, 0 failed** (the temp-0 byte bar — greedy targets never reach this change).

## A/B

Apple M5 Max 128 GB, macOS 26.6.2, `Qwen3.8-Flash-Next-MLX-Serve-mixed-4-8bit`, `MLX_SERVE_ROUND_COST_PERSIST=0`, one boot per arm, 100 s idle before every timed arm, medians of 3, `--max-concurrent 4`. Baseline rows re-read from the diagnosis report's own artefacts with the same parser.

**Forced depth 3, solo** (`MLX_SERVE_MTP_FORCE_DEPTH=3 MLX_SERVE_MTP_TRACE=1`, temperature 1.0 / top_p 0.95). Greedy control and shipped default are separate boots of the same binary:

| cell | greedy per-draft | greedy tok/s | shipped per-draft | shipped tok/s | round greedy → shipped |
|---|---|---|---|---|---|
| code / xhigh | 49.8% | 84.4 | **68.3%** | **104.7** | 29.8 → 29.2 ms |
| code / medium | 62.2% | 98.3 | 65.1% | 103.1 | 29.4 → 29.1 ms |
| prose / xhigh | 34.8% | 70.1 | 38.6% | 75.4 | 29.4 → 28.8 ms |
| prose / medium | 37.0% | 71.6 | **48.7%** | **85.8** | 29.8 → 29.0 ms |

The cost the old sampled arm paid is gone: `corr` 0.47–0.49 ms (was 2.57–2.64), `predraft` 1.81–1.93 (was 2.84–2.92), and `[qwen4] MTP draft rerank` + `[mtp] draft rerank engaged` are in every sampled log.

**Draft temperature 0.6 vs 1.0**, same cells, per-draft: 60.8/54.3/47.7/38.6/62.2/42.9 against 63.6/69.9/42.0/44.7/55.6/44.7 — 1.0 wins four of six, mean 53.4% vs 51.1%.

**Two and four concurrent chats, auto depth** (256-token creative cells, in-boot greedy controls):

| cell | old behaviour | new | greedy control |
|---|---|---|---|
| N=2 creative, MTP share | 19% | **98%** | 96–98% |
| N=2 creative, per-draft | 33.3–48.2% | **66.5–68.6%** | 62.3–67.7% |
| N=2 creative, group tok/s | 74.8 | **82.2** | 87.6 / 87.2 |
| N=2 creative, `chain` ms/round | (audit: 2.82) | 0.86–1.03 | 0.73–0.90 |
| N=4 creative, group tok/s | 108.2 | 106.0 | 113.9 / 109.9 |
| N=4 creative, per-draft | 30.6–52.3% | 39.5–75.0% | 42.5–67.6% |

At N=2 the creative-to-greedy gap closes from 14.6% to 5.7%. N=4 is flat because the planner speculates on only ~17% of tokens there in **both** arms — a separate finding, not this change.

**llmprobe `--bench-only --runs 3 --rungs 8k --sampling creative`**, old behaviour vs new, one boot each:

| cell | old | new | tokens per decode step |
|---|---|---|---|
| `--reasoning xhigh` | 82.1 tok/s | **86.6** | 2.45 → 3.08 |
| `--reasoning medium` | 86.9 tok/s | **101.3** | 2.87 → 3.56 |

## What is not established

- Cross-boot comparisons carry this box's 5–12% same-evening drift. The in-arm terms (`per_draft`, `acc_idx`, `corr`, `predraft`, `chain`, MTP share) are not affected and are what the claims rest on.
- n = 3 per cell; differences under ~10% in tok/s are not resolved. The 19% → 98% MTP share, the `corr`/`predraft` collapse and the per-draft deltas are far outside that.
- The `MTP_SAMPLED_PROPOSAL_MIN_TEMP` routing itself is pinned by a unit test, not by a throughput cell: a temperature-0.3 confirmation cell on the shipped build read 39.4% per-draft against the greedy control's 48.7% in a different boot, which is inside that cell's own run-to-run band and does not resolve either way. The crossover bracket comes from greedy-vs-sampled at the *same* temperature, which is the comparison that matters.
- Sampled output is not byte-comparable against today by construction (a different token sequence, the same distribution), so there is no byte bar for it. The greedy bar is `test_mtp_equivalence.sh`.
- Not measured: long context (every cell ran under 2k KV), N > 4, ragged widths, heterogeneous sampling inside one group (the homogeneity fallback is never exercised live), and `MLX_SERVE_MTP_BATCHED_QWEN4=1`.


beamivalice:
Amended in place (one commit, head 1bb49348): comment cleanup per the repo's comment rules, and one fix found while reading — the non-batched fallback in the group draft step proposed every sampled row with the first row's sampling params; it now uses each row's own. Full suite on the amended tree: 2537 passed, 0 failed.

ddalcu:

The acceptance math is sound, which is the part that had to be. `shortlistProposal` builds `q` from the same `filteredProbsRows` output the categorical draws from and then scatters it to full vocab, so `min(1, p/q)` divides by the density the draft actually came from — exactness does not require the 32-candidate shortlist to be the true top-32, only that `q` is the real proposal, and a coarse shortlist costs acceptance rather than correctness. `sampleResidual` clamps with `maximum(p - q, 0)`, so the tokens `q` can never produce are exactly the ones the residual covers, which is what the toy-vocab marginal test pins. The batched arm is correctly fenced too: `sampleRowsHomogeneous(...) and !sampleRowsAnySeed(...)` sends any seeded row down the per-row path, so the hand-rolled shared categorical in `shortlistProposalRows` cannot serve one.

- `src/generate.zig:10534` — the branch is based on `ef7d2977`, but `main` has since merged the seeded-draft-key change, so this PR's new proposal draw is written against the old contract: `sampleFromProbsLazy(q32, s)` is a two-argument call, while `origin/main:src/generate.zig:10656` now reads `fn sampleFromProbsLazy(probs, sampling, s)` and binds `seedKey(sampling)`.
  Failure scenario: as written this does not compile against current `main` — and the tempting conflict resolution, keeping the two-argument form the base had, silently un-fixes the seed. A request with `"seed": N` would draw its MTP drafts from MLX's global RNG again, which is the defect `CLAUDE.md` records as closed ("A `seed` binds EVERY sampler with a fresh key PER DRAW … the lazy path used MLX's global RNG"). The per-row path is the one a seeded request takes, because the batched arm already declines on `sampleRowsAnySeed`, so this is precisely the seeded case. Rebasing onto current `main` and threading `row_params` into the call is the whole fix; the PR body's "sits on top of #432 plus `ef7d2977`" is what makes it easy to miss that `main` moved past that base.

beamivalice:
Rebased onto current main (1386382d, which carries #427's seeded-draft-key contract) and threaded the key through: `shortlistProposal` now passes the row's sampling params into the three-argument `sampleFromProbsLazy`, and both callers hand it `mtpSamplingDraw(...)` so a seeded request's shortlist draws advance the per-draw counter like every other MTP draw. The batched arm is unchanged: it already declines any seeded row. Head a87a7c47, one commit; full suite 2552 passed, 0 failed.

ddalcu:

Re-reviewed after the rebase. The seed finding from `1bb49348` is fixed: `shortlistProposal` now takes `sampling` and both production call sites pass `mtpSamplingDraw(...)` (`src/generate.zig:5474`, `:5690`), so each shortlist draw gets its own `seedKey` with a bumped `draw` — a fresh key per draw, as `CLAUDE.md` requires. The remaining null-key categorical is the batched arm, and it stays fenced behind `!sampleRowsAnySeed(...)` at `:5682`. I also checked the two things I had left open: the `draft.suppress_mask = null` comment holds, because both shortlist builders run `rerankMaskCoarse` over the coarse readout before the top-32 select (`src/mtp.zig:1259`, `:1367`), so a reserved id is at `-inf` and cannot be a candidate; and the exactness argument survives `scatterShortlistQ`, since `draftTop32` yields distinct ids, `put_along_axis` therefore scatters each candidate's mass exactly once, and `sampleResidual`'s `maximum(p - q, 0)` covers the complement.

Two ownership gaps, both on MLX-error paths — which matter now that `installErrorHandler` (#353) makes those catchable instead of `exit(-1)`, so the server keeps running and keeps leaking.

- `src/generate.zig:10924` — `shortlistProposalRows`' `errdefer` frees `ids[0..done]`/`qs[0..done]`, but `done` is only advanced at the END of an iteration (`:10943`), so the handles the current iteration has already created fall outside it.
  Failure scenario: the `mlx_astype` at `:10937` or `scatterShortlistQ` at `:10942` fails on row j — `ids[j]`, allocated at `:10936` and by then a real array, is never freed by the unwind. One leaked `mlx_array` per failed batched draft step, repeating for as long as the condition that produced the MLX error persists. Setting `done = j + 1` right after `ids[j]` is built (and leaving `qs[j]` null-checked, which the `errdefer` body already does) closes it.
- `src/generate.zig:5686` — `mtpGroupSampledDrafts` assigns `c.draft_arrs[step]` for every row, but `c.n_drafted` is only bumped by the caller after the call returns (`:5819`), and `MtpPreDraft.deinit` frees just `draft_arrs[0..n_drafted]`.
  Failure scenario: the greedy loop at `:5686` writes rows 0..2, then `shortlistArgmax` or a later `shortlistProposal` fails on row 3; the error propagates out past `:5819`, the chains are torn down with `n_drafted` still at `step`, and rows 0..2's draft ids leak. The pre-existing `qwen4DraftSelectBatched` branch has the same shape, but its only fallible op is one `mlx_slice` — this function's window now spans a per-row shortlist argmax plus a per-row proposal, so it is worth advancing `n_drafted` inside the assigning loop rather than after it.

ddalcu:
🤖 **MLXServe-PRBot approved** — reviewed at head `2e396c64`, no issues found.

Both ownership gaps from `a87a7c47` are closed, and closed correctly. `done = j + 1` now moves the instant `ids[j]` exists rather than at the end of the iteration (`src/generate.zig:10941`), so an `mlx_astype` or `scatterShortlistQ` failure unwinds through it; the `errdefer` body's null checks cover the half-built row, since `qs` is `@splat(.{ .ctx = null })` and freeing the empty handle `mlx_array_new()` hands back is what the rest of the file already does. And `c.n_drafted` is now bumped inside each assigning loop in `mtpGroupSampledDrafts` (`:5665`, `:5688`, `:5694`, `:5703`) instead of only at the caller, so a mid-loop MLX error leaves `MtpPreDraft.deinit` a count that actually covers the ids already written. Every row still reaches exactly one of those assignments — greedy rows through the `q_probs == null` loop, sampled rows through whichever proposal arm the `batched` gate picks — and the caller's later `c.n_drafted = i + 1` is now redundant rather than contradictory.

PR #434, opened 2026-09-15
perf(mtp): one filtered verify block per group, one filter pass, a masked greedy verify

beamivalice:
Stacked on #433 (`perf/mtp-sampled-shortlist-drafts`, head `a87a7c47`) — review that one first; this
branch is `a87a7c47` plus one commit. Three items from an audit of the grouped (concurrent) MTP path
under temperature > 0, developed and gated one at a time on an M5 Max.

## What changed

**One filtered verify block per group, on the group's own eval.** `Transformer.forwardRowAxisVerify`
already projects the lm_head ONCE for every row of a group (`[batched] vocabulary verify projection
engaged`) and hands each row a slice. `verify_joined_logits` publishes that block; `mtpGroupVerify`
hands it to the round, and `mtpGroupSampledAccept` filters it once and builds every sampled row's
accept graph onto the single `mlx_async_eval` the greedy rows have shared since #412. Before, each
sampled row's `mtpRoundFinish` built its own `probsAllPositions` over its own slice, dispatched its own
eval and blocked on its own host reads — N build-wait-build staircases with the GPU idle across every
read. Nothing about acceptance moves: each row keeps its own acceptance mode, threshold and draw key,
and the accept test is still a host loop on the row's own prng. The arm declines to the per-row path
when the rows do not share one sampler, any row is seeded, or a row was padded past `1 + m`.

**Top-k and top-p in one pass.** A bounded `applyTopP` ranks the row's top `top_k` columns and
scatters its keep flags at exactly those column ids, so the `applyTopK` that every sampling site ran
first selected the same shortlist and masked a strict superset of the nucleus: one full-vocab
selection and one full-row `where` per filtered row, for nothing. `filterTopKTopP` skips it. Pinned
byte-identical to the old pair.

**The greedy spec verify masks reserved ids.** The sampled verify filters through `applySuppressMask`
and the rerank draft select takes the mask; the greedy MTP argmax took the raw logits at both sites, so
a flagged special or one of this pack's 243 padding rows past id 248077 could be committed as the
correction token — an id the serial sampler can never draw. `verifyArgmax` masks first, with `-inf` in
the logits' own dtype so the greedy block is not widened to f32.

## Bars

- `zig build test -Doptimize=ReleaseFast`: 9/9 steps, exit 0, on the squashed head.
- `MTP_FORCE_ENABLE=1 MTP_TEST_MODEL=<Flash Next> ./tests/test_mtp_equivalence.sh`: **11 passed, 0
  failed** (the temp-0 byte bar).
- New tests, each red first: a ragged three-row joined block whose per-row slice is byte-identical to
  that row's own `probsAllPositions` (the "one row cannot reveal an axis bug" class); the decline
  predicate; `filterTopKTopP` against the two-pass composition over `[m, V]` and `[1, L, V]` at
  V = 8192 and 248320, six (top_p, top_k) settings including the degenerate ones, a tie-quantized row
  and an all-equal row; a three-position block whose raw argmax is a padding id.
- Seeded reproducibility, live on this head: the same `seed` at temperature 1 with MTP forced on gives
  byte-identical output twice solo, twice at N=2 with one partner, twice at N=2 with a different
  partner, and all six agree with each other.

## Ledger

# Ledger — one PR for the grouped-MTP sampled path (G4+G3, G5, G7, G10)

Worktree `~/claude-tmp/wt-group-mtp`, branch `perf/mtp-group-sampled-accept`, base
`fork/perf/mtp-sampled-shortlist-drafts` = `1bb49348` (PR #433, amended mid-run; the rebase note is
in the report). Old-arm binary = a ReleaseFast build of `1bb49348` staged at
`~/claude-tmp/gp-old/zig-out/bin/mlx-serve`. Box: M5 Max 128 GB, macOS 26.6.2, port 11245,
`MLX_SERVE_ROUND_COST_PERSIST=0`, `--max-concurrent 4`, `MLX_SERVE_MTP_FORCE_DEPTH=2`,
`MLX_SERVE_MTP_TRACE=1`, 300 s empty-GPU wait per boot, 100 s idle before every cell,
killed by port match.

| # | candidate | verdict | sha | evidence |
|---|-----------|---------|-----|----------|
| A | G4+G3 — publish the group's joined verify logits, filter them ONCE, build every sampled row's accept graph onto the group's single async eval | **KEPT** | `c3905fb6` | gate 1 below |

### Gate 1 — base `a87a7c47` (arm `old`, boot 12:03) vs candidate A (arm `newA`, boot 12:19)
Forced depth 2, one boot per arm, 100 s idle before every cell, medians over that cell's
`[mtp-trace]` windows. The GREEDY cells are the in-boot control: candidate A touches only the sampled
accept, so greedy must not move, and across these two boots it does not (that is what licenses reading
the creative rows at ~1%).

| cell | round ms old | round ms new | delta | decode tok/s old | new | acc/round old | new |
|---|---:|---:|---:|---:|---:|---:|---:|
| N=1 creative (solo path, control) | 24.99 | 24.91 | −0.3% | 82.3 | 80.9 | 1.10 | 1.00 |
| N=2 greedy (control) | 37.56 | 37.25 | −0.8% | 106.2 | 108.6 | 1.08/0.98 | 1.08/0.98 |
| N=2 creative | 38.09 | 37.84 | −0.7% | 106.6 | 102.9 | 1.09/1.02 | 0.98/0.95 |
| N=4 greedy (control) | 61.93 | 61.88 | −0.1% | 130.7 | 129.8 | 1.08-1.11 | 1.08-1.11 |
| N=4 creative | 64.13 | 63.18 | **−1.5%** | 130.4 | 129.7 | 1.01-1.17 | 1.07-1.14 |

Drift-corrected against the same-boot greedy control: N=4 −1.4%, N=2 nil.

**The in-arm lap that decides it — the per-row `gap`** (ms, rows in publish order; `gap` is the
round-to-round wall a row spends outside its own round, which is where a later row's wait for the
earlier rows' finishes lands):

| cell | per-row gap |
|---|---|
| N=4 greedy, both boots | 0.05, 0.29, 0.21, 0.13 (old) / 0.05, 0.30, 0.21, 0.13 (new) |
| N=4 creative old | 0.06, **2.07, 1.39, 0.72** |
| N=4 creative new | 0.05, **0.31, 0.22, 0.14** |
| N=2 greedy, both boots | 0.04, 0.13, 0.04, 0.11 (old) / 0.04, 0.12, 0.04, 0.11 (new) |
| N=2 creative old | 0.04, **0.70**, 0.05, **0.71** |
| N=2 creative new | 0.04, **0.12**, 0.04, **0.12** |

The sampled rows' gap excess over their OWN boot's greedy control — 1.78 / 1.18 / 0.59 ms at N=4 and
0.57 ms at N=2 — goes to 0.02 / 0.01 / 0.01 and 0.00. That is the build-wait-build staircase, and it is
the lap the change targets; it moves by 20x its own spread while the control is unchanged.

**Verdict, stated both ways.** On the headline bar (round ms down more than 3%) candidate A does NOT
clear it: −1.5% at N=4, nil at N=2. On the second clause (an in-arm lap moving beyond its own spread)
it clears it decisively, and the mechanism is confirmed: the creative rows' per-row latency now equals
the greedy control's exactly. Kept on that basis; one `git reset` reverses it.
**Not resolved:** N=2 creative decode tok/s 106.6 -> 102.9. Its acceptance fell 1.09/1.02 -> 0.98/0.95
in the same cell, and tokens/round tracks that (2.09 -> 1.98), so this is a CONTENT fork on one rep —
an unseeded sampled arm draws different tokens once the draw ORDER changes, which is exactly what
moving the correction build does. Gate 2 carries two reps of this cell.

| B | G5 — top-k and top-p in ONE pass (the bounded nucleus already masks to the shortlist, so the top-k pass was redundant) | **KEPT** | `4969eb7a` | gate 2 below |

### Gate 2 — candidate A (arm `oldA`) vs candidate A+B (arm `newB`)
Boot order counterbalanced against gate 1: **A+B booted first** (12:44), A second (13:00). Two reps of
every creative cell. Medians over the cell's `[mtp-trace]` windows; `dec` = summed
`timings.predicted_per_second`.

| cell | A round ms | A+B round ms | A dec | A+B dec |
|---|---:|---:|---:|---:|
| N=2 greedy (control) | 37.53 | 37.47 | 106.5 | 106.4 |
| N=2 creative r1 / r2 | 38.31 / 38.09 | 37.67 / 37.76 | 107.0 / 109.0 | 109.5 / 108.9 |
| N=4 greedy (control) | 61.91 | 62.02 | 128.3 | 128.3 |
| N=4 creative r1 / r2 | 63.22 / 63.35 | 63.10 / 62.82 | 127.0 / 128.3 | 126.8 / 120.6 |

Sampled penalty (creative mean minus the same-boot greedy control): N=2 **+0.67 -> +0.25 ms**,
N=4 **+1.38 -> +0.94 ms**. The same ~0.42-0.44 ms at both N, which is what the mechanism predicts:
with candidate A the group filters ONE block per round, so removing one full-vocab selection and one
full-row `where` from that block is an N-independent saving. Greedy controls flat to 0.2%; the two
creative reps agree to 0.22-0.28 ms, so the 0.42 ms move is outside the measurement's own spread.
Round ms: −1.3% (N=2), −0.5% (N=4) — below the 3% headline bar, kept on the in-arm-lap clause plus the
byte-identity proof (a byte-identical change cannot cost quality).
**Resolves the gate-1 loose end:** N=2 creative decode tok/s across four cells on three boots reads
81.8, 84.4, 84.5, 87.5 while the N=2 greedy control sits at 86.8-87.2 in every one of them. The N=2
creative tok/s is dominated by content-fork variance; gate 1's 106.6 -> 102.9 was a low draw, not a
regression. The r2 cell of N=4 creative on A+B (dec 120.6, acc 0.951 against 1.04-1.07 elsewhere) is
the same effect.

| D | G10 — the greedy spec verify argmaxes over the reserved-id suppress mask | **KEPT** | `7cec29bb` | gate 3 below |

### Gate 3 — candidate A+B (`newB`, gate 2's first boot) vs A+B+D (`newD`, boot 13:24)
D is a correctness candidate, so the bar is: tests pass, the greedy byte bar holds, and the
forced-depth cost cells do not move.

| cell | A+B round ms | A+B+D round ms | A+B dec | A+B+D dec | acc A+B | acc A+B+D |
|---|---:|---:|---:|---:|---:|---:|
| N=2 greedy | 37.47 | 37.61 (+0.4%) | 106.35 | 106.39 | 1.023 | **1.023** |
| N=4 greedy | 62.02 | 61.98 (−0.1%) | 128.25 | 128.22 | 1.049 | **1.049** |
| N=2 creative (2 reps) | 37.72 | 37.70 | 109.5 / 108.9 | 109.8 / 106.5 | 1.10 | 1.08 |
| N=4 creative (2 reps) | 62.96 | 63.33 (+0.6%) | 126.8 / 120.6 | 124.3 / 124.9 | 1.00 | 0.99 |

The greedy acceptance is IDENTICAL to three decimals in both greedy cells (1.023 and 1.049) and greedy
decode tok/s is flat to 0.04%: the mask changes nothing about what this model actually emits at
temperature 0 — it only removes ids it could have emitted. Cost within ±0.4% on the greedy path (one
`where` over the row's own `[1, 1+m, V]` block in the logits' own dtype, so the block is not widened to
f32 either).

### Candidates not taken

| # | candidate | outcome |
|---|-----------|---------|
| G7 (seed contract) | thread a per-request key into the sampled MTP draws | **already upstream** — #427 landed `sampleFromProbsLazy(probs, sampling, s)`, `mtpSamplingDraw` and the keyed correction/bonus draws while this branch was in progress. The remaining global-RNG categorical on this path is the group's batched shortlist proposal, which declines whenever any row is seeded, so the contract holds by construction. Verified live rather than re-implemented (see Bars). No code from me. |
| G6 (dtype scalars) | `scalarOf(v, dtype)` at the f32-scalar sites so the verify block stays bf16 | **not attempted** — the audit measured it time-neutral (these kernels are launch-bound) and it would move the sampled path's numerics for a transient-memory-only gain. The one new site this PR adds (`verifyArgmax`) already uses a dtype-matched `-inf`. |

## Not verified

- **Long context.** Every cell ran under 2k KV, like the audit's.
- **N > 4, ragged widths, and a mixed group** (some rows greedy, some sampled, or two samplers in one
  group) — the homogeneity and padding declines are unit-tested, never exercised live.
- **`MLX_SERVE_MTP_BATCHED_QWEN4=1`** (the merged-forward group path) was left at its default off.
- **Sampled output is not byte-comparable across arms** and acceptance is a content fork: an unseeded
  sampled request draws different tokens the moment the draw ORDER changes, which is exactly what
  moving the correction build does. Every creative tok/s number here carries that variance (measured:
  the N=2 creative cell reads 81.8 to 87.5 tok/s across four cells on three boots while its greedy
  control sits at 86.8-87.2 in all of them), which is why the round cost and the per-row `gap` lap
  against a same-boot greedy control are the evidence and tok/s is not.
- The two perf items are **1-1.5% of the round each**, below the 3% headline bar this PR was gated
  against; they are kept on the in-arm-lap clause. The ledger states both readings per candidate.


ddalcu:

The two perf items check out. `verify_joined_logits` has exactly one writer (`forwardQwen4VerifyRows`, `src/transformer.zig:22734`) and it is only reachable from `forwardRowAxisVerify`, which calls `releaseJoinedVerifyLogits` at its top — so the block cannot be published twice without the first being dropped, and `deinit` covers the tail. The `offs` walk over the sampled rows and the projection's `offset += row.seq` walk only coincide when every row is sampled, and `use_joined` requires exactly that (`k == gens.len`) on top of a shape check, with a real concatenate otherwise. The prepared handover is safe even though it skips the Phase-4b dispatch, because every subsequent host read is still preceded by its own `mlx_array_eval`. `sampleRowsHomogeneous` compares temperature, top_k, top_p and the mask ctx, which is precisely the set `probsAllPositions` reads, so one filter really does serve the block. And `filterTopKTopP` is sound: `topRanksDescending(k)` selects the same columns from raw logits as from a top-k-masked row, the softmax is taken over those k values either way, and the bounded `keepByIndex` scatter masks everything off the shortlist — so the dropped `applyTopK` was genuinely subsumed.

- `src/generate.zig:4431` — the reserved-id fix lands on MTP's two verify sites but not on the other block decoders, which commit a raw greedy argmax the same way.
  Failure scenario: `nextDrafter` builds `verify_argmax` with a bare `mlx_argmax_axis` over `verify_logits`, and `mtpAcceptRowGreedy`'s counterpart there feeds it into `next_pending` -> `self.next_token_id` (`:4529`, `:4626`), so it is emitted, not merely compared against a draft. `nextDflash` has the identical shape at `:5018` -> `:5195`. On a pack with padding rows past the defined vocabulary, a collapsed position under `--drafter` or DFlash commits an id the serial sampler can never draw — the exact defect this PR writes up for MTP, with `self.sampling.suppress_mask` equally in scope at both. `CLAUDE.md` asks for a class guard when a live failure reveals a class, and the block decoders are treated as one family elsewhere (`generate.tokenStops` in all five). `nextPld:4032` and `nextDflash:4879`/`:4938` are fine and need no change — those argmaxes are acceptance tests and draft proposals, never the committed token.

beamivalice:
Fixed: `verifyArgmax` now covers every block decoder that COMMITS an argmax — `nextDrafter` and `nextDflash` built `verify_argmax` with a bare `mlx_argmax_axis` and commit `verify_argmax[accepted]` as `next_token_id`, so both are routed through it with the request's `suppress_mask`; the bot's exclusions are untouched, since PLD's per-position argmax and DFlash's draft and correction argmaxes are acceptance tests and proposals, never the committed token. The bar is the shared helper exercised at each decoder's own block width (a block whose raw argmax is a padding id must yield the best legal id per position, and the unmasked arm must still yield the padding id) plus a scoped source scan over the three committing decoders — red first on `nextDrafter` before the fix — because the drafter and DFlash rounds need a real checkpoint to reach; the greedy byte bar on normal prompts is unaffected by construction, since the mask only removes ids that were never legal (and the MTP equivalence run on the pre-fix head of this branch was 11 passed, 0 failed). One caveat on this push: the box's Xcode license agreement was reset part-way through (`xcrun` now answers "You have not agreed to the Xcode license agreements"), so after the two-site fix and its tests went green I could not re-run the full `zig build test` on the final head — that one bar is owed once someone accepts the license.

ddalcu:

The code half of the class fix is complete. All four sites that commit an argmax now route through `verifyArgmax`: `nextDrafter:4432` and `nextDflash:5020` (the two I flagged), `mtpRoundFinish:7414`, and `mtpGroupAccept:7072`. I re-checked that the exempt ones really are exempt — `nextPld:4033` compares its per-position argmax against `draft[i]` and never emits it, and `nextDflash`'s `:4881`/`:4940` are a correction proposal and a draft argmax — so the line the doc rule draws is the right one. Widening the instance test to widths 2/3/5 is the correct reading of the "one row cannot reveal an axis bug" class.

- `src/generate.zig:19193` — the new class guard scans three of the four committing sites, and the one it omits is the one this PR itself added.
  Failure scenario: the scan keys on the literal `verify_argmax` variable name (`mlx_argmax_axis(&verify_argmax` for the negative, `verifyArgmax(&verify_argmax` for the positive) and enumerates `nextDrafter`, `nextDflash` and `mtpRoundFinish`. `mtpGroupAccept` is neither in the list nor spelled that way — it builds into `var am` at `:7071` and publishes it as `st.verify_argmax`, which `mtpRoundFinish` then takes via `group_argmax` and feeds to `mtpAcceptRowGreedy`, whose `next_pending = am[accepted]` is emitted. So reverting `:7072` to `mlx_argmax_axis(&am, ...)` passes this guard green while every concurrent greedy MTP request can again commit a padding row — and that is the busiest of the four paths. This is the exact hazard `CLAUDE.md`'s "No source-scan tests" note describes ("they pin text, not behaviour, and pass against their own needles"); the needle here is a local variable name. Adding `mtpGroupAccept` with its own spelling closes it, though a guard that keys on the committed value rather than on identifiers would not have the hole in the first place.

beamivalice:
Full suite on the final head fe321e78 (the bar the previous reply listed as owed): 2561 passed, 171 skipped, 0 failed. Built and run once the box was free.

beamivalice:
Fixed by construction rather than with a fourth needle: `verifyArgmax` now RETURNS a `CommittedArgmax` and is its only constructor, and every site that commits a token from an argmax takes that type — the drafter's and DFlash's locals, `MtpRoundState.verify_argmax`, `mtpRoundFinish`'s `group_argmax` handover, and `mtpAcceptRowGreedy` (its pure core `mtpGreedyVerdict` keeps the hermetic verdict tests) — so `mtpGroupAccept`'s `var am` is covered and a raw `mlx_array` argmax no longer type-checks at any commit site, present or future; the source-scan test is deleted, since you are right that it passed against its own needles. The behaviour bar stays (the helper at widths 2/3/5: a block whose raw argmax is a padding id must yield the best legal id per position, and the unmasked arm must still yield the padding id) with one line saying the compile-time type is what makes the class complete, and CLAUDE.md plus the gotcha now state the rule as "a committed argmax is a `CommittedArgmax`, constructible only by `verifyArgmax`".

One bar is owed on this head: `zig build test -Doptimize=ReleaseFast` compiles every step clean and the individual new tests pass, but the aggregate GPU test step has been parked at 0% CPU for 100 minutes behind the benchmark server that holds the box, so I have not seen the full suite go green on `596a9f36`; the run is still waiting and I did not boot a server of my own.

ddalcu:
🤖 **MLXServe-PRBot approved** — reviewed at head `596a9f36`, no issues found.

Replacing the scan with a type is the right answer, and it closes the hole I raised: `CommittedArgmax` is now what every commit site takes, so `mtpGroupAccept:7103` — the site that spelled its argmax `am` and therefore slipped past the old needle — is covered by construction rather than by naming. `mtpAcceptRowGreedy` taking the wrapper while `mtpGreedyVerdict` stays the pure core keeps the existing verdict tests honest, and both were repointed. The only raw `mlx_argmax_axis` left on a block decoder is `nextDflash:4909` over `corrected`, which lands in `prev` as a draft proposal, not a commit — correctly exempt, as is PLD's acceptance-test argmax.

Ownership through the refactor is clean. `mtpRoundFinish` copies the wrapper out of the round state and nulls the source before its `defer`, so the group's handle has exactly one owner; `group_argmax` implies greedy (the group only builds one at `temperature <= 0.01`), so the `!group_argmax and !stochastic` arm cannot overwrite a live handle; and `verifyArgmax`'s own `errdefer` plus `mtpGroupAccept`'s per-row `errdefer` cover the partial-build paths that the old explicit `catch` did. The lengths handed to `ids()` are exact at both new call sites — `nextDrafter` builds `verify_input` as `[1, 1+m]`, and DFlash's row is `bs` wide with `m = bs - 1`.

Worth noting only as context, not as an objection: Zig has no private fields, so `CommittedArgmax{ .arr = ... }` is constructible in principle. But that takes deliberately writing the wrapper around a raw argmax, which is a far higher bar than a scan that a rename or a differently-named local defeats silently — which is exactly what happened here.

PR #436, opened 2026-09-15
disk-cache: a spec sidecar write must not arm the MLX latch for the next request (fixes #435)

### LINKED ISSUE #435: 500 "generation failed": a spec sidecar write raises inside mlx and the latched error is charged to the NEXT request
## What I see

Serving `qwen3.8-flash-next` (`qwen4_exp`) with MTP and the SSD prefix cache, roughly 4% of logged `/v1/chat/completions` requests answer `500 {"error":{"message":"generation failed"}}`. There is no pattern by prompt size: 13-token probes and 103k-token prompts both hit it. `/metrics` keeps reporting the server healthy, and the next request usually succeeds — so it reads as random.

## The chain in the server log

```
[mlx] expected a non-empty mlx_array at .../mlx/c/map.cpp:49
  [disk-cache] spec persist failed: MlxError — entry keeps its old spec
[scheduler] batched decode aborted: MLX failure (MlxFailure) — failing all 1 slots in the group
  -> GenerationFailed
```

Numbers from that log: 366 `[mlx]` lines, **every one of them this same message**, each immediately followed by the `spec persist failed` line; 343 `decode aborted`; 166 `-> GenerationFailed`. Of 3204 logged requests, 136 failed — and **136 of 136 had that `[mlx]` line in their own block or the block before it** (125 in the same request, 28 carried over from the previous one).

## Root cause: two bugs stacked

**1. An empty `mlx_array` reaches the tensor-map insert.** `insertSpecTensors` (`src/kv_disk_cache.zig`) inserts `a.aux_state` whenever `rows_ok` holds:

```zig
const rows_ok = hist == @as(c_int, @intCast(limit)) and
    (a.aux_state.ctx != null or a.qsa_pooled.ctx != null);
```

The `or` arrived with #381 (the QSA raw keys became a 32-row ring with the pooled bank as the history). `qsaHistoryRows` reports the checkpoint POSITION out of `qsa_hist_rows` even when `aux_state.ctx == null`, so a pooled-only head passes the guard and `mlx_map_string_to_array_insert` is handed an empty array. mlx-c raises (`expected a non-empty mlx_array`, `mlx/c/map.cpp:49`) and our error handler latches it.

**2. The raise is caught; the latch is not.** Spec persistence is best-effort by contract — "a failed write costs the entry its spec, never the entry" — so the caller logs the failure and keeps serving. But the MLX latch is process-wide, and `checkErrorDecode` consumes it once per decode tick. The next request's tick finds the previous request's latched error and fails itself. A harmless disk write turns into a 500 on somebody else's prompt, which is what makes this so hard to place from the outside.

## Fix

Opened as a PR (linked below): require the raw-history tensor for the head half, and drop the latch the best-effort write raised.

- `rows_ok` requires `aux_state.ctx != null`. `loadSpecSnap` already declines a head half without `h.aux`, so a pooled-only head should persist neither half of it.
- `writeSpecSidecar` drops the latch it raised, via the same `errorPending()`-before / take-after idiom `restore_dump.zig` already carried for its diagnostic dump.

The regression test builds a snapshot with `qsa_hist_rows` set and an empty `aux_state`; before the fix it reproduced the exact `map.cpp:49` message, after the fix it asserts the latch stays clear and the MTP KV half still loads.

## Environment

mlx-serve 26.9.3-dev (MLX 0.32.2), M4 Max, macOS 26.6.2. Launch: `--ctx-size 140000 --kv-quant 8 --prefix-cache-disk 24GB --prefix-cache-mem 8GB --prefix-cache-entries 128 --max-concurrent 1 --mtp --metrics`, env `MLX_SERVE_MTP_REGIME=0 MLX_SERVE_MTP_ADAPTIVE_MIN_KV=4096 MLX_SERVE_NGRAM_WARM=0 QWEN4_PLE_PREFETCH_PREFILL=1`.

I have not seen this reported anywhere, and `main` still carries both halves of it.

Sinojen:
Fixes #435.

## What changed

A head-half spec sidecar is now written only when its raw-history tensor exists, and a best-effort sidecar write can no longer leave a latched MLX error behind for the next request.

**`src/kv_disk_cache.zig`**

- `insertSpecTensors`: `rows_ok` requires `a.aux_state.ctx != null`. `qsaHistoryRows` returns the checkpoint POSITION from `qsa_hist_rows` even when the raw ring is gone, so a pooled-only head used to pass the `or` with `a.qsa_pooled.ctx != null` and hand an empty `mlx_array` to `mlx_map_string_to_array_insert` — which raises inside mlx-c (`expected a non-empty mlx_array`, `map.cpp:49`). `loadSpecSnap` already declines a head half without `h.aux` (`specAdoptPlan` will not adopt a pooled bank without its history), so persisting neither half is the consistent behaviour.
- `writeSpecSidecar`: drops the latch it raised. Every caller treats this write as best-effort by contract ("a failed write costs the entry its spec, never the entry"), but the latch is process-wide and `checkErrorDecode` reads it once per decode tick — that is how a failed sidecar write turned into a 500 on an unrelated request.

**`src/mlx.zig`** — `dropLatchedErrorUnless(had_error)`, the `errorPending()`-before / take-after idiom, moved next to `errorPending`/`takeError`. **`src/restore_dump.zig`** — uses the shared helper instead of its inline copy.

`CLAUDE.md` gets the two rules and `docs/gotchas/engine-mlx.md` the story.

## Why the head half is dropped instead of repaired

A pooled bank without its raw history is not restorable: `specAdoptPlan` returns `decline_head_no_history` when `head_aux` is absent, and the load path refuses `h.pooled` without `h.aux`. Writing a pooled-only head would produce a sidecar every restore rejects. The KV half of the sidecar is unaffected and still loads, so the entry is not lost — only its draft state, which is what "best-effort" already means here.

## Verification

- New test `DiskTier: a head snap with no raw-history tensor drops the head half and arms no MLX latch`. Red before the fix, and red for the right reason: it reproduced the production message verbatim (`MLX error: expected a non-empty mlx_array at .../mlx/c/map.cpp:49`). Green after, asserting no latch is pending, that the MTP KV half still round-trips, and that the loader declines the head.
- Full suite (`zig build test --summary all`, Debug): 2536 passed, 191 skipped, 1 failed. The single red is the live-disk test `volumeSpace: free is what the OS grants, never less than statfs' f_bavail` — outside this diff, and it depends on this machine's volume readings.
- Live: a server rebuilt from this branch (`zig build -Doptimize=ReleaseFast`) served requests including tool calls with zero `[mlx]` and zero `-> GenerationFailed` lines, against a pre-fix log with 366 raises and 136 failed requests out of 3204.

I could not re-run the suite on this exact rebase afterwards: the shared checkout picked up an unrelated in-progress edit in `src/scheduler.zig`. On the pre-rebase tree (identical hunks) the suite was 6/6 with 0 failures apart from the disk test above.


ddalcu:

The diagnosis and both mechanics are right. `mlx.check` only reads the C return code and never touches `mlx_error_latched`, so a raise really does leave the latch armed after the error has been converted and reported — and since every mlx call inside `writeSpecSidecar` goes through `try mlx.check` (`insertSpecArray:1660`, `insertSpecSlice:1671`/`:1682`, the `mlx_save_safetensors` at `:1594`), a raise always produces a Zig error, which is what makes `errdefer` sufficient rather than needing `defer`. Reading `errorPending()` before the op so a pre-existing latch survives is the right shape. The `rows_ok` tightening is correct and consistent: the `hist` block already null-guards `aux_state`, `qsa_rows` is the checkpoint position rather than a tensor witness, and `loadSpecSnap` declines a head half without `h.aux` anyway — so dropping the head half writes nothing the loader would have taken. The `h.lv{d}` marks loop is separately guarded by `mk.rows.ctx == null`, so it is not a second empty-array source.

- `src/kv_disk_cache.zig:1578` — the latch drop is scoped to the spec sidecar, but the sidecar is one of three mlx-raising writers inside the same swallowed call, so the rule this PR adds to `CLAUDE.md` is only half-applied.
  Failure scenario: `writeChunkFile` does the identical `mlx_map_string_to_array_insert` + `mlx_save_safetensors` pair (`:1930`, `:1934`) and `persistSsmCheckpoints` builds arrays through `try mlx.check(mlx_array_set(...))` (`:2243`). Their errors are re-raised out of `appendCommitWithSpec` — and then swallowed by every caller: `src/prefix_cache.zig:2049`, `:2175`, `:2304`, `:2338` each log and continue, because the disk tier is best-effort end to end ("entry stays resident"). So a Metal OOM while writing a KV chunk or an SSM checkpoint lands exactly where #435 landed: the request that caused it proceeds, and the next decode tick's `checkErrorDecode` fails an unrelated one. The new rule reads "a swallowed MLX failure must DROP the latch it raised", and these are swallowed MLX failures. One `had_error`/`dropLatchedErrorUnless` pair at the top of `appendCommit`/`appendCommitWithSpec` would cover the sidecar and both siblings in one place, rather than per writer.

Sinojen:
Agreed — the per-writer pair in `writeSpecSidecar` only closed the trigger we happened to hit, not the class. Pushed a6f1a89 which applies your suggestion: one `had_error`/`dropLatchedErrorUnless` pair at the top of `appendCommitWithSpecBounded` (the funnel both `appendCommit` and `appendCommitWithSpec` route through), so `persistSsmCheckpoints`, `writeChunkFile` and `appendSsmOnly` are all covered at the single swallow boundary. `writeSpecSidecar` keeps its own pair since it swallows internally instead of raising out.

To make the funnel guard testable I added a latching fault injector (`mlx.armLatchingFaultForTest(k)` — arms at the k-th `check()` in test builds and latches like a real mlx-c raise). The new test asserts both halves: a commit that raises drops *its own* latch, and a foreign pre-existing latch survives the call untouched.

The injector also paid for itself immediately: the fault path segfaulted on a pre-existing double-free — the `persistSsmCheckpoints` catch block deinited `chunk_sizes` while the `errdefer` right after its declaration still owned it. Fixed in the same commit.

Guards run on the rebased head: `zig build test -Dtest-filter="arms no MLX latch"` and `-Dtest-filter="drops the latch it raised"` both pass, full `zig build test` exits 0. The CLAUDE.md rule line from the first commit already reads "a swallowed MLX failure must DROP the latch it raised" — the funnel now applies it to all three writers, so the rule is no longer half-applied.

ddalcu:

Moving the latch drop to the `appendCommit` funnel is the right fix and it lands where I hoped: `:1146` covers `persistSsmCheckpoints`, `writeChunkFile` and `appendSsmOnly` in one place, while `writeSpecSidecar` keeps its own pair because it swallows internally rather than propagating. Reading `errorPending()` before the call so a foreign latch survives is pinned by the second half of the new test, and the latching fault injector is a genuinely better tool than the existing `fault` — it reproduces the state a real mlx-c raise leaves, which is the whole subject of the bug. Removing the manual `chunk_sizes.deinit` at `:1301` is correct: `errdefer chunk_sizes.deinit` at `:1255` already owns it, and the fault test finding that segfault is exactly the kind of thing the injector was worth building for.

- `src/kv_disk_cache.zig:1335` — the double free just removed from the `persistSsmCheckpoints` catch is still present, unchanged, in the `persistQsaHistory` catch one block below.
  Failure scenario: that arm calls `ssm_res.deinit(self.allocator)` and `chunk_sizes.deinit(self.allocator)` and then `return err` — which runs `errdefer ssm_res.deinit` (`:1305`) and `errdefer chunk_sizes.deinit` (`:1255`), freeing both a second time. `ArrayList.deinit` does not null its pointer, so this is a plain double free of the chunk-size buffer and of `SsmPersistResult.positions`/`bytes`; in ReleaseFast it is heap corruption rather than a clean crash. It is reachable from the same direction the new test drives — an MLX raise landing inside `persistQsaHistory` instead of `persistSsmCheckpoints` — and `armLatchingFaultForTest` with a larger `k` would reach it. Deleting the two manual lines, as `:1301` just did, is the whole fix.
- `src/kv_disk_cache.zig:1324` — pre-existing, unchanged by this PR, but in the same funnel: the "cap so tight nothing new landed" early return deinits `chunk_sizes` manually (correct — it is a success return, so no errdefer fires) and never deinits `ssm_res`.
  Failure scenario: `errdefer ssm_res.deinit` only fires on an error, and ownership of `positions`/`bytes` transfers to the entry only on the normal path further down, so every byte-capped commit that lands nothing new leaks the checkpoint position and byte slices. Small per commit, but this path fires repeatedly on a near-budget hybrid.

PR #437, opened 2026-09-15
Nested lists, rules, checkboxes, and a code ground that stops at the line

lojza3d:
### What the reader saw

A model writes outlines, checklists and rules; the transcript drew none of them.

- An indented item fell through to a paragraph, so every nested list arrived flat.
- `---` between sections rendered as three hyphens.
- `- [ ]` / `- [x]` rendered as a bullet followed by brackets.
- `~~struck~~` rendered as ordinary prose. Foundation parses it into an `inlinePresentationIntent`, like bold and inline code; nothing was reading that one.
- An item's own second line ("- Packaging" then an indented line under it) ended the list and started a paragraph back at the margin.
- An inline code span that wraps (a long path) tinted its first line out to the right margin.

### What changes

- Depth is a stack of indent stops (`MarkdownText.ListDepth`), not a count of spaces: models write two or four for the same one level, and both read as one. Any block that is not an item starts the depth again.
- A continuation line joins the item with U+2028, a break INSIDE the paragraph. With `\n` it still carried the item's paragraph style, which is why it looked right in the attribute tree and wrong on screen: TextKit starts a paragraph at every newline, a paragraph takes `firstLineHeadIndent` (the margin) plus its own spacing, and `headIndent` reaches only the lines that wrap. One blank line may sit inside an item, before its second paragraph; two end it.
- `---` (and `***`, `___`) is a one-cell `NSTextTable` with a top border, the same idiom the quote bar already uses. A line carrying anything else stays text.
- `- [ ]` / `- [x]` become `□` / `☑` markers drawn in `labelColor` rather than the bullet's secondary: a box is the item's state, not its punctuation.
- Strikethrough maps the intent, beside the inline-code one.
- `InlineCodeGround.rects` builds the tint per line fragment and ends each rect at that line's last visible glyph. The old rects came from `enumerateEnclosingRects`, which is selection geometry: a range continuing onto the next line takes its first fragment to the container's trailing edge.

### Out of scope

A fenced code block inside a list item still renders at the margin rather than indented under it, and a lazily continued line (CommonMark's unindented continuation) still ends the list. Numbered lists already kept the model's own numbers, and nesting them works where the markdown nests (each level its own `1.`); a model that writes decimal outline numbering instead (`1.1`, `1.1.1`) is not writing list markers at all, so those lines read as the parent item's own text.

### Tests

`MarkdownListTests` (nesting at two and four spaces, a third level, coming back out, the continuation line and that it is not a new paragraph, prose at the margin still ends the list, boxes, brackets that are not a box), `MarkdownBlockStylingTests` (the rule is a bordered block, all three spellings, a line with text is not a rule, strikethrough on and off), `InlineCodeGroundTests` (one ground per line, none past its line's own text, the break's space trimmed, an empty range draws nothing; headless `NSLayoutManager`). `bash app/test.sh`: 3422 tests, 0 failures.

Live: the same document rendered before and after, plus a narrower window so the code span breaks somewhere else.

Docs: rules in `app/CLAUDE.md`, story in `docs/gotchas/app.md`.

### Screenshots

Before / after, same chat.

md_before_regular

md_after_regular


ddalcu:

The parsing changes hold up. `ListDepth` behaves on the cases that matter — a list that starts indented reads as its own top, coming back out past the original stop re-anchors rather than going negative, and `indent == last` neither pushes nor pops — and the `depth.reset()` guard keys on `blocks.last`, which is correct because `.thematicBreak` is appended as a block and so ends a list the same way a paragraph does. The continuation loop terminates (`i = at + 1` with `at >= i`), correctly refuses a nested item (`listItem(in: next) == nil`), and the one-blank/two-blank rule falls out of the `at = blank ? i + 1 : i` lookahead. `isThematicBreak` sits after the fence collector, so a `---` inside a ```` ``` ```` block is still consumed there, and it cannot swallow a table delimiter (the `|`) or a `- -` (two marks). In `InlineCodeGround.rects` the trailing-whitespace walk is safe on a surrogate boundary — a lone half bridges to U+FFFD, which is not in `whitespacesAndNewlines`, so the loop stops rather than over-trimming — and the caller's existing band inset absorbs the switch from used-rect to line-fragment geometry.

- `app/Sources/MLXServe/Views/ChatView.swift:5734` — `level` is unbounded, so `firstLineHeadIndent` grows without limit on a deep outline.
  Failure scenario: `ListDepth.level` pushes a stop for every strictly-increasing indent and returns `stops.count - 1`, so a model writing a directory tree or a deep JSON outline as a nested list — each level two spaces in — yields level 11 at twelve deep and level 19 at twenty. `step` is `CGFloat(level) * 18` with no clamp, so that is 198pt and 342pt of `firstLineHeadIndent` plus the marker's hanging indent, against a transcript column that is a few hundred points wide in a narrow window or with the inspector open. The deep items get squeezed into a sliver at the right edge — the failure mode the comment on `listIndentStep` anticipates ("not so much that a three-deep outline runs out of column") but that nothing enforces. A `min` on the level, or clamping `step` against the container width, bounds it; the existing tests cover three levels, which is well inside the safe range.

lojza3d:
> `app/Sources/MLXServe/Views/ChatView.swift:5734` — `level` is unbounded, so `firstLineHeadIndent` grows without limit on a deep outline.
>   Failure scenario: `ListDepth.level` pushes a stop for every strictly-increasing indent and returns `stops.count - 1`, so a model writing a directory tree or a deep JSON outline as a nested list — each level two spaces in — yields level 11 at twelve deep and level 19 at twenty. `step` is `CGFloat(level) * 18` with no clamp, so that is 198pt and 342pt of `firstLineHeadIndent` plus the marker's hanging indent, against a transcript column that is a few hundred points wide in a narrow window or with the inspector open. The deep items get squeezed into a sliver at the right edge — the failure mode the comment on `listIndentStep` anticipates ("not so much that a three-deep outline runs out of column") but that nothing enforces. A `min` on the level, or clamping `step` against the container width, bounds it; the existing tests cover three levels, which is well inside the safe range.

Right, and a directory tree or a JSON outline written as a list is exactly the shape that reaches it. Fixed in 748b10f: ListDepth.maxLevel (5) caps the level the renderer is handed, so past six levels the items share an indent instead of walking off the right edge. The stack of stops is untouched, so it still tracks the real structure and coming back out lands where it should; only the number the paragraph style reads is clamped. Pinned by a nine-level outline asserting the indent holds past the cap and still steps below it.

ddalcu:
🤖 **MLXServe-PRBot approved** — reviewed at head `748b10fe`, no issues found.

The indent clamp is right, and right in the place that keeps the rest intact: `min(stops.count - 1, maxLevel)` bounds only the value `level(forIndent:)` hands back, while `stops` keeps tracking the real structure — so popping back out from a ninth level to a second still returns 1 rather than something derived from the clamped view. Five levels caps `firstLineHeadIndent` at 90pt plus the marker's hanging indent, which leaves column even in a narrow pane. The new test pins both halves of the behaviour, equality past the cap and a real step below it, and since every item carries the same `"• "` bullet the `headIndent` comparison is unaffected by the marker width. Nothing else moved from `6fd1c573`, which I reviewed above.

PR #438, opened 2026-09-15
perf(qwen4): fuse HC and GDN prefill operations

nikolai-vysotskyi:
Fuses Qwen4 HC residual-write/normalization/mixing and GDN convolution/normalization/gates during prefill. Both paths default on; `MLX_SERVE_HC_PREFILL=0` and `MLX_SERVE_GDN_PREFILL_FUSED=0` restore the composed paths. HC sizes derive from config/weights; inject stays a native matmul.

On Flash-Next mixed-4/8, the retained paths in #408 measured main → fusions: M4 Max 128 GB, llmprobe 16K **723 → 744** and 32K **728 → 747 tok/s** ([ddalcu's A/B](https://github.com/ddalcu/mlx-serve/pull/408#issuecomment-5674934215)); M5 Max 128 GB, cold 64K mean **1749.5 → 1849.5 tok/s** ([beamivalice's A/B](https://github.com/ddalcu/mlx-serve/pull/408#issuecomment-5658856761)). These are prior-arm measurements, not a new benchmark of this head.

Built from fresh main `d7452ad`. Local M5 Max 128 GB: ReleaseFast build passed; `zig build test` passed (2558 passed, 171 skipped, 0 failed; macOS 26.5). Compiled-reference parity covers HC=1..8, varying hidden sizes, cold GDN history, batch 2 and coalesced 8703-row chunks; negative cases cover unsupported geometry, dtype and disabled prefill.

Replaces #408 as requested. Fresh-head full-model A/B remains pending; this PR stays draft until that check.


beamivalice:
Review Bot:

One small commit on the PR, three edits, no behaviour change on the shipped pack:

1. Make the defer decision share the fused read's preconditions. In hc_prefill.zig, give eligible the inject weight and the stream dtype so it returns false when inject_flat is null or anything is not bf16. hcWriteOrDefer then stops deferring writes that hp.norm will decline a moment later. That is the only edit with runtime effect, and only on a pack with a quantized inject weight.
2. Rename the stale test reference in tests/CLAUDE.md line 121 from "at S 1..9 …" to "at decode and prefill widths".
3. Use one batch-2 bound for both arms: either make hc_prefill.eligible cap at max_seq / batch like gdnPrefillFusedFor, or drop the batch-2 arm from both since qwen4_exp prefill is batch 1. One line either way.

beamivalice:
Follow-up once this lands: 

The qwen4 MTP head's HC layer (qwen4MtpForwardLive / qwen4MtpForwardBatched) still uses plain hcRead + hcWrite. Its read already takes the new prefill arm through hcReadFusedFor, but the two writes stay composed because the head never defers. Threading hcWriteOrDefer / hcReadPending through the head folds them into hc_prefill.norm's WR arm. Depends on hc_prefill.zig from this PR, so it waits for the merge.

nikolai-vysotskyi:
@beamivalice Fixed the three points in this update:

- Prefill eligibility checks the dense inject handle/dtype and stream dtype. The defer caller supplies the **next read's** inject weight (attention→MLP, MLP→next attention), so null/quantized/FP32 inject and FP32 streams write immediately. No next read means no prefill defer. Decode/verify selection is unchanged.
- HC now caps sequence length at `max_seq / batch`, matching GDN; tests cover batch-2 widths4351/4352.
- Updated the stale test name in `tests/CLAUDE.md`.

The regression first failed at the pending-write assertion. Tests cover unsupported inputs, the positive BF16 defer path, and output equality after immediate write/flush. ReleaseFast passed; full Zig suite on M5 Max128GB/macOS26.5: **2559 passed,171 skipped,0 failed**. A fresh-head benchmark is next; the expanded context statistics for the retained combined implementation are running separately.

beamivalice:
Bring it out of draft I think this one is ready.

ddalcu:

(The body says this stays draft pending the fresh-head A/B, but GitHub reports `isDraft=false`, so I reviewed it.)

The kernel side holds up where it is easiest to get wrong. The norm kernel's threadgroup geometry is exactly matched by its guard: `hidden % 128 == 0` with `128 <= hidden <= 4096` puts threads at `H/4` in [32, 1024] and simdgroups at `H/128` in [1, 32], so `partial[32]` is precisely sized and simdgroup 0 is always full — which is what makes `if(sg==0)partial[lane]=0` actually zero all 32 slots rather than leaving garbage for the second `simd_sum`. Moving `S` out of the template and passing `seq` as an input is the right call per the template-JIT rule, and it is safe here because `GdnPreworkCfgKey`/`GdnNormGateCfgKey` still carry `seq` and `batch` — so neither the new `TAB` template arg nor the grid, both set inside the cache-miss block, can go stale when the chunk width changes. Both input-name lists were extended to match the widened `inputs_arr`. And the sigmoid table is a pure sweep of the bf16 bit domain rather than anything model-derived, so a process-wide singleton is right here in a way the grammar token table was not.

- `src/transformer.zig:33248` — the prefill norm-gate fusion is gated by the DECODE kill switch, so `MLX_SERVE_GDN_DECODE_FUSED=0` silently disables half of what this PR adds.
  Failure scenario: `gdnNormGateFused` opens with an unconditional `if (!gdnDecodeFusedEnabled()) return null;`, and this PR routes prefill widths through that same function by widening the guard below it. `gdnPreworkFused` scopes its decode check to `in.seq < 3` precisely so prefill is independent, which is what makes this look unintended rather than deliberate. Anyone A/Bing the decode fusion the way `CLAUDE.md` prescribes — flip the switch, interleave the arms in one process — now also loses the prefill norm-gate fusion, so the decode measurement carries a prefill regression it has no way to attribute. The new `MLX_SERVE_GDN_PREFILL_FUSED=0` lever is fine in the other direction: it declines through the range check. Scoping the decode gate the way prework does would make the two levers independent.
- `src/hc_prefill.zig:12` — `enabled()` re-reads the environment on every call while every sibling gate in this PR's blast radius memoizes.
  Failure scenario: `eligible()` calls `enabled()`, and `eligible()` runs per layer per prefill chunk; `gdnPrefillFusedFor` (`transformer.zig:33067`) has the same shape. `gdnPreworkEnabled` and `gdnDecodeFusedEnabled` sitting a few lines above both latch into `gdn_prework_env` / `gdn_decode_fused_env` on first read, so this is a deviation from the file's own convention rather than a general nit. It is bounded — a chunk is not a token, so this is dozens of environment walks per chunk, not per tick — but it is free to fix with the same `?bool` latch.

ddalcu:
🤖 **MLXServe-PRBot approved** — reviewed at head `31fb01b3`, no issues found.

Both findings are closed. Hoisting `prefill` and gating the decode switch on `!prefill` separates the two levers cleanly, and I walked each arm: at prefill widths with `MLX_SERVE_GDN_PREFILL_FUSED` on, neither the decode switch nor the S 1..9 range check applies; with that switch set to `0`, `prefill` goes false and the range check declines on `seq > 9`, so the composed chain really is restored rather than falling through to the decode arm; and at decode widths `prefill` is false because `seq < 17`, so `MLX_SERVE_GDN_DECODE_FUSED` keeps exactly its old meaning. The reorder also leaves `dv != 128` and the dtype decline below it untouched, so no guard was traded away for the independence.

The two env reads now latch into `?bool` the way `gdn_prework_env` and `gdn_decode_fused_env` do, and `gdnPrefillFusedFor` keeps the geometry test ahead of the memo and the override ahead of both — so a test override still wins and the environment is only consulted once a call actually reaches prefill geometry.

ddalcu:
Added some small code review fixes and merged, thanks @nikolai-vysotskyi , sorry it took so long

nikolai-vysotskyi:
Post-merge numbers from the machine this was developed on (M5 Max 128 GB, macOS 26.5, Flash-Next mixed-4/8).

Method: one binary built from main `d7452ad` + this PR (`223b09da...`), llmprobe 0.6.7 `--bench-only --rungs 16k,32k,64k,128k --runs 3`, `--kv-quant off`, `--prefill-chunk 8192`, prefix cache off, MTP/PLD off, one arm per boot.

| rung (actual input tokens) | fusions ON | fusions OFF | delta |
|---|---:|---:|---:|
| ~16.4k (17,085) | 1618 | 1488 | +8.7% |
| ~32.8k (32,820) | 1850 | 1628 | +13.6% |
| ~65.5k (65,468) | 1873 | - | |
| ~131k (131,167) | 1858 | - | |

The OFF arm stopped answering partway into the 64k rung (client saw `ECONNREFUSED`, no crash report on disk), so the control column ends at 32k. The arms ran back-to-back rather than interleaved, and the ON arm's own drift was -2.1% (steady). Decode sat at 51-55 tok/s in both arms; this protocol disables speculation.

Sanity check on the shipped build: 26.9.4 logs `[hc-prefill] engaged` and `[gdn-prefill] engaged`, and a user's llmprobe ladder on it reads 1880 tok/s at 16k with MTP on - the same plateau as the ON column above.

The four paths left out of this PR (QSA-pair, PLE ahead, PLE packed, MoE prefill group, HC upmix) still have no isolated number; as a combination they read 2163.9 -> 2323.4 tok/s on a 64,947-token uncached prompt in one HTTP screen. Tracking them in #460.

PR #439, opened 2026-09-15
perf(qsa): verify gather floor per KV scheme

ddalcu:
Flash Next decode under MTP dipped between 16k and 32k of context. Serial decode is monotone; the dip is the verify round. The verify-width QSA gather copies a fixed 14k-row union, and on dense KV that loses to the in-place mask arm until about 32k keys.

Fix: the floor is per KV scheme, 32768 on dense KV and 16384 on quantized (unmeasured, keeps the old value). `MLX_SERVE_QSA_VERIFY_GATHER_MIN_KV` still sets one floor for both.

All numbers M4 Max, Qwen3.8-Flash-Next mixed 4-8 bit, dense KV, 128 output tokens, temp 0, two reps.

**Before vs after, MTP fixed depth 6 (S=7)**

| prompt | prefill tok/s before / after | decode tok/s before / after | ms per verify round before / after |
|---|---|---|---|
| 17k | 748 / 750 | 45.2, 45.1 / 45.1, 55.7 | 66.0 / 62.1 to 64.5 |
| 34k | 737 / 735 | 46.9, 47.2 / 46.3, 46.7 | 68.0 / 68.8 |

Prefill is untouched (the change only affects verify widths). Decode tok/s at 17k is two values because the mask arm is not bit-identical to the gather, so one rep took a different greedy path with higher acceptance (2.46 vs 1.91 per round). Round cost is the honest metric: ms per round = tokens / tok/s / `[spec-stats] attempts`.

**Serial curve, no MTP, 3 reps each (unchanged by this PR, shows the dip is not serial decode)**

| prompt | decode tok/s | prefill tok/s |
|---|---|---|
| 1k | 56.0 | 750 |
| 2k | 54.0 | 801 |
| 4k | 52.5 | 830 |
| 8k | 51.8 | 787 |
| 16k | 51.4 | 764 |
| 32k | 50.8 | 747 |

Full suite 2552 passed, 0 failed. The M5 Max ladder should be rerun with the same arms before trusting 32768 there.

Also carries the app browser navigation changes that were staged in the tree.


ddalcu:

The per-scheme split itself is well argued: the union is fixed at `S * budget` rows, so what changes with the scheme is what the gather costs per row — a dense copy against an in-place mask read, versus a dequant that only touches gathered rows — and that is exactly the term the two floors separate. Keeping quantized at the old 16384 and calling it unmeasured is the honest framing, and the env override deliberately collapsing both back to one value is tested. On the app side I checked whether `resolveURL`'s new `file://` arm widens what a model can reach and it does not: `ToolExecutor` already registers a `readFile` handler, so local file content was already in the agent's reach and this is a preview convenience.

- `src/transformer.zig:5877` — the memo is now only written when the env var is set, so on every normal boot the floor lookup does a fresh `getenv` on the verify hot path.
  Failure scenario: the old body assigned `qsa_verify_gather_min_kv_cached` unconditionally before returning. The new one caches only inside the `if (getenv(...))` arm, so with `MLX_SERVE_QSA_VERIFY_GATHER_MIN_KV` unset — the default — `cached` stays null forever and both call sites re-walk the environment. Both are per layer: `:21250` in the forward's `want_blocks` and `:6181` in `qsaVerifyGatherAttn`. That is two environment scans per layer per verify round on a path this PR is measuring at ~65 ms/round with sub-2% effects. It also makes `warmQsaEnvCaches`' `_ = qsaVerifyGatherMinKvFor(false)` at `:4980` a no-op, which is the opposite of why that call exists. Caching the resolved no-env answer too (a sentinel, or one slot per scheme) restores it.
- `src/transformer.zig:6181` — the eligibility check asks a different question than the gate that leads to it, and they disagree on TurboQuant.
  Failure scenario: `:21250` decides `want_blocks` from `ctx.cache.config.scheme != .off`, while `qsaVerifyGatherAttn` derives the same flag from `kv_view.k_triple_q.ctx != null`. For the affine schemes those agree — both `denseView` arms set `k_triple_q` and `has_quant_triple` together — but the `.turboquant_2 / .turboquant_4` arms return `.{ .k = …, .v = …, .owned = … }` with no triples at all. So on a TurboQuant cache between 16385 and 32768 keys at verify widths, the forward reads "quantized", clears the 16384 floor and runs `qsaSelectBlocks` plus its `mlx_array_eval` for every layer — then the gather reads "dense", applies 32768 and declines `.kv_floor`. The select is paid and thrown away each round, which is a real per-layer cost on the exact path this PR is tightening. Nothing is computed wrongly, but `has_quant_triple` (the field whose own comment calls it the promise) or the scheme, used at both sites, would make the two agree.

ddalcu:
🤖 **MLXServe-PRBot approved** — reviewed at head `a4d205b9`, no issues found.

Both findings are closed, and the second one is closed for the right reason rather than just made consistent. Splitting the memo into `qsa_verify_gather_min_kv_env` plus an `_env_read` latch means the `getenv` happens exactly once whether or not the variable is set, so the default boot no longer re-walks the environment per layer and `warmQsaEnvCaches` does real work again — it is now the call that performs that single read.

On the predicate: `Scheme` has exactly four members and only the `.affine` arm of `denseView` sets `has_quant_triple`, so `config.scheme == .affine` at the gate (`:21255`, and the scheduler's `batchKvLenOfWith`) and `kv_view.has_quant_triple` at the eligibility check (`:6183`) are now the same question on every scheme, TurboQuant included. And the new comment gives the physical reason rather than only the equivalence: TurboQuant hands the arm a dense view, so the gather genuinely copies dense rows there and the dense floor is the correct one — which is what makes `== .affine` the right spelling instead of `!= .off`.

The only loose end is test-local and harmless: the floor test sets `_env_read = true` and its `defer` restores `_env` but not that flag, so a later call in the same test binary keeps the latched state. Nothing else reads this knob in tests and it cannot reach a serving path.

PR #322, opened 2026-08-31
Add Korean localization to MLX Core

hemanpapa-coder:
## Summary
- add Korean app metadata and a 687-entry `ko.lproj` localization catalog
- localize dynamic model, settings, agent, task, chat help, and media-generation strings through a small `L10n` helper
- add localization tests, including format-placeholder preservation

## Validation
- `cd app && swift test --filter KoreanLocalizationTests` — 4 tests passed, 0 failures
- built and launched the app on an Apple Silicon Mac
- verified the Korean welcome screen, chat, Settings, Agents, Tasks, model browser, and image/video/audio panes
- verified the separately installed Korean app bundle and its ad-hoc code signature

## Notes
- model/repository identifiers, user content, and raw server logs intentionally remain untranslated

PR #182, opened 2026-08-15
Chinese Localization

justinluque:
image

**Localization infrastructure and configuration:**

* Added a String Catalog (`Localizable.xcstrings`) supporting both English (`en`) and Simplified Chinese (`zh-Hans`), and documented its integration and testing in `app/CLAUDE.md`. The catalog is compiled and included in the app bundle by both the build script and the release workflow, with automated checks to ensure all languages are present and complete.
* Updated `Info.plist` and `Info-MAS.plist` to declare supported localizations and set the development region, ensuring the app advertises its language support to the system and App Store.
* Documentation  in `app/CLAUDE.md`, how to add new languages, and how the audit and build checks enforce coverage and prevent silent failures. 

**Code changes for localization:**

* Replaced hardcoded user-facing strings in models such as `ChatErrorNotice`, `ChatGreeting`, `ChatSessionTitle`, and `CommunityLinks` with `String(localized:)` calls, ensuring they are translated at runtime according to the active locale. 
* Added logic to localize only the displayed placeholder for chat session titles, while keeping the stored sentinel values in English for compatibility across languages and app versions.
* Updated number formatting in `ChatErrorNotice` to use the user's locale instead of always grouping by English conventions, making the display correct for all supported languages.

Testing:
**swift test: 2708 tests, 0 failures (25 skipped).**

PR #442, opened 2026-09-16
docs: add Simplified Chinese translations

LXD-8:
**What changed / why**

Adds Simplified Chinese docs: `README.zh-CN.md` plus `docs/zh-CN/` for the nine pages the README links to (cli, api, integrations, models, app, performance, building, faq, reasoning-protocols), and a one-line `English · 简体中文` switcher in `README.md` and in each of those English pages. Chinese readers currently have no translated entry point; the app UI is a separate effort (see #182, #322).

Docs only — no code touched. Every one of the ten pairs was checked mechanically against the English source (M5 Max, macOS 27.0): heading / code-fence / table-row / list-item counts and bold+italic runs identical; every relative link and every `#anchor` resolves on disk; the set of absolute URLs is unchanged; every inline code span, flag/env name and number in the English original still appears in the translation; every code block is byte-identical to the English block except for translated `#` comments. Sentences: 86 → 96 in the README (split for Chinese), 1.00 ratio in the other nine.

Docs only — no code touched, so nothing here can change what the binary does; `zig build test` is green on this branch anyway (run on an M5 Max with the Metal Toolchain and brew webp installed, per `docs/building.md`).

- [x] Searched open and closed issues, open PRs and recent commits; #182 (closed Chinese String Catalog) and #322 (open Korean `lproj`) are the existing threads


ddalcu:
🤖 **MLXServe-PRBot approved** — reviewed at head `4cec9a24`, no issues found.

I re-ran the mechanical checks independently rather than taking the description's word for them, and they hold.

Every relative link in `README.zh-CN.md` and the nine `docs/zh-CN/` pages resolves on disk. Inline code-span parity is exact in all nine pairs — 93/93 in `cli.md`, 89/89 in `api.md`, 69/69 in `integrations.md`, and so on, with zero English spans missing from any translation — which is the check that matters most here, since that set is where the flag names, env vars and defaults live. The six anchor links the Chinese README aims into `faq.md` match GitHub's slug rules for CJK headings (full-width `？` dropped, spaces to hyphens, ASCII lowercased), so `#mlx-serve-比-lm-studio-更快吗` really does land on `## mlx-serve 比 LM Studio 更快吗？`.

The switchers are symmetric: all ten English pages gained the two-line header, each `docs/zh-CN/` page links back to `../<page>.md`, and `README.zh-CN.md` carries the reciprocal line at line 16 — the same position as the English one — plus an explicit note that the English text governs on any discrepancy. Every internal link from the Chinese README points into `docs/zh-CN/` rather than leaking back to the English pages.

PR #443, opened 2026-09-16
Add Simplified Chinese localization to MLX Core

LXD-8:
**What changed / why**

MLX Core gets Simplified Chinese. `app/Sources/MLXServe/Resources/zh-Hans.lproj/` carries `Localizable.strings` (~1,060 entries) and `InfoPlist.strings`; both plists declare `CFBundleLocalizations` + `CFBundleDevelopmentRegion`; `L10n` resolves the strings SwiftUI cannot resolve on its own, at the call sites that pass a variable instead of a literal.

It is built the way #322 is — the shape you said you'd go with — so `ko.lproj` and `zh-Hans.lproj` can sit on the same plumbing: the helper and the call sites are language-agnostic, and the catalog is the only Chinese-specific part. `L10n` asks `Bundle.main`, so SwiftUI's own literal path, the preferred-language list and the per-app override in System Settings all agree (no second resolver, per the bot note on #322).

Not translated on purpose: model/repo ids, quantization labels, SF Symbols, log lines, and the English phrases `ChatErrorNotice` matches server errors against.

M5 Max, macOS 27.0, Xcode 27.0, webp 1.6.x from brew.

- `zig build test`: **9/9 steps succeeded, 0 fail** (`zig build test --summary all`, log attached).
- `cd app && swift test --build-system native`: **3474 tests, 24 skipped, 0 failures.**
- With the default build system this machine reports 5 failures, all in the LaTeX font/rendering group; the same group is red on unmodified `main` here (verified in a clean worktree — the SwaTex resource bundle lands where that layout does not look). Nothing in that set touches localization.
- Built `app/MLX Core.app` and checked what it actually ships: `plutil -extract CFBundleLocalizations` → `[en, zh-Hans]`, and lookups out of `Contents/Resources/zh-Hans.lproj` return Chinese. Screenshots: this process has no screen-recording permission on this Mac, so the evidence in this PR is the bundle probe (script in the thread) rather than images — happy to attach captures if you want them a different way.
- `LocalizationTests` (new) checks the catalog parses, resolves through a real bundle, keeps every placeholder, and that unknown keys fall back to English.

- [x] Searched open and closed issues, open PRs and recent commits; #182 (closed Chinese String Catalog) and #322 (open Korean lproj) are the existing threads


ddalcu:

`L10n` is the right shape for the model-produced strings SwiftUI cannot resolve on its own, and routing both paths through `Bundle.main` so the language is decided exactly once is the correct call. One thing worth stating so it does not get "fixed" later: I diffed format specifiers across all 1062 pairs in `Localizable.strings`, and the only real mismatches are the plural-suffix idiom — `"%lld tool%@"` → `"%lld 个工具"`, and five siblings. Those are correct and safe: Chinese has no plural suffix, the dropped `%@` is a trailing vararg that simply goes unused, and the leading `%lld`/`%@` still binds argument one. The remaining four flagged pairs are my regex catching `% s` / `% f` inside English prose like "~10% slower", not specifiers.

- `app/Sources/MLXServe/Views/TrayChrome.swift:90` — the lookup key is uppercased before the table is consulted, so every tray section header is guaranteed to miss and stay English.
  Failure scenario: `Text(L10n.text(title.uppercased()))` asks the table for `"SERVER"`, `"IN MEMORY"` and `"MEDIA GENERATION"` (the three `TraySectionHeader` call sites at `StatusMenuView.swift:247`, `:462`, `:522`), but the table is keyed on the source literals. I checked: `"Server"` is present and translated, `"SERVER"` is not — so that header is translated in the file and unreachable at runtime — and neither `"In Memory"` nor `"Media Generation"` appears in any casing. The uppercasing has to happen after the lookup (`L10n.text(title).uppercased()`), and the two missing keys still need adding. Worth sweeping the other non-literal `L10n.text` arguments while you are in there: `ModelSettingsSheet.swift:92` passes `ContextSizeDisplay.formatTokens(n)` and `WelcomeView.swift:633` passes `state.percentFormatted`, which are runtime-formatted numbers ("32K", "47%") being used as table keys — harmless today since they fall through to themselves, but they are lookups that can never hit.
- `app/Sources/MLXServe/Services/StreamCoalescer.swift:8` — a search-and-replace wrapped `.window` inside doc comments, so four comments now cite an expression that cannot compile.
  Failure scenario: `MenuBarExtra(.window)` became `MenuBarExtra(L10n.text(.window))` here and at `VoicePulse.swift:6`, `VoiceTrayDot.swift:9` and `VoiceTrayPanel.swift:115`. These are precisely the comments that record why the tray must never animate — the "tray locks up but the Picker still works" story — and `CLAUDE.md` treats a contract comment as a spec someone will read and act on. Nothing breaks at build time, which is what makes it easy to land; but it also means the replacement that produced them was unguarded, so it is worth confirming nothing similar reached a string that crosses a CLI or wire boundary.

LXD-8:
Thanks — both points are real. Fixed in two commits (`75076d6`, `4c06586`); the sweep's blast radius turned out wider than the two named sites, so here is the full accounting.

**Tray headers / uppercasing-before-lookup**

- `TrayChrome.swift` is now `Text(L10n.text(title).uppercased())`, with a doc comment stating that the caps are a render-time treatment so the lookup cannot migrate back inside. Added the two missing entries: `"In Memory" = "内存中"`, `"Media Generation" = "媒体生成"` — `"Server"` was already in the table and is now reachable.
- `AgentsWindow.swift:857` had the identical shape (`group.title.uppercased()`): all five `AgentToolGroup` titles are already translated, so they were dying on the vine too. Fixed. `:160` was already in the correct order.
- `StatusMenuView` passed `"\(count) models"` into the header's `detail` — the same class one frame up. `TraySectionHeader` now renders `detail` verbatim and the count is localized at its producer: `L10n.format("%lld models", …)` → `"%lld 个模型"`, following the existing `%lld tool%@` idiom you confirmed as safe.
- Swept every remaining non-literal `L10n.text` argument and dropped the doomed lookups: `ModelSettingsSheet.swift:92` (`formatTokens` → `"32K"`), `WelcomeView.swift:633` (`percentFormatted` → `"47%"`), `WebSourcesChip.swift:67` (a single uppercased domain-initial letter), `MediaModelChooser.swift:136` (its `downloadLabel` producers already `L10n.format` — this was a second lookup on already-localized text). All are now plain `Text(...)` on a String, which renders verbatim by construction.

**Wire boundary (your second note's generalization)**

`CLISetupInstructions.swift:214` rendered `tab.command` through `L10n.text` while the copy button (line 232) puts the raw `tab.command` on the pasteboard — a key collision there would silently fork the displayed command from the copied one. Now verbatim. Checked every remaining non-literal argument: all are display-only values (CLI/model display names, tab titles, install hints) that are meant to be table keys; nothing else feeds a pasteboard, request body, or launch argument.

**Comments**

Restored `MenuBarExtra(.window)` in all four. Repo-wide scan for `L10n` inside comment lines matched exactly those four, so the comment damage is bounded and now zero.

**Guards**

`LocalizationTests` gains two tests: a whole-sources regex rejecting `L10n.text(x.uppercased())` (verified red on the pre-fix code — it names `TrayChrome.swift`, then green after), and an assertion that every `TraySectionHeader(title:)` literal resolves to a translation through the `.lproj` bundle, reusing the existing `Bundle(path:)` harness.

**Verification**

`swift test --filter LocalizationTests`: 7/7 green, including the two new guards. The full `swift test --parallel` suite has pre-existing order-dependent failures on this machine (LaTeX rendering/fonts, ModelRoots/ToolModelRoots, TranscriptTypography) — the failing set changes between runs on identical code, and the same families fail on the unmodified head `c5208e9`; those suites read real model dirs and shared defaults and are unrelated to this change.

ddalcu:

Both earlier findings are fixed, and verified empirically rather than from the diff: zero `L10n` calls remain inside comments, every surviving `.uppercased()` now sits outside the lookup (`TrayChrome.swift:93`, `AgentsWindow.swift:160`, `:857`), and `"Server"`, `"In Memory"`, `"Media Generation"` and the new `"%lld models"` all resolve in the catalog. Reverting the runtime-data lookups was the right call too — `formatTokens(n)`, `percentFormatted`, `tab.command` and the favicon monogram are values, not keys. Catching the second uppercase-before-lookup site at `AgentsWindow.swift:857` on your own was the part I had missed. And `testTraySectionHeaderTitlesAreTranslated` is a real behavioural test: it resolves against the actual `zh-Hans.lproj` bundle rather than asserting on text.

- `app/Sources/MLXServe/Views/AgentsWindow.swift:143` — two section labels pass literals that are absent from the catalog, so they stay English exactly the way `"In Memory"` did.
  Failure scenario: `sectionLabel("Your agents")` here and `sectionLabel("Starters")` at `:148` feed the same `L10n.text(title).uppercased()` helper you just corrected, but I checked the catalog and neither `"Your agents"` nor `"Starters"` has an entry in any casing — so the Agents window renders two English headers above Chinese content. The five `AgentToolGroup` titles that flow through `:857` are all present, so this is just these two literals.
- `app/Tests/MLXCoreTests/LocalizationTests.swift:111` — the new class guard is narrower than its comment claims, which is why the above survived it.
  Failure scenario: the comment says it "guards the whole class, not the instance", but the pattern is `L10n\.text\([A-Za-z0-9_.]+\.uppercased\(\)\)` — one transform, and one argument shape. `.lowercased()`, `.capitalized`, or a `String(...)` wrapper all slip through, and the companion catalog test is scoped to `TraySectionHeader` call sites inside `StatusMenuView.swift`, so `sectionLabel` in another file is outside both. Asserting over every `L10n.text` argument that is not a string literal, or checking catalog coverage for all header helpers rather than one, would make the guard match the claim.

LXD-8:
Fixed — and this time the sweep belongs to the guard rather than to me. Three commits: `7b254af`, `20235b0`, `16f23d2`.

**The two labels, and the class they belong to**

Instead of adding `"Your agents"` and `"Starters"`, I enumerated the sources for helpers whose parameter reaches an `L10n` lookup, then checked every literal passed to them: **25 helpers, 43 literal call sites, 34 distinct strings with no catalog entry**. Beyond the two you named: `"THIS MAC"`, `"BEST MODELS FOR YOUR MAC"`, `"Sessions"`, `"Create"`, `"Width"`, `"Height"`, `"Steps"`, `"First frame"`, `"Last frame"`, `"Add image..."`, `"Add clip..."`, `"Add audio..."`, `"Images"`, `"Clips"`, `"Audio"`, `"Temperature"`, `"Repeat penalty"`, `"CFG scale"`, `"STG scale"`, `"Audio guidance"`, `"Decode"`, `"Prefill"`, `"Decode speed"`, `"Tokens generated"`, `"Used"`, `"Remaining"`, `"Model max"`, `"GPU-safe max"`, `"New"`, `"optional — ends here"`. All 32 are translated; the remaining two are allowlisted as Latin by design — the `"BASE"` HTTP verb and the `"Top-p"` sampling symbol. Wording follows the catalog's own choices (步数 from `"Steps (%lld)"`, 温度 from `"Temperature (%@)"`, 强度 from `"Guidance scale"`, 首帧/末帧 as in the first-frame note).

The same enumeration surfaced 16 more call sites of the value-not-key class you cleared last round (`L10n.text(String(format: "%.2f", …))` on slider readouts, a recorder's time string, a slider's own value). Those lookups are gone. Two of them carried prose — the Stop button and the clip-duration suffix — and now localize the sentence while the number stays verbatim, through `L10n.format` and the catalog.

**The guards now match their claims**

- The transform guard reads the whole argument with balanced parens (nested calls included) and rejects any shape: `uppercased`, `lowercased`, `capitalized`, or a `String(...)` wrapper. It also asserts the scan still sees more than 200 call sites, so a pattern that stops matching cannot pass quietly. This is what caught the 16 above.
- The coverage guard derives its helper list from the sources — every function whose parameter reaches `L10n.text/format` — rather than naming one, and checks every literal passed to those helpers in every file. The Latin-by-design values are an explicit two-entry allowlist, not an accidental gap.
- Both verified red on the unfixed tree and green after: the transform guard names `title.lowercased()`, the coverage guard names `"Starters" reaching sectionLabel`.
- One pass per file over a call index instead of a scan per helper: 0.4 s, down from the 18 s my first version took.

`swift test --filter LocalizationTests`: 7/7.

ddalcu:

Both findings from `4c065869` are closed, verified against the tree rather than the diff: `"Your agents"` and `"Starters"` now resolve in the catalog, the comment corruptions remain at zero, and all three `.uppercased()` sites keep the lookup first. The two rewrites in this push are the interesting ones and both are right — `L10n.format("Stop (%.1fs)", …)` replaces a key that was the already-formatted `"Stop (3.2s)"` and could never have matched, and `VideoGenView.swift:885` now sends only the translatable fragment through the catalog instead of the whole `String(format:)` result.

The rewritten guards are a real improvement over what I complained about. `callIndex` extracts arguments with balanced parens and walks *inside* argument lists, so nested `Text(L10n.text(…))` calls are actually reached; helper discovery now walks the sources instead of a hand-written list, which is what closes the `sectionLabel`-in-another-file gap; and both tests carry tripwires (`calls > 200`, `sites > 30`) so the scan cannot silently stop matching and pass green. That last detail is the part that makes a scan guard trustworthy.

- `app/Sources/MLXServe/Views/ModelBrowserView.swift:1556` — three runtime-formatted values are still used as catalog keys, and they pass the new guard because it enumerates transform spellings rather than asserting the argument is a literal.
  Failure scenario: `L10n.text(state?.percentFormatted ?? "")` here, `L10n.text(Self.formatTokens(appState.serverOptions.ctxSize))` at `SettingsView.swift:1554`, and `L10n.text(downloads.downloads[model.repo]?.percentFormatted ?? "")` at `MusicGenView.swift:314` are the same pattern this push reverted at `WelcomeView.swift:633` and `ModelSettingsSheet.swift:92` — "47%" and "32K" asked of the table. Nothing renders wrong, since the lookup falls through to the key, so this is dead work rather than a user-visible bug. The reason it matters is the guard: `testNoL10nArgumentIsTransformedBeforeResolution` checks the key for `.uppercased(`, `.lowercased(`, `.capitalized` and `String(`, and none of these three contains any of them, so the class the test exists for is live in the tree and green. Asserting that a non-literal argument is either a declared helper parameter or an allow-listed exception would catch the shape instead of the spelling.

LXD-8:
Both closed. Two commits: `9afdcab`, `fc20e43`.

**The three sites — and the 25 you did not have to name**

`ModelBrowserView.swift:1556`, `SettingsView.swift:1554` and `MusicGenView.swift:314` now render their values verbatim. Enumerating the same pattern across the tree found **28 call sites** in the class: durations (`progress.elapsedText(now:)`, `formatElapsed`), sizes (`preciseGB`, `MemoryInfo.format`), counts (`formatCount`), relative dates, a sandboxed path, `"2/5"` revision labels, schedule descriptions assembled by interpolation, and voice catalog names like `"Bella — English, female"`. Every one of them was a table read that could never resolve, so dropping the lookup is behaviour-neutral — the same string renders, minus the dead work.

What I did *not* drop are the producers that do reach the catalog on some branch: `statusLine`'s "Checking…"/"Unreachable", the variant menus' "Retry"/"Resume"/"Download", model and variant labels, tool names, preflight copy. They are now allow-listed explicitly, with the entry criterion written down: a branch that resolves in the table, the same bar as adding a key.

**The guard asks the shape question now**

Your read was exactly right — the old rule enumerated transform spellings, and `state?.percentFormatted ?? ""` contains none of them. An argument is now accepted only when it is:

- a literal, or
- an expression carrying a non-empty literal (`copied ? "Copied" : "Copy"`), or
- a plain member path (`m.lanDisplayName` — a runtime label the catalog does name), or
- a producer on `labelProducers`.

Everything else fails. `defect?.explanation ?? …` and `run.summary ?? run.status.label` are listed as expressions rather than callees, since their live branch is one side of the coalescing. Transform spellings are still rejected outright — that was a real class and costs nothing to keep — and first arguments are now split at a top-level comma, so copy containing one is no longer truncated.

Verified the way you asked for it: reintroducing the three sites makes the guard fail and name all three (`SettingsView.swift: L10n(Self.formatTokens(...))`, `ModelBrowserView.swift: L10n(state?.percentFormatted ?? "")`, `MusicGenView.swift: L10n(downloads.downloads[model.repo]?.percentFormatted ?? "")`), and it is green without them. `swift test --filter LocalizationTests`: 7/7.

LXD-8:
Correction to my previous comment, before anyone treats that sweep as behaviour-neutral: one of the 28 was wrong, and `ef35cc8` restores it.

`MusicGenView.swift:258` — `CoverWeightsFetch.modeLabel(t, decision:)` returns `MusicTask.label` verbatim on its ready / notApplicable branches, and all three task labels are catalog keys (`"Cover"` 翻唱, `"Text to music"` 文生音乐, `"Vocal to BGM"` 人声转背景音乐). Removing that lookup would have un-translated the mode picker's tabs, which is a regression, not dead work. It stays, and `modeLabel` joins the allow-list with that evidence.

I re-checked the other 27 against the same bar — does *any* branch return a catalog key — and they hold: those producers either compose ("2/5", "8.5 GB", "every 3 hours", sandboxed paths, "n/m") or return model/user data (agent names, music keyscales, neither of which is a key). So: 27 dead lookups removed, one producer kept. The "behaviour-neutral" claim in my last comment covered all 28 and was wrong for that one.

ddalcu:

The guard is now the right shape, and the comment on `testEveryLookupArgumentCanBeAKey` says so honestly: classifying the argument as "can be a key" (a literal, an expression carrying one, a plain member path, an allow-listed producer) versus "is a value" catches the property instead of the spelling, and it names `state?.percentFormatted ?? ""` as exactly what the old transform list let through. The allow-list is held to the same bar as adding a key — a branch that actually resolves — which is what keeps it from becoming a place to hide misses. Reverting the pure-runtime lookups is right across the board: sizes, counts, percentages, relative dates, `MemoryInfo.format`, `abbreviatingWithTildeInPath` are values, and the catalog can never hold them.

- `app/Sources/MLXServe/Views/ModelBrowserView.swift:1637` — this push took several strings that mix English prose with interpolation off the localization path entirely rather than splitting them, so they now render English in a Chinese UI.
  Failure scenario: the quant menus here and at `:1746` went from `L10n.text(selected ? "\(quant.label) — in use" : "\(quant.label) — use")` to the bare expression. Removing the lookup was correct — an interpolated result can never match a key — but the ` — in use` / ` — use` suffixes have no catalog entry in any form, so they stay English. Same for `BundleDownloadBar.swift:72` (`"Downloading: "` and `"Downloading model N/M: "`) and both chained-window sentences at `VideoGenView.swift:1048`; I checked each against the catalog and none is present. This PR already demonstrates both correct treatments — `L10n.format("Stop (%.1fs)", …)` at `SettingsView.swift:2693` for a whole formatted sentence, and looking up only the fragment at `VideoGenView.swift:885` — so these are the same shape, just not carried over. Worth noting the guard cannot see them either: `literalArgument` skips any literal containing `\(`, which is why interpolated copy falls outside both the lookup check and the catalog check.

ddalcu:
🤖 **MLXServe-PRBot approved** — reviewed at head `a269ea6f`, no issues found.

All six strings are back on the localization path in the right shape — the format string is the key, the runtime values are substituted after resolution. I verified each new entry exists in the catalog with exact specifier parity (`"%@ — in use"`, `"%@ — use"`, `"Downloading: "`, `"Downloading model %lld/%lld: "`, and both chained-window sentences), and the argument types line up with what the specifiers expect: `Int64` against `%lld`, `String` against `%@`. The `Int64(a.index)` / `Int64(chainWindows)` conversions are load-bearing there, not decoration.

One detail worth recording rather than flagging: the new `"Chained windows: %lld"` entry has no `L10n` call site, which looks like a dead key but is not — `Text("Chained windows: \(chainWindows)")` at `VideoGenView.swift:1045` is a SwiftUI `LocalizedStringKey` with interpolation, so SwiftUI resolves it against that exact key itself. That is the idiomatic path, and `L10n` is correctly reserved for strings produced by model types, as its own doc comment says.

Swept the whole catalog once more at this head: 1105 pairs, zero unexplained specifier mismatches once the deliberate plural-suffix idiom is accounted for, and zero `L10n` calls left inside comments. Everything raised over the previous four heads is closed.

Issue #444, opened 2026-09-16
File pickers open in the bottom left corner of the screen


lojza3d:
### Before opening

- [x] I searched open and closed issues, open PRs and recent commits for this, and it is not already reported or fixed

### Priority

low (minor annoyance)

### Area

Mac app (UI, downloads, model browser)

### mlx-serve version

26.9.3

### Hardware + macOS

M5 Max, Tahoe 26.6.2 (25G83)

## What happens

Every "Choose file…" / "Choose image…" button in the Create panes opens its Finder panel stuck to the bottom left corner of the screen instead of over the window. It happens in all four panes (Voice, Music, Video, Image) and in the 3D pane, on every click, with the app's main window centered on the display.

The composer's paperclip attachment picker is not affected: it comes up centered over the window, as expected.

## Where

`AppActivation.runModal(_:)` orders the panel front before running it:

```swift
panel.level = .modalPanel
panel.makeKeyAndOrderFront(nil)
let response = panel.runModal()
```

The `makeKeyAndOrderFront` line is there to keep the panel from coming up behind another app when one is frontmost.

A window that has never been positioned has its frame at the screen origin, and in AppKit coordinates the origin is the bottom left. Ordering it front is what puts it on screen at that frame, and `runModal` no longer gets to place it.

`AppActivation.beginPanel(_:completion:)` does not call `makeKeyAndOrderFront`, which is why the one picker that goes through it (the composer's) is positioned normally. That is the whole difference between the two paths.

Nothing is stored in the app's preferences for the panel's position, only its size (`NSNavPanelExpandedSizeForOpenMode`), so there is no saved frame to blame.

### Reproduction steps

1. Open any generation tab.
2. Click "Choose file..." in the reference field (image, video, audio...)

## Suggested fix

Give the panel a frame before it is shown, and keep the order-front that fixes the behind-another-app case:

```swift
panel.level = .modalPanel
panel.center()
panel.makeKeyAndOrderFront(nil)
let response = panel.runModal()
```

## Suggested guard

`AppActivationTests` already pins the app-activation ORDER through a fake that records its steps (`testAccessoryAppFlipsToRegularBeforeActivating`). The same shape fits here: a fake panel recording `position` / `orderFront` / `runModal`, asserting the panel is positioned before it is ordered front. Order is the whole bug, and a screenshot is not a test.

### Server log tail

```text
n/a
```

ddalcu:
🤖 **MLXServe-PRBot**

[...]

PR #445, opened 2026-09-16
Lay the Create panes out the same way, starting with Audio & Music

lojza3d:
## Why

The four Create panes grew one at a time, and it shows: the same ideas sit in different places, the same control is drawn three ways, and the model you are generating with can be anywhere in the form. Consistency here is not housekeeping, it is the product. Someone who opens Voice, then Music, then Video should not have to re-learn where anything is, and the first thing a new user sees of this app is one of these panes.

The rule the panes now follow is that **the model comes first**, because the model decides what everything below it means: whether there is a voice to clone at all, how long a reference it wants, whether source-audio modes exist, what Advanced even holds. Reading the model last is how you end up setting something the model cannot do.

## Scope

All five Create panes, in one branch, in stages. **Voice and Music are done**; Video, 3D and Image follow in this same branch, which is why this is a draft.

One PR rather than five because the point of the change is that the panes agree with each other. Landing them one at a time would leave the app internally inconsistent for as long as that took, which is the state this is trying to end. It does make the review bigger, and I am sorry about that; the commits are one pane each and the shared vocabulary is five small types, so it should read as "adopt the shared thing" rather than as five rewrites.

## Highlights

**Model first.** Model, then the inputs it governs, then the modality's own options, then Advanced, then Generate set apart from the settings it acts on. The model's transfer bar moved from beside Generate to under the model it belongs to, and "Keep model loaded after generating" moved out of Advanced onto the switcher's row, because residency is a property of the model, not of the output.

**One drop target.** Every audio slot is the grey well the picture panes already used, and the well now **stays** once the slot is filled. A block that disappears the moment you use it takes your place in a long form with it. Voice's reference well is split down the middle into its two ways in (pick a file, record one), so the rectangle you can drop a file on is the rectangle that offers to make one.

**Editors you can resize where it earns it.** The lyrics box is a document, and 90pt of it was a keyhole, so it has the video pane's drag handle and its height sticks. The handle itself is now one type shared by both panes.

**One chip.** The buttons above a text box (Enhance, Templates) and the icon buttons hanging off a typed field (the tempo menu, the seed dice) are one style, one grey, one corner.

**Advanced options** keeps one label and one chevron on the same side in both states, and the whole header row opens and closes it. Its controls are packed against each other and wrap when they run out of room, instead of sitting a grid column apart.

**The preview player** is one button that toggles between Play and Stop, and the disc animates while it plays.

## Behaviour, not just layout

Reading these panes closely turned up four things worth fixing:

- A **reference clip could reach a model that cannot clone a voice.** The pane hid the control, but the pane is not the only door: the chat tool builds its own request from the same saved path, and the service sent `ref_audio` unconditionally. The gate is in the request now (`AudioGenRequest.clonableReference`), the way the music service already gates its own fields.
- The **video pane's prompt drag handle tracked at half the cursor's speed.** `DragGesture` defaults to `coordinateSpace: .local`, which is the handle's own space; the handle sits below the box it resizes, so the space moves as the box grows and the translation subtracts its own effect. Both handles measure globally now. Both also set the cursor rather than pushing it, because a handle can be removed while the pointer is inside it and the balancing pop never arrives.
- A **focused number box ignored values nobody typed.** The guard exists so a repaint cannot move the caret mid-edit, but it also swallowed a tempo picked from the menu beside the box, which then appeared only when focus left. It now compares against what the box itself would read.
- `AudioClipPlayer.pause()` **left the history shelf lit** under a clip that had stopped making sound, and after the player button became a toggle nothing called it. It is gone, along with the comment that said a paused clip counts as playing.

## Shared components touched, and what that means elsewhere

Four types are shared with the panes not yet done, so please look at these with Image, Video and 3D in mind:

- `MediaModelChooser.pane(…)` gained an optional `accessory:` for the switcher's row. Default `nil`, so the other panes are unchanged.
- `MediaDropWell` gained an optional `caption:`, defaulting to what it said before; `MediaDropWellOption`, `MediaDropWellPair` and `MediaDropWellFilled` are new siblings that share its surface.
- `NumberField`, `OptionalNumberField` and `SeedField` went from `.caption` to `.body`, because a bezeled field is as tall as its font and a caption-sized box sat visibly below the pickers beside it. **This changes the Image and Video panes' number boxes too**, and the seed dice is a chip in all three now.
- `EditorResizeHandle` replaces the handle the video pane had written out inline.

## Tests

New pure helpers, each with its own tests: `SourceTrackLength` (a source track is as long as its clip, read out of the WAV header rather than by loading ~100 MB of samples), `NumberFieldRepaint`, `FlowLayout.rows`, `AudioGenRequest.clonableReference`. Plus the clamp-on-decode test for the new sticky lyrics height, next to the video twin that already had one.

Full suite green (`bash app/test.sh`), and both panes exercised by hand on macOS 26.6.2: generate, cancel, drop a file, record, pick from the network, switch models mid-form, relaunch and check the sticky state.

## Deliberately not in this PR

- **Image and Video have no history shelf at all** (Voice and Music share one, 3D has its own thumbnail strip). That is a missing feature rather than a layout inconsistency, so it wants its own issue and its own PR.
- **`outputFolderLink` is copied five times**, differing only in which folder it opens. An obvious candidate for the shared vocabulary, but it is pure cleanup and would widen the diff for no user-visible change.
- **A drop-target section reserves 6pt all round** for a dashed highlight that is usually invisible (`MediaDropModifier`'s `.padding(6)`), so a section that happens to be a drop target has more outer rhythm than one that is not. The Music pane compensates with a stated 8pt; drawing the highlight inside the section's own bounds instead would fix the cause, but it re-spaces all five panes.
- **File pickers open in the bottom left corner of the screen** (#444). Found while working here, one line, but unrelated to layout.

### Voice

Before:
screenshot_3674

After:
screenshot_3676

### Music

Before:
screenshot_3675

After:
screenshot_3677

## Notes

Rules and stories from this work are written into `app/CLAUDE.md` and `docs/gotchas/app.md`: what `.menuStyle(.borderlessButton)` does to a styled label, why a handle below its box needs a global coordinate space, why a focused number box has to compare against what it reads, and that `.frame(maxWidth: .infinity)` centres without an `alignment:`.


ddalcu:
🤖 **MLXServe-PRBot approved** — reviewed at head `8e487d2b`, no issues found.

A 3.9k-line layout change is reviewed at its seams rather than line by line, so what I actually checked: the new pure helpers and their guards. `SourceTrackLength.seconds(header:)` is the one piece parsing an untrusted file and it holds up — the slice is re-based when `startIndex != 0` so the absolute offsets are valid, `while offset + 8 <= d.count` covers both the id and size reads, the `fmt ` arm refuses a chunk whose declared size cannot cover the fields it is about to read (`size >= 16, body + 16 <= d.count`), every chunk step advances by at least 8 so a zero-size chunk cannot spin, and a `UInt32` size widened to `Int` cannot overflow the walk. `PromptMarkerInsert` clamps both ends of a stale selection. `VideoQualityMatch` is the pure mirror the view actually calls for its highlight (`VideoGenView.swift:462`), which is the factor-and-test shape `CLAUDE.md:92` asks for, and `testSomeTierIsActuallyClampedSomewhere` keeps the clamp arm from going untested. The 12 catalog additions carry no duplicate keys and consistent placeholders.

PR #447, opened 2026-09-17
Feature/benchmarks

ddalcu:
image


ddalcu:

This opens a publicly writable endpoint, so I checked that surface first and most of it is genuinely careful. The rules are create-only (`!data.exists() && newData.exists()`), so a row can never be overwritten or deleted by a client; every scalar is type- and range-bounded; and `"$other": { ".validate": false }` closes both the row and the `hardware` object against unknown keys. On the app side the upload is opt-in exactly as the comment claims — both `client.submit` call sites sit behind explicit buttons (`BenchmarkView.swift:367`, `BenchmarkSessionSheet.swift:92`), `sessionId` is a fresh `UUID().uuidString` per run rather than a machine identifier, and the community fetch fires only on `.task(id: pane)` when the Community pane is opened, not at launch. The website side is clean too: `esc()` at `website/benchmarks/index.html:403` covers `& < > " '`, and every community-derived string — `machineLabel`, `modelId`, the notes, the setting chips — goes through it before reaching `innerHTML`, so an attacker-controlled `note` or `armLabel` cannot inject markup.

- `database.rules.json:44` — `settings` and `flags` bound each value but place no limit on how many children a row may carry, or how long their keys may be, on an endpoint anyone can write to without authentication.
  Failure scenario: `"settings": { "$key": { ".validate": "newData.isString() && newData.val().length <= 60" } }` validates values only, so a crafted POST can create one "result" row whose `settings` map holds thousands of long keys — megabytes in a single write — and `flags` right below it is the same shape. Because the write rule is create-only, that row cannot be overwritten or removed through these rules afterwards, so the storage is permanent from the client side. The file already contains the fix elsewhere: `hardware` enumerates its keys and closes with `$other: false`. The client emits a closed key set too — `BenchmarkSettings.flatten` (`app/Sources/MLXServe/Models/BenchmarkModels.swift:356`) writes exactly `engine`, `version`, `kv_quant`, `kv_attn_mode`, `decode_attn_quant`, `prefill_chunk`, `drafter` and friends — so enumerating those two maps the same way costs nothing in what legitimate submissions can send.

PR #449, opened 2026-09-17
Expert streaming from SSD for Qwen3.8-Flash-Next: the bf16 checkpoint and quantized packs on a 128 GB Mac

beamivalice:
## Expert streaming for Qwen3.8-Flash-Next: serve the bf16 checkpoint (and any qwen4_exp pack) from SSD

Serves the original bf16 `Qwen/Qwen3.8-Flash-Next` (335 GB) on a 128 GB Mac by keeping the trunk resident and streaming the routed experts from SSD, and lets any quantized qwen4_exp pack do the same under a resident budget. `--ssd-budget-gb N` is a total resident target in GiB (trunk + workspace + expert cache); `--expert-cache-gb` is the low-level knob; the per-model `ssd_budget_gb` setting rides `model-settings.json` and the app's Model Settings sheet. Streaming is gated to `model_type == qwen4_exp`; a pack loaded without a budget is byte-identical to main.

### Mechanisms (one commit; kill switches in the levers table)
- **Engine** (`src/expert_stream.zig`): byte-span store over the checkpoint's expert banks (fused bf16 HF layout, or the MLX pack's nine slices per expert with each tensor's width solved from its packed shape), per-layer group-exact LRU (every hit of a route is protected before any admission), a prefill union bridge (hits copied from the cache, misses read, admission by frequency into empty slots), one MLX barrier per MoE layer with the slab lease released at the next layer's readback, a resident-budget ledger with admission against the wired limit, MTP refused at the door (an explicit `--mtp`; a per-model `mtp` setting is dropped with a warning).
- **I/O** (`src/expert_io.zig`): F_NOCACHE positioned-read pool (fd cache validated by inode/mtime, 64 MiB coalescing, byte-midpoint partition, direct pread when page-aligned), page-aligned slabs with epoch leases, zero-copy import through `mlx_array_new_data_managed_payload` verified by pointer identity (a non-aliased import is a named load error; MLX counts the slab once).
- **Compute**: bf16 selected-expert Metal kernels (`src/expert_bf16_kernels.zig`, gate/up at every row count, down from two rows; parity vs fp32 truth no worse than the `gather_mm` composite at every shape); a streamed quantized pack runs the resident fused quantized kernels over the slab with remapped ids.
- **KLD tooling**: `mlx-serve kld capture` (a teacher fixture: prompt/generated ids + full-vocab f32 logits per position) and `kld compare` (teacher-forced KLD / top-1 / NLL, all positions and to the teacher's first end-of-turn token, per-position JSON). `MLX_SERVE_NGRAM_BF16_DIR` serves any pack with the original bf16 n-gram table.
- **Surfaces**: `/v1/models` rows carry `streaming`, `streaming_required` and `ssd_budget_gb` at top level; every refusal is a named 503/400 (`expert_streaming_required`, `ssd_budget_below_resident`, `ssd_budget_exceeds_wired_limit` naming the `iogpu.wired_limit_mb` that would admit, `expert_streaming_mtp_unsupported`, `expert_streaming_unsupported_layout`); the app shows an SSD badge and a budget row for streaming rows and never restarts the server on a named refusal.

### Verification
- Hermetic suite: 2684 passed, 0 failed (main + 120 new tests, every mechanism red-first; executor reports and the codex audit + fix round are in the journal). Release exe builds; `swift build` clean.
- `tests/test_bf16_streaming.sh` on the final tree: boot at 60 GB, greedy determinism, a 4k prompt, and a `kld compare` against a bf16 fixture within the envelope (PASS).
- Identity: the quantized pack streamed at `--ssd-budget-gb 50` reproduces the resident load's bytes and top-20 logprobs on greedy prompts (at equal prefill chunk width) and its `kld compare` digits on all 60 prompts / 3,840 positions.
- Quality (teacher = this engine's bf16 streamed path, 16 prompts x 512 raw continuation, to-EOS): experts 8-bit g64 (streamed, 157 GB pack) 0.035, 6-bit g128 0.044, 5-bit g64 0.050, the 4/8-bit pack 0.081, 3-bit g128 0.163, iQ 3.3bpw 0.199; bf16 kernels-off vs on 0.0012 (60x64 chat). KV 8-bit costs nothing measurable.
- Speed, quiet box, two boots (`--ssd-budget-gb 60`, bf16): short-prompt decode 6.4 cold / 8.6 warm tok/s, 4k prefill 160 tok/s, warm 10-prompt loop median 6.5, decode hit 92%, 0.37 GB of fills per row, RSS 64 GiB. The 4/8-bit pack streamed at `ssd 50`: warm loop 24 tok/s vs 66 resident, prefill 507 vs 2110.

### KLD vs this engine's bf16 (teacher = the streamed bf16 checkpoint's own greedy output; 16 prompts x 512 tokens of raw wikitext continuation; 7,186 positions scored to the teacher's first end-of-turn; dense 8-bit g64 in every pack; weights exclude the n-gram table)

| Quant | Weights (GB) | Mean KLD | Top-1 | NLL |
|---|---|---|---|---|
| experts 8-bit g64 + bf16 n-gram table (streamed, ssd 80) | 136.1 | 0.0353 | 0.943 | 0.361 |
| experts 6-bit g128, dense 8-bit | 101.5 | 0.0443 | 0.936 | 0.370 |
| experts 5-bit g64 + bf16 n-gram table | 89.8 | 0.0445 | 0.937 | 0.371 |
| experts 5-bit g64, dense 8-bit | 89.8 | 0.0501 | 0.932 | 0.378 |
| 4/8-bit pack + bf16 n-gram table | 75.3 | 0.0749 | 0.920 | 0.397 |
| 4/8-bit pack (experts 4-bit g64, dense 8-bit) | 75.3 | 0.0814 | 0.913 | 0.407 |
| 4/8-bit pack, KV 8-bit | 75.3 | 0.0818 | 0.914 | 0.404 |
| experts 3-bit g128 + bf16 n-gram table | 55.2 | 0.1580 | 0.879 | 0.494 |
| experts 3-bit g128, dense 8-bit | 55.2 | 0.1628 | 0.877 | 0.501 |
| iQ 3.3bpw (experts 3/2-bit imatrix, dense 8-bit) | 54.3 | 0.1990 | 0.865 | 0.542 |

Two rows use the original bf16 n-gram table (`MLX_SERVE_NGRAM_BF16_DIR`) instead of the pack's 4-bit one; the 8-bit row streams because the 157 GB pack cannot load resident.

### Known limits
- Streaming is text-only (the vision tower is not loaded) and never speculative (MTP refused by name).
- A warm bf16 decode forward is ~150 ms = 68 route (the bf16 trunk's own GPU work) + 78 fill (latency-bound, <1 miss/layer); the bf16 kernels are a wash at one row.
- The quantized case pays the per-layer barrier (a resident forward is ~15 ms, the streamed route term alone 49).
- The forward-level bf16 byte-identity fixture skips until a tiny HF+pack pair exists; a quantized pack loaded resident shows no budget row in the app (set the per-model key or the flag to stream it).
- CLAUDE.md is over its 100 KB policy (135 KB on main, +1.7 KB here).

### Levers
| env | default | effect |
|---|---|---|
| `MLX_SERVE_EXPERT_BF16_KERNELS=0` | on | the `gather_mm` composite instead of the bf16 kernels |
| `MLX_SERVE_NGRAM_BF16_DIR=<hf dir>` | unset | the original bf16 n-gram table for any pack |
| `MLX_SERVE_QSA_*`, `MLX_SERVE_MTP_*` | unchanged | untouched by this PR |


ddalcu:

At ~11k lines this is more than one pass can cover exhaustively, so I scoped to the classes `CLAUDE.md` calls out as the expensive ones here: launch-flag plumbing, per-model settings, and the zero-copy slab lifetime. Those check out. `--ssd-budget-gb` and `--expert-cache-gb` are both in `main.zig`'s match list with named range refusals, retained on the Scheduler (`scheduler.zig:169-171`) and carried into the cold-load `LoadRequest` (`:2021-2023`), so neither is flag-eater material and `ensureLoaded`'s second construction site sees them. The per-model `ssd_budget_gb` rides the single `applyModelSettings` chokepoint (`:2562`) rather than being read at a site, which is what keeps both load paths in agreement. And refusing a non-aliased import by name (`error.ExpertSlabImportCopied`) is the right shape — a silent copy there would double the resident cost the budget exists to bound.

- `src/expert_stream.zig:842` — when a slab's payload is never released, the leak that results is invisible outside the test suite.
  Failure scenario: `SlabOperand.destroy` frees the array, then spins up to 32 rounds of `mlx_synchronize` + `mlx_clear_cache` waiting for `payload.released`. If it never lands it bumps `slab_release_timeouts`, sets `self.* = undefined` and returns — without `slab.destroy()` or `allocator.destroy(payload)`. Leaking rather than freeing is the correct call, since freeing a slab MLX still owns would be a use-after-free of GPU-visible memory, so the branch itself is right. The problem is that `slab_release_timeouts` has exactly three references in the tree: the declaration at `:803`, this increment, and one assertion at `:1774`. It is never logged, never in `/props`, never in `metrics.zig`. So a page-aligned expert slab — the largest allocation this feature makes — can leak repeatedly while the resident-budget ledger keeps reporting the budget as held, and the operator sees nothing. Given the file already logs its other declines by name, a one-line warning on the first timeout (and ideally the counter beside the other expert-streaming figures in `/props`) would make the one failure mode that defeats the budget observable.

beamivalice:
Verified the bot's finding: `slab_release_timeouts` had three references (declaration, increment, one test) and nothing surfaced it. 311c4ebd logs a warning at the timeout with the bytes left mapped and puts `slab_leaks=` on the per-forward `[expert-stream]` breakdown line. The leak-instead-of-free branch itself is unchanged, as the bot agreed it should be. Rebased onto main 3c6206d9; suite 2685 passed, 0 failed.

ddalcu:
🤖 **MLXServe-PRBot approved** — reviewed at head `311c4ebd`, no issues found.

The finding is closed, and closed in both places it needed to be. The timeout branch now warns by name before returning (`expert_stream.zig:843`), quoting the exact byte count that stays mapped and the running total — so the first occurrence is legible rather than inferred — and `slab_leaks` rides the per-forward `[expert-stream]` line at `:1207`, which makes it continuously visible instead of only at the moment of failure. Naming the bytes is the part that matters: it is the number that reconciles a drifting resident footprint against a budget the ledger still believes it is holding.

The placement is right too — the warning sits inside the `released == 0` branch, after the 32-round spin and before the early return, so it fires exactly when the slab is being abandoned and not on the normal path that reaches `slab.destroy()`.

Scoping this head the same way as the last: the only change to this PR's own work since `28c5f86b` is those four lines in `expert_stream.zig` — everything else in the head delta is `main` being merged in (the localization and benchmark PRs that landed meanwhile). My earlier review remains scoped to launch-flag plumbing, per-model settings and the slab lifetime rather than the full ~11k lines, and nothing in those areas moved.

ddalcu:

The delta since `311c4ebd` is tight and all of it is an improvement. `ExpertSlabImportCopied` is now carried the whole way by name — `loadErrorFromName` recognises it, `loadRefusalFor` gives it a type and a message that asks for the Mac model and OS version, the app's `LoadFailureVerdict` allowlist accepts it, and each hop has a test — which closes the half of "a load failure crosses the inference thread by NAME" the previous head left open: the error existed at the import site but nothing downstream could name it. `mtpUnderStreaming` replacing the two-argument `mtpRefusal` at the decision point is also right, and its four input combinations are each pinned (`expert_stream.zig:2288-2291`), so the `refuse` / `drop_settings` / `off` split cannot drift. I checked the `kld` flag change too: `--no-mtp` is still accepted by the parser at `:110` even though the usage text now advertises only `--mtp`, and every `--no-mtp` in `tests/` targets the server rather than `kld`, so nothing that passes it breaks.

- `src/server.zig:12867` — the `ssd_budget_below_resident` refusal still bills the user for an MTP head that this path guarantees is never resident, and now contradicts the usage text the same commit corrected.
  Failure scenario: the message reads "leaves no room for an expert cache after the resident trunk, **the MTP head**, the prefill union and the fill buffers." But `mtp_resident` is a hardcoded `false` at both streaming sites (`scheduler.zig:1919` and `:3750`) and `resolveExpertCache` adds `if (mtp_resident) split.mtp else 0`, so that term is structurally zero here — which is exactly why this same delta removed it from `--ssd-budget-gb`'s help in `main.zig`. The two now disagree, and the one a user actually hits is the refusal. Someone raising the budget to make room reasons about a head that is not there, which is the shape `CLAUDE.md` records from the FLUX case: the guard was right and the message sent the user looking for a problem that did not exist. Dropping those three words makes the refusal quote the terms it actually compared.

beamivalice:
Agreed: `ssd_budget_below_resident` billed a head that is never resident under streaming (`mtp_resident` is fixed false at both scheduler sites and now at the kld site). Commit `db107552` drops the three words so the refusal names only the terms it compared, matching the corrected `--ssd-budget-gb` help. Hermetic suite 2685 passed / 0 failed on the new head.

ddalcu:
🤖 **MLXServe-PRBot approved** — reviewed at head `db107552`, no issues found.

The refusal now names the three terms it actually compares, and the two user-facing texts agree: `server.zig:12867` reads "after the resident trunk, the prefill union and the fill buffers", matching `--ssd-budget-gb`'s help at `main.zig:339`. That is the same set `resolveExpertCache` sums once `mtp_resident` is false, so someone raising the budget is now reasoning about terms that exist.

I swept the rest of the tree for the same staleness and the remaining "MTP head" mentions are all correct in their own context. `scheduler.zig:2687` is the comment on `resolveExpertCache` itself and qualifies it properly — "the MTP head (only when it stays resident)" — which is right, because that function stays general even though both streaming callers pass false. `kld.zig:171` says "refused under streaming". Neither is a budget breakdown shown to a user hitting this error.

Nothing else moved in this head.

beamivalice:
Rebased onto `ef5e667d` (was conflicting): the five commits are unchanged in content and order; the one conflict was both sides appending test blocks at the end of `src/transformer.zig`, resolved by keeping both. Gates on the rebased tree: hermetic suite 2687 passed / 0 failed, release exe builds, Swift build clean, 115 app tests pass in the serial arm.

One note unrelated to this PR: on `main` itself, `LaTeXPackagingTests.testEveryBuildPathPatchesSwaTexFontLookupBeforeCompiling` now fails because `ef5e667d` removed the `swift build` step from `.github/workflows/ci.yml` while the test still requires every build path to contain one.

ddalcu:
🤖 **MLXServe-PRBot approved** — reviewed at head `2d986fe9`, no issues found.

This head is a rebase, not new work. The branch's own patch against `main` is unchanged from `db107552` — same 37 files, same 10,906 insertions and 120 deletions — and diffing the two patches shows only `index` hashes, `@@` offsets and context lines moving. What moved underneath is `main` absorbing `ef5e667` (the ds4/GGUF benchmark fix), which is why the head-to-head delta looks like it touches the benchmark views and CI.

The one place worth checking on a rebase is where that concurrent commit edited a function this PR also edits, and it composed cleanly: `ModelSettingsSheet.formHeight` now reads `var n = isGguf ? 1 : 2` from `main` alongside this PR's `if showSsdBudget { n += 1 }`, and the two gates are orthogonal — `showSsdBudget` keys on `live?.streaming`, which only a qwen4_exp checkpoint sets, while `isGguf` routes to ds4 or llama.cpp. They cannot both fire, and if they somehow did the row count would still be right.

Everything I approved at the previous head stands.

beamivalice:
CI's `Run Zig tests` step was hanging on this PR (70+ min, where `main` finishes in ~15). Root cause: a test helper in `src/expert_io.zig` (`benchPrint`) wrote to fd 1, and one always-on test printed a line through it. Under `zig build test` stdout is the build runner's protocol pipe, so the runner read that line as a message header and waited forever; the test process sat idle in `test_runner.mainServer`. Standalone runs of the test binary never showed it. Commit `0f39cfef` points the helper at stderr (one line, six call sites). `zig build test --summary all` on the new head: 9/9 steps, 2687 passed / 174 skipped / 0 failed in about 3 minutes; `main` in the same form: 193 s.

ddalcu:
🤖 **MLXServe-PRBot approved** — reviewed at head `0f39cfef`, no issues found.

One commit on top of `2d986fe`, and it is a real fix rather than cosmetics: `benchPrint` wrote straight to fd 1, which under `zig build test` is the build runner's protocol stream, so bench output would interleave with — and corrupt — the runner's IPC rather than just being noisy. Moving it to fd 2 is the right descriptor.

I checked the fix covers the class rather than the instance: `std.c.write(1, …)` and `getStdOut` now appear nowhere in any of this PR's new files (`expert_io.zig`, `expert_stream.zig`, `expert_quant.zig`, `expert_bf16_kernels.zig`, `kld.zig`). The only other raw-descriptor write among them is `kld.zig:633`, which writes to an opened file handle in a drain loop rather than a fixed descriptor, so it was never in scope.

Nothing else moved; everything approved at the previous heads stands.

beamivalice:
CI on `0f39cfef`: the Zig step now completes (the hang is gone); the run fails only in the Swift step on `LaTeXPackagingTests.testEveryBuildPathPatchesSwaTexFontLookupBeforeCompiling`, which fails identically on `main` since `ef5e667d` removed the `swift build` step from `ci.yml`. Head `7264fe15` adds the CLAUDE.md rule and gotcha for the stdout/test-runner trap; no code change.

ddalcu:
🤖 **MLXServe-PRBot approved** — reviewed at head `7264fe15`, no issues found.

Documentation only: the one-line rule in `CLAUDE.md` plus the story in `docs/gotchas/engine-mlx.md`, which is the split the growth policy asks for — and the entry is 10 lines of defect / cause / fix / guard, inside the 20-line bar, with no round-by-round log.

The mechanism is worth having written down because it explains an asymmetry that would otherwise look like flakiness: under `zig build test` the binary runs with `--listen=-`, so fd 1 is the runner's message stream; a stray line is read as a message header, the runner computes a bogus body length and blocks, while the test process sits in `mainServer` waiting for a request that never comes. Standalone, the same binary passes because stdout is just a terminal. That matches the fd 1 → fd 2 fix in the previous head exactly, and naming the tell (a `zig build test` that never returns while `test-build` passes, `sample` showing `mainServer` in `readSliceAll`) is the part that makes it findable next time.

I checked the new rule holds tree-wide rather than just in this PR's files: no `std.c.write(1, …)` or `getStdOut()` remains anywhere under `src/`, so it is being stated against a codebase that already satisfies it.

beamivalice:
New commit on top: the per-layer expert cache is LFU by default (`MLX_SERVE_EXPERT_LFU=0` restores LRU), with the route being resolved protected from eviction for the whole resolve. Lossless: the policy only changes which experts stay resident.

Quiet box, `--ssd-budget-gb 60`, short prompt, three interleaved boots per arm, medians:

| Short prompt | LFU | LRU |
|---|---|---|
| Warm decode tok/s | 10.47 | 8.69 |
| Cold decode tok/s | 6.28 | 6.42 |
| Forward 64: fill ms / hits | 28.6 / 451 of 480 | 34.4 / 442 of 480 |

Greedy bytes identical across all twelve answers. Hermetic suite 2689 passed / 0 failed. The PR body carries the same table plus a note on the KLD floor: against an f32-stream teacher of the same weights, the 8-bit row scores 0.030, and the two bf16 kernel arms sit 0.016 and 0.017 from that reference and 0.010 from each other, with the streamed path reproducing itself exactly.

beamivalice:
Correction to the previous comment: the loop A/B is in and LFU loses on independent prompts, so the default goes back to LRU and LFU is opt-in (`MLX_SERVE_EXPERT_LFU=1`), commit on top. Same protocol, three interleaved boots per arm, medians:

| | LFU | LRU |
|---|---|---|
| Short prompt, warm decode tok/s | 10.44 | 8.66 |
| Short prompt, cold decode tok/s | 6.25 | 6.40 |
| 10 independent prompts, loop median tok/s | 5.91 | 6.56 |
| 4k-context decode tok/s | 5.07 | 5.12 |

Cause: the frequency counts never age, so one prompt's hot set outranks the next prompt's fresh experts. An aged variant (count decay, or a reset per request) is being measured; it becomes the default only if the loop no longer loses. The PR body table is updated to match.

beamivalice:
Withdrew the two cache-policy commits (LFU and its opt-in flip): the PR is back at `7264fe15`, its state before them, and the body is restored. The decode-speed work continues on a local branch and will come as its own PR once it clears its bar.

ddalcu:
🤖 **MLXServe-PRBot approved** — reviewed at head `7264fe15`, no issues found.

Two commits adding an opt-in LFU policy. The part worth checking is the new `victim(keep)`, because it introduces a `return best orelse 0` that would evict a protected expert if every occupied slot were protected — and `0` is not even chosen, it is just the fallback. That path is unreachable: `admission_cap = capacity - hits` and `admit_count = @min(admission_cap, misses)`, so at the k-th admit the protected-and-resident count is `hits + k <= hits + admit_count - 1 <= capacity - 1`, leaving at least one unprotected or empty slot every time. Empty slots also short-circuit ahead of the protection test, so a slot evicted for being not-ready is picked up immediately.

The other way that could break is a duplicate expert in `union_ids` — the hit/miss pass runs before any admission, so a repeat would be counted as two misses and admitted twice, leaving `expert_to_slot` and `slot_to_expert` disagreeing. It cannot happen: the union is built with the `last[expert] == i` last-occurrence filter (`:382`), so each expert appears exactly once with its multiplicity carried in `counts`.

The lever reaches, which is the failure I check for on opt-in knobs: `expertCachePolicyFromEnv()` is read at `:1118` and passed into `initPolicy` at the real construction site `:1120`, not left on the `init` default. It also treats `MLX_SERVE_EXPERT_LFU=0` and an empty value as off rather than as "present therefore armed", which is the `getenv != null` trap.

Worth recording that `keep` changes the default LRU path too, not only LFU: a hit in the current route can no longer be evicted by an admission in the same round. That makes the "every hit of a route is protected before any admission" property in the PR description actually true rather than aspirational, and it can only avoid a re-read, never cause one. Both policies are covered by tests (`:1482` LFU, `:1502` the LRU/LFU comparison).

PR #457, opened 2026-09-18
Load Prism Hadamard packs (Bonsai 2, model_type prism_hadamard_qwen35)

h9q2cyxvgm-ui:
Prism's Bonsai 2 27B MLX pack (`prism-ml/Ternary-Bonsai-2-27B-mlx-2bit`) stores its quantized projections in a blockwise Hadamard-rotated basis. The bytes are ordinary affine 2-bit/g128, so a stock loader accepts them and emits fluent garbage; no MLX server runs the pack today.

The pack needs, per Prism's reference runtime: `y = qmm(WHT_1024(x ⊙ signs))` on every quantized linear incl. `lm_head`, and the inverse (`signs ⊙ WHT_1024(row)`) on embedding rows. Signs are per input width (the `<module>.signs` tensors are identical within a width).

- `model.zig`: accept the type in the qwen3_5 arm; read the block from the `modules` manifest, refuse mixed/unvalidated blocks.
- `transformer.zig`: width-keyed sign table from `*.signs` at init; rotate in `qmatmul` (gated on quantized weights), invert in `rawEmbedding`.
- `mlx.zig`: `mlx_hadamard_transform` extern.
- `model_discovery.zig`: allowlist entry, so the pack is registered, not just loadable via `--model`.

Ordinary checkpoints are untouched. Validated on an M5 Max: greedy output byte-identical to PrismML's llama.cpp fork on three prompts, 3.5k-token needle exact, thinking correct, negative control (loaded as plain `qwen3_5`) garbage. 42.6 tok/s plain decode; the pack ships no drafter. Tests for the manifest parser and discovery; `zig build test -Dtest-filter=prism_hadamard` green.

Follow-up: a bf16 path for the transform.


h9q2cyxvgm-ui:
Superseded by 89eeb24, which landed the same contract with native kernels and the MTP graft while this was in flight. For what it's worth, this independent port validated the premise: greedy output byte-identical to PrismML's llama.cpp fork on the PQ2_0 pack, and the same weights loaded as plain qwen3_5 emit garbage. Closing.

PR #461, opened 2026-09-18
fix: fold a non-leading system on /v1/responses to stop the tool-schema double-bill

kmahara:
## Summary

On `/v1/responses`, a `system`-role message that is **not** at index 0 reached the chat template unfolded. Codex sends exactly this (a top-level `instructions` system plus a non-leading system in the input), Qwen-family templates raise on it (`System message must be at the beginning`), and the raise drops the request into the **silent generic fallback** — losing the stop token and re-billing the expanded tool schema **once per system**, which duplicates the schema and inflates the prompt.

`/v1/messages` already folded this unconditionally; `/v1/responses` did not.

## Root cause

`responses.parseInput` returned `pi.messages` without folding, then `cachedFormatChat` fed them straight to the template. `chat.foldSystemMessages` (already used on the `/v1/messages` path) was never applied to the Responses path.

## Fix

Call `chat.foldSystemMessages` at the end of `parseInput`, before returning — the **same unconditional fold** `/v1/messages` applies (family-agnostic contract: every template we serve raises on a system turn that is not first). The joined string is registered in `pi.owned_strings` for `deinit`. The native template now renders: no schema duplication, stop token restored.

## Verification

Captured Codex request (full system + 294 functions), raw-byte template dumps before/after:

| | before | after |
|---|---|---|
| rendered prompt | 874,122 B | 463,436 B |
| system markers | 2 | 1 |
| tool-schema blocks | 2 | 1 |
| fallback WARN | 1 | 0 |

Saved **410,686 B (~401 KiB, ~1.89×)** per request; the rendered bytes match the server's logged `prompt: N chars`. Fewer prompt bytes = less prefill work (prefill tok/s A/B not run; the prompt reduction is the measured quantity).

- Regression test: `parseInput folds a non-leading system into the leading one` (red on revert: expected 2, found 3).
- Full `zig build test` suite green. (One pre-existing `kv_disk_cache` `volumeSpace` cell is a disk-activity flake, green on isolated rerun — unrelated.)

`CLAUDE.md` and `docs/gotchas/tool-calling.md` updated so the fold is documented at both parse sites.


ddalcu:
🤖 **MLXServe-PRBot approved** — reviewed at head `c7bcf77c`, no issues found.

PR #462, opened 2026-09-18
feat(app): add an in-app UI language setting

LXD-8:
**Base:** `main` @ `18cac51`

**What changed / why**

Settings ▸ Interface gains a **Language** row (System / English / 简体中文), so a per-app override no longer needs System Settings. `AppChrome` sets `\.locale` on every scene root and points `Bundle.main` at the chosen `.lproj` through `BundleLanguageOverride` (L10n and AppKit menus). `L10n.text` reads `LanguageLookupRevision`, an `@Observable` counter bumped only when the bundle actually changes, so observation re-runs exactly the bodies that built copy while their `@State` — a half-typed draft, a scroll position — survives. System, an unknown value, or a missing `.lproj` leaves lookups as they were.

- M5 Max, macOS 27.0: `swift test --no-parallel` with `SDKROOT=MacOSX26.5.sdk` — 3567 tests, 0 failures; `main` @ `18cac51` baseline — 3558 tests, 0 failures. `LanguageSettingsTests` covers the preference value (empty/unknown → system), labels, `.lproj` resolution and no-match pass-through; `AppearanceSettingsTests` scans every window scene for `.appChrome()`.

| before | after |
|---|---|
| before | after |

- [x] Searched open and closed issues, open PRs and recent commits; no existing thread for this


ddalcu:

- `app/Sources/MLXServe/Services/AppearanceSettings.swift:189` — switching the language updates literal copy but leaves every `L10n.text`/`L10n.format` string in the old language, because `BundleLanguageOverride.languageBundle` is a plain global that no view observes and `apply` runs *after* the render pass that the preference change triggered.
  Failure scenario: with the main window open, pick 简体中文 in Settings ▸ Interface. `Text("…")` literals re-resolve against the new `\.locale` and turn Chinese, but the ~371 `L10n.text(...)`/`L10n.format(...)` call sites under `Views/` produced their `String` before the swizzled bundle was swapped, and nothing invalidates them — the window renders half Chinese, half English until each of those views happens to be rebuilt for an unrelated reason (a tab switch, new chat message). `init()`'s launch-time `apply` covers cold start, so only the live switch shows it.

LXD-8:
Fixed in e6ab978, and the ordering diagnosis was exactly right.

The row now applies the language and *then* writes the preference (`AppLanguage.select`): the write is what re-renders, and the `L10n` lookups run inside that render, so swapping the bundle afterwards left every string it had just built in the previous language. The redundant post-render `onChange` on the modifier is gone — it could never have fixed this — and `LanguageSettingsTests` pins the new seam (the choice is both resolved and recorded).

Screenshots are in the body now, from this tree running locally.

ddalcu:

The ordering fix is right: `AppLanguage.select` swaps the bundle before the `@AppStorage` write, so every `L10n` lookup that runs during the resulting render sees the new `.lproj`. One gap is left.

- `app/Sources/MLXServe/Services/AppearanceSettings.swift:175` — only `AppChrome` and `InterfaceSectionContent` declare a dependency on `InterfacePrefKey.language`, and nothing at all observes `BundleLanguageOverride.languageBundle`, so a subtree that resolves its copy through `L10n` (a plain `String`, not a `LocalizedStringKey`) has no invalidation source when the choice changes — only the views that read `\.locale` re-render.
  Failure scenario: open the Agents window (or the Benchmarks window, or the tray popover) next to Settings and switch to 简体中文. That window's `AppChrome` re-evaluates and its literal `Text("…")` copy turns Chinese, but its `L10n.text(...)`-derived labels keep the strings they were built with and stay English until that window's views are rebuilt for an unrelated reason. Keying the scene root on `languageRaw` (`.id(languageRaw)`) or routing the override through an `ObservableObject` would close it.

ddalcu:

`.id(languageRaw)` does close the invalidation gap — with the bundle swapped before the write, the rebuilt subtree resolves every `L10n` string in the new language. It buys that with the whole subtree's state.

- `app/Sources/MLXServe/Services/AppearanceSettings.swift:192` — the identity change is applied to each scene's root, so switching the language destroys and recreates every descendant view and resets all of their `@State`, not just the copy.
  Failure scenario: type a message into the composer without sending it, open Settings ▸ Interface and pick 简体中文 — `@State private var inputText` (`ChatView.swift:1808`) is new storage under the new identity, so the draft is gone. The same applies to `composerWalk`, `composerHeight`, `inputFocused` and every scroll position in every open window. Scoping the `.id` to the copy-bearing subtree, or bumping an `ObservableObject` the `L10n` consumers observe, would refresh the strings without discarding user input.

LXD-8:
Reworked along the observation route rather than by scoping the `.id`. The root `.id(languageRaw)` is gone. `LanguageLookupRevision` is an `@Observable` holding a generation token that `L10n.text` reads while the body is built, so SwiftUI's observation tracking registers the read and a bump re-runs exactly the bodies that built `L10n` copy — in every open window — with no identity change and no per-call-site edits. The bump happens only when `BundleLanguageOverride.apply` resolves to a different bundle, so the per-scene `onAppear` reapplies are free. Validated empirically: mutating an `ObservableObject` did not re-run descendant bodies, while reading the `@Observable` property did; a composer draft held in `@State` survives a switch. `swift test` passes: 3560 tests, 26 skipped, 0 failures.

ddalcu:

`LanguageLookupRevision` is the right shape — the read inside `L10n.text` registers the dependency during body evaluation, so the swap re-runs exactly the bodies that built copy, in every window, and the `@State` finding is closed. The `bundlePath` guard against a no-op reapply is a good catch too. One surface is still outside the refresh.

- `app/Sources/MLXServe/MLXServeApp.swift:291` — the menu bar is built from `Commands` literals that never call `L10n.text`, so no read registers against the revision, and the only `apply` call sites are `MLXServeApp.swift:40` (launch) and `AppearanceSettings.swift:198` (`onAppear`); nothing re-evaluates the command menus after a live switch.
  Failure scenario: launch in English, pick 简体中文 — every window's copy turns Chinese while the Agent and Tools menus (and their items) stay English until the app is relaunched, which is when `init()` applies the bundle again. The file's own doc comment lists "AppKit's own `NSLocalizedString` (menus, window titles)" as covered, and it is — but only at launch.

LXD-8:
The command menus now re-resolve on a live switch. `MLXCoreApp` reads `InterfacePrefKey.language` alongside the command menu's existing `@AppStorage` keys, so the preference write in `AppLanguage.select` re-evaluates the Scene body and rebuilds Agent/Tools (and every other command) against the `.lproj` it swapped beforehand. No identity change, so no window `@State` is affected.

The menu redraw itself is GUI-only and cannot be exercised by `swift test`; the suite is green (3560 tests, 26 skipped, 0 failures), and the mechanism is the same App-body invalidation the menu's existing `.disabled(appState.chatDeletionTarget == nil)` already relies on.

ddalcu:
🤖 **MLXServe-PRBot approved** — reviewed at head `c7050291`, no issues found.

ddalcu:
🤖 **MLXServe-PRBot approved** — reviewed at head `a29c2f49`, no issues found.

LXD-8:
This PR is behind main (`main @ e2b4ccaa`). A rebase would help keep the diff clean. The language setting feature and tests are ready.

ddalcu:

The `Resolution` split is the right shape and closes a real hole: the development language ships no `.lproj`, so reading it as "no override" would have left the copy in the language the user just switched away from. `resolve` only claims `.sourceLanguage` when the code actually matches `developmentLocalization`, the fallback closure re-enters the exchanged selector (so it is the untouched Foundation answer, not recursion), and the `resolution != resolved` guard keeps a per-scene reapply from bumping the revision.

- `app/Sources/MLXServe/Services/LanguageSettings.swift:167` — the `.sourceLanguage` arm returns `value ?? key`, but Foundation's contract for `localizedString(forKey:value:table:)` is that an **empty** value means the key as well, and `NSLocalizedString`'s `value:` parameter defaults to `""`.
  Failure scenario: today every caller is `L10n.text`, which passes `value: key`, so nothing goes blank — but this file's own doc comment (`:17`) advertises `NSLocalizedString` (menus, window titles) as a supported path through the override. The first `NSLocalizedString("Quit", comment: "")` written against `Bundle.main` returns `""` with English selected and the correct string in every other language, which reads as a SwiftUI layout bug rather than a lookup one. `value?.isEmpty == false ? value! : key` matches what the unswizzled implementation does.

LXD-8:
Fixed in `0b62817`. You are right, and the contract matters precisely because this file advertises that path: `NSLocalizedString`'s `value:` defaults to `""`, and Foundation reads an empty value as "use the key", so under `.sourceLanguage` — which has no `.lproj` to consult — the override would have returned the empty string and blanked the menu item. My `value ?? key` only handled the `nil` written by `L10n.text`, which is why nothing shows it today.

The arm is now `value?.isEmpty == false ? value! : key`, matching the unswizzled implementation.

Test: `LanguageSettingsTests.testAnEmptyValueUnderTheSourceLanguageFallsBackToTheKey` — `""` returns the key, a non-empty value is returned verbatim. Red before the fix with `XCTAssertEqual failed: ("") is not equal to ("Quit")`; `LanguageSettingsTests`: 12 passed.

ddalcu:
🤖 **MLXServe-PRBot approved** — reviewed at head `0b628179`, no issues found.

`value?.isEmpty == false ? value! : key` now matches Foundation's own rule, so the default `value: ""` that `NSLocalizedString` passes resolves to the key instead of an empty string. The test covers both halves at once — an empty value falling back to the key, and a real default value still winning — with `resolution` asserted as `.sourceLanguage` first so it cannot pass through the wrong arm.

LXD-8:
Please re-review this PR at its current head: `acf079c5ce5e430d917fdc77dcdb5de1b33e07b8`. It is newer than the head your last review covered — the remaining fixes went in after that review, and where needed the branch was rebased onto a newer `main`. `build-test` is green on this head.

LXD-8:
A nudge on the batch, since you review these one at a time and none has moved since the PRBot's last pass: **#462, #463, #464, #480, #482, #483, #485, #487, #488** are all ready to land. The bot approved each at an earlier head; I asked for a re-review at the current heads on 2026-09-23 and nothing has changed since.

They are independent — the only overlaps are #463×#464, #464×#480 and #464×#488 (same files, different regions) and #485×#487 (the four console files). I have the merge resolution for that last pair worked out and will apply it to whichever of the two lands second, so either order is fine.

Two of them also share files with other people's open PRs (#445, #349, #411, #322), so landing them sooner keeps everyone's rebases small.

PR #463, opened 2026-09-18
fix(app): route the AppKit dialogs and the media-size hints through the catalog

LXD-8:
**Base:** `main` @ `18cac51`

**What changed / why**

An `NSAlert` reads `Bundle.main`, never SwiftUI's environment, so the app's dialogs stayed English whatever the UI language. They now look their text up: the four chat import errors, the system-prompt update flow (no-backup arm included), the small-context warning, the download-failure alert and the server crash alert. The producer side had the same gap — the media size notes, composer tips, `TaskNotifier` bodies and music download line baked their values in before any lookup, so `L10n.text` could never match a key. They build through `L10n.format` now, and the size notes plus `AgentBudget.contextWarning` use `L10n.formatUngrouped`, so quoted counts stay `1536`, not `1,536`.

- M5 Max, macOS 27.0: `swift test --no-parallel --filter 'CustomResolutionTests|LocalizationTests|ComposerTip|MusicGen'` — 86 tests, 0 failures. No screenshots: the touched surfaces render unchanged in English, so this is copy routing rather than a visual change (CONTRIBUTING asks for shots on UI changes).

- [x] Searched open and closed issues, open PRs and recent commits; no existing thread for this


ddalcu:

- `app/Sources/MLXServe/Services/AgentPrompt.swift:252` — the `??` arm of the "System prompt updated" dialog is the one string in this sweep that still bypasses the catalog, so the no-backup case renders an English body under a translated title.
  Failure scenario: with 简体中文 selected, run the system-prompt update flow on a tree where `updateSystemPromptToDefault()` returns nil (no prior `~/.mlx-serve/system-prompt.md` to back up). `messageText` reads 系统提示词已更新 while `informativeText` reads "Updated to the latest default." — every other literal in the file, including the `.map` arm one line above, goes through `L10n`.

LXD-8:
Fixed in a214c3d: the `??` arm of the "System prompt updated" dialog goes through `L10n.text` now.

The review also pointed at the mirror image of this class from the translation side (#464): four producers baked values into the sentence before any lookup, so their catalog entries could never match. All four are routed through the catalog in this commit — the composer tips (title, body, workspace detail), the task notification titles, the music download line — and the tip card renders the result as-is. The media size notes use a new `L10n.formatUngrouped` because the locale-aware path grouped them to `1,536 px`, which `CustomResolutionTests` refuses.

ddalcu:

The `??` arm is fixed and the `ComposerTip` / `TaskNotifier` / `MusicGenService` producers now look their copy up. Two things this head leaves behind.

- `app/Sources/MLXServe/Services/TaskNotifier.swift:48` — the notification *titles* are localized but the bodies next to them are not: `"Task completed."` (line 48), `"The task run failed."` (line 54), `"a tool"` (line 58) and the `"Wants to run “%@”. %@"` sentence at line 60 are still interpolated literals.
  Failure scenario: a scheduled run with no `run.summary` fails while 简体中文 is selected — Notification Center shows `⚠ 夜间评估 失败` over the English body `The task run failed.`

- `app/Sources/MLXServe/Services/AgentBudget.swift:76` — `contextWarning` went from interpolation to `L10n.format`, which passes `locale: Locale.current`, so the two token counts now render grouped; this is the exact case `formatUngrouped` was added for one file over, and it was not applied here.
  Failure scenario: launch a CLI against a 8192-token model — the alert reads "The model advertises a 8,192-token context; claude needs 32,768+ to work well", where the same numbers print ungrouped on `main` and in the media panes' hints.

ddalcu:

Both earlier findings are closed: `AgentBudget.contextWarning` is on `formatUngrouped` now, and the three notification bodies look their copy up. One thing the new shape introduces.

- `app/Sources/MLXServe/Services/TaskNotifier.swift:63` — splitting the sentence at `L10n.format("Wants to run “%@”", tool) + ". \(reason)"` leaves the joining punctuation outside the catalog, so a translated clause is terminated by an ASCII period the translator cannot change.
  Failure scenario: an approval notification with 简体中文 selected renders `想要运行“ls”. …` — the key's own quotes are full-width (`#464` line 1540: `"Wants to run “%@”" = "想要运行“%@”"`) while the sentence-ending period stays `. `, against the full-width convention the rest of the catalog follows. Keeping the period inside the key (`"Wants to run “%@”. %@"`, with the reason as the second argument) puts it back under the translator's control.

LXD-8:
Both the punctuation and the reason slot now live inside the key: `TaskNotifier` uses `L10n.format("Wants to run “%@”. %@", tool, reason)`, with the zh-Hans value `想要运行“%@”。%@`, so the full-width 。 is in the translator's string. A reason of `needs network` renders `想要运行“ls”。needs network` under 简体中文. The older `"Wants to run “%@”"` entry was kept rather than removed because `TasksView.swift:325` still resolves it through its interpolated `LocalizedStringKey` (`Label("Wants to run “\(pending.toolName)”")`), so it is not orphaned. `swift test` passes: 3551 tests, 26 skipped, 0 failures, including `LocalizationTests`.

ddalcu:
🤖 **MLXServe-PRBot approved** — reviewed at head `53f4316d`, no issues found.

ddalcu:
🤖 **MLXServe-PRBot approved** — reviewed at head `4b94b6ad`, no issues found.

LXD-8:
This PR is behind main (`main @ e2b4ccaa`). A rebase when convenient would keep the diff clean.

ddalcu:
🤖 **MLXServe-PRBot approved** — reviewed at head `06ee290d`, no issues found.

Delta since the last head is the catalog moving into this PR plus three alignments, all consistent: the context-warning key now spells `Settings ▸ Server` on both sides (`AgentBudget.swift:76` and the entry at `:1376`), the two memory-failure strings moved their trailing `\n\n` inside the key so the catalog entry matches the literal exactly rather than being concatenated after the lookup, and the mic string follows main's `MLX-Serve` rename. Checked the 39 added entries: no duplicate keys, placeholder arity and order consistent, and every key has a matching Swift literal in this branch — including the new `Approve`/`Deny` notification actions, which resolve against entries already on main.

LXD-8:
Please re-review this PR at its current head: `098d39fdf696239a10639432ae36c8de55412049`. It is newer than the head your last review covered — the remaining fixes went in after that review, and where needed the branch was rebased onto a newer `main`. `build-test` is green on this head.

PR #469, opened 2026-09-19
docs: add missing model types to supported models table

yachen4ever:
The supported models table in `docs/models.md` (and `zh-CN/models.md`) was missing several `model_type` values that are already in the discovery allow-list (`model_discovery.zig`) and have config parsing branches in `model.zig`.

**New rows added:**
- `gpt_oss` — OpenAI gpt-oss 20B-A3.6B / 120B-A5.1B MoE (Harmony format, thinking always on)
- `spark2_5` — XHToken Spark-X2.5 1.7B / 4B (Spark turns, thinking default on)
- `k2_horizon` — IFM K2-Horizon dense 0.9B–32B (IFM tags, thinking default on)

**Qwen row updated** — added `qwen3_vl`, `qwen4_exp`, `prism_hadamard_qwen35` to the `model_type` column:
- `qwen3_vl` / `qwen3_vl_text` — Qwen3-VL family (ChatML, Qwen3-VL ViT tower with M-RoPE)
- `qwen4_exp` / `qwen4_exp_text` — Qwen3.8-Flash-Next (ChatML, GDN + QSA + hyper-connection + MTP)
- `prism_hadamard_qwen35` — Prism Bonsai 2 / Ternary-Bonsai-2-27B (2-bit Hadamard-packed Qwen3.5)

Example model links are included where a recommended MLX pack exists on HuggingFace.

Both English and Simplified Chinese versions updated.


ddalcu:

The `gpt_oss` row checks out — it is in the allow-list (`model_discovery.zig:53`) with a parse branch at `model.zig:2347` and the harmony arm in `chat.zig:1744`. The other three additions do not.

- `docs/models.md:18` — `spark2_5` (and `k2_horizon` on line 19) appear nowhere in the tree: neither string is in `supported_model_types` (`model_discovery.zig:33-55`), neither has a branch in `model.zig:parseConfigFromJson`, and the chat formats the rows promise ("Spark turns", "IFM tags") have no implementation in `chat.zig` either.
  Failure scenario: a user follows the table, downloads `Spark-X2.5-4B-MLX-4bit`, and the directory never appears in the Model Browser — `isSupportedModelType` (`model_discovery.zig:95`) prefix-matches only `lfm2`, so a `config.json` declaring `"model_type": "spark2_5"` is skipped silently, which is exactly the outcome the allow-list comment describes.

- `docs/models.md:10` — `qwen3_vl` and `prism_hadamard_qwen35` are not `model_type` values this server recognizes; both families are served through the **`qwen3_5`** arm. Qwen3-VL is detected by the `vision_config` block that sets `config.qwen_vision` (`model.zig:503-504`, `:1454`), and the prism-ml Bonsai checkpoints declare `qwen3_5` themselves — see the fixture at `model.zig:5893`, which is a prism-ml 2-bit config with `"model_type": "qwen3_5"`.
  Failure scenario: the `model_type` column is the thing a user checks their `config.json` against, so this tells them a value that would make discovery skip the directory. The example link and the Vision column are right; it is the type list that is wrong.

- `docs/models.md:29` — both files lose their trailing newline (`\ No newline at end of file` on the last hunk of each), an unintended change to a line the PR does not otherwise touch.
  Failure scenario: every later diff that touches the last paragraph carries the same marker, and `cat`/`tail` output runs into the next prompt.

ddalcu:
🤖 **MLXServe-PRBot approved** — reviewed at head `0a5c0459`, no issues found.

Correction to my previous comment: the `spark2_5`, `k2_horizon` and `prism_hadamard_qwen35` findings were wrong. I grepped a stale working tree instead of `origin/main`. All three are in the allow-list (`model_discovery.zig:56-58`) with parse branches in `model.zig` (`:2411`, `:3493`, `:2434`), and the chat formats the rows name are real — the IFM think markers are at `chat.zig:2314` and `k2_horizon`'s thinking-on default at `model.zig:1167`. Those rows were accurate as first written.

The `qwen3_vl` note held: it is not a `model_type` the loader knows, and moving it to the Vision column as "via `vision_config`" matches `model.zig:1454`. The trailing newline is restored in both files.

PR #471, opened 2026-09-19
feat: MiniCPM-V 4.6 support (SigLIP2 tower + qwen3_5 trunk, images and video)

yachen4ever:
## What

Native MLX serving for **openbmb MiniCPM-V 4.6** (SigLIP2-400M tower + Qwen3.5-0.8B trunk; reference pack `mlx-community/MiniCPM-V-4.6-4bit`, the conversion mlx-vlm 0.5.x makes of it) — image AND video input on the chat surfaces.

The trunk IS qwen3_5 (hybrid GDN/full attention via `layer_types`, qk-norm, attn output gate), so it collapses onto `qwen3_5_moe` and every existing trunk path — batched GDN decode, rope, per-layer weight probing — fires unchanged; the vision wrapper keys on `config.minicpm_vision`, not the model_type.

## The vision side (per mlx-vlm's `minicpmv4_6`)

- A stock pre-LN SigLIP with one family twist: the **vit_merger** folds each 2x2 patch window into one token at tower layer `insert_layer_id` (4-token window attention around a window-mean residual, then Linear → GELU(erf) → Linear); a second 2x2 **merger** projects into language-model space — 16x patch→token compression (`downsample_mode: "4x"` skips the first fold).
- Position ids are **bucketed** (`floor(i · side / n)`, integer math) — the one NaFlex tower that never resamples its position table.
- The processor slices an image into a source view + up to 9 slice views (448px scale resolution, grids picked by aspect); the prompt block is `<image_id>k</image_id><image>…</image>` then `<slice>`-wrapped pad runs, marker order matching the encoder's piece order because the splice scatters rows positionally.
- **Video**: a sampled frame IS processed as an image (128-frame linspace cap, numpy half-even rounding); fills use `video_token_id` with a separate `<image_id>` counter, and `VideoData` carries a per-view table since grids vary per frame.
- EOS: the reference strips `<|endoftext|>` (248044) from the eos list once `<|im_end|>` (248046) is declared — sampling can hit the raw EOS first and end a turn before any token. Mirrored.

## Parity

`tests/dump_minicpm_vision_fixtures.py` executes mlx-vlm's OWN minicpmv4_6 modules on the pack's weights (the math the conversion was made with — no torch oracle needed) and dumps staged tower/merger outputs for three grids; the zig parity test asserts values (cos > 0.999 on tower hidden and final features), never shapes — a transposed 2x2 window order keeps every shape while corrupting the data. Live tests are filter-run (same convention as the lfm2 tower test): the fault-injection sweeps share a process-global op counter, so a full suite with the env vars set is not deterministic.

## Verified

- `zig build test`: 2603 pass / 0 fail (2776 total, env-free)
- Parity fixtures: 3 grids, cos > 0.999 both stages
- Live e2e on M4 Max: text (clean `<|im_end|>` stop), image (1024×1024 → 64 source + 6×70 slice tokens = 484, count matches the encoder output exactly), video (3 frames → 1452 tokens), mixed image+video (484 + 968), deterministic across repeats, RSS flat

## Notes for review

- Commit 1 is recognition only (text serving); 2 is the tower + parity harness; 3 wires the image pipeline; 4 adds video; 5 is docs.
- The mlx-c handle rule discovered while porting (a wrapper moved into a replace-style owner must never be freed by its alias — it segfaulted only under server heap churn) is written up in `docs/gotchas/engine-mlx.md` with the rule in CLAUDE.md. `lfm2_vision.zig`'s `pixelUnshuffle`/`posEmbed` carry the same latent `replace` + defer-free pattern and would be worth a follow-up.


ddalcu:

The processor port lines up with the reference on the parts I could check against this tree: `slicedGrid` keeps `grid[0]`=cols / `grid[1]`=rows consistently through `refineSize`, the cell loops and `tile_index = 1 + row*cols + col`, and `minicpmImageSegment` reads those same fields back (`offset / tile_cols` for the row, `offset % tile_cols == 0` for the newline), so the prompt block mirrors the encode order. The `\n`-between-rows placement matches `get_grid_placeholder`'s `"\n".join(rows)`, and `pads = grid_h*grid_w/divisor` stays exact because every view is snapped to a multiple of `patch*4`. Two defects in the video encode path.

- `src/scheduler.zig:4739` — `emb` is freed and *then* used as the output handle for `mlx_concatenate_axis(&emb, …)`, so the concatenate writes through a destroyed wrapper.
  Failure scenario: any MiniCPM-V video with more than one view — which is every video, since a clip yields one view per frame at minimum — takes this branch. `mlx_array_free` releases the object `emb.ctx` points at, and `mlx_concatenate_axis` then assigns into it; the write lands in freed heap. It survives a quiet run and corrupts under server churn, which is the exact class the PR documents in `docs/gotchas/engine-mlx.md`. Every other concatenate in the tree keeps the fresh handle instead — `ane.zig:1628-1637` is the shape: `var joined = mlx_array_new(); … try mlx.check(mlx_concatenate_axis(&joined, vec, 0, s));`.

- `src/scheduler.zig:4736` — the single-view branch swaps a fresh empty into `parts.items[0]` instead of freeing it, so the wrapper `forwardPatches` returned is never released.
  Failure scenario: a one-view video (a single frame inside the slice budget) leaks one `mlx_array` plus its underlying array reference per request. `mlx_array_set` copies the array into `emb`'s own object — `ane.zig:1636` does the same thing and lets its source be freed normally — so the source must still be freed here, not orphaned.

yachen4ever:
Both fixed (e1f619e..): the concatenate no longer assigns through a freed `emb` wrapper (the free moved onto the error path only, where it belongs), and the single-view arm now frees the source wrapper `mlx_array_set` copied from instead of orphaning it with a fresh-empty swap. Verified on the tree: `zig build test` 2603/0, video and mixed image+video e2e on `MiniCPM-V-4.6-4bit` deterministic with flat RSS. Fitting that this landed in the same PR that documents the class — the rule's own author walked straight into it on a path whose parts list is one element longer than the one I tested first.

ddalcu:
🤖 **MLXServe-PRBot approved** — reviewed at head `f039d426`, no issues found.

Both handle-ownership findings are closed: the concatenate now writes into the live `emb`, and the single-view arm frees the source after `mlx_array_set` before swapping in the fresh empty, so the parts cleanup stays balanced. `mlx_vector_array_new_data` copies the array objects at construction, so the vector is unaffected by that free.

PR #472, opened 2026-09-19
docs: add missing model types to supported models table

yachen4ever:
Six model_types are in the discovery allow-list and have config parsing branches in model.zig, but were absent from the docs table:

- `gpt_oss`: OpenAI gpt-oss 20B-A3.6B / 120B-A5.1B MoE (Harmony format)
- `spark2_5`: XHToken Spark-X2.5 1.7B / 4B (Spark turns, thinking default on)
- `k2_horizon`: IFM K2-Horizon dense 0.9B–32B (IFM tags, thinking default on)
- `qwen3_vl` / `qwen3_vl_text` on the Qwen family row (Qwen3-VL ViT tower)
- `qwen4_exp` / `qwen4_exp_text` on the Qwen family row (Qwen3.8-Flash-Next)
- `prism_hadamard_qwen35` on the Qwen family row (Prism Bonsai 2)

English and zh-CN tables updated together.


ddalcu:
🤖 **MLXServe-PRBot approved** — reviewed at head `6c514186`, no issues found.

Tree is identical to #469's final head. Re-checked the six types against `origin/main`: `gpt_oss`, `spark2_5`, `k2_horizon`, `qwen4_exp` and `prism_hadamard_qwen35` are all in `supported_model_types` (`model_discovery.zig:33-58`) with parse branches in `model.zig`. One note on the description rather than the diff: it lists `qwen3_vl` / `qwen3_vl_text` as added to the Qwen row, but the table does not add them — correctly, since neither is a `model_type` the loader knows (Qwen3-VL is detected by `vision_config`, `model.zig:1454`), and the row says so in the Vision column.

PR #473, opened 2026-09-19
Linux port: build + serve against the mlx-omarchy Vulkan MLX fork

joshuaswarren:
# Linux port: serve the mlx-omarchy Vulkan fork of MLX

Adds a Linux build graph so `mlx-serve` runs on Linux (tested on Arch ARM8 / Omarchy, Apple Silicon M1 Max under Asahi) against the [mlx-omarchy](https://github.com/joshuaswarren/omarchy-mlx) Vulkan fork of MLX instead of Apple Metal.

## What the build graph does

- `build.zig`: new target-gated `addLinuxServe` mirroring the macOS serve graph minus everything Apple-specific:
  - ds4 (Metal), embedded llama.cpp (GGUF), and the ANE offload swap to compile-time stubs behind a new `build_options.macos_engines` flag (`ios` keeps its existing low-mem/sandbox meaning). Linux serves MLX safetensors only — the same surface as the iOS build.
  - The macOS-only semver 26.2 `default_target` minos is now host-gated (on Linux it would be read as a glibc floor and break the native graph).
  - Staged libs: `lib/mlx/lib/libmlx.so` + `libmlxc.so` (rpath'd via `$ORIGIN`), `libjinja-linux.a` built by the staging script (the committed `libjinja.a` holds Mach-O objects), system `libwebp` instead of Homebrew, Avahi's `libdns_sd` compat lib for mDNS peer discovery.
  - Configure-time checks fail loudly when `lib/mlx` is not staged or the `lib/opencode2-mlx-serve` submodule is missing (its files are `@embedFile`d — a clean checkout without it died mid-compile with a bare FileNotFound).
- `scripts/build-mlx-linux.sh`: stages the Linux mlx + mlx-c pair into `lib/mlx/`, the counterpart of `scripts/build-mlx.sh`:
  - mlx comes from a staged mlx-omarchy tree (`scripts/prepare-mlx.sh` output; upstream MLX + the Honeykrisp Vulkan backend, `MLX_BUILD_OMARCHY=ON`).
  - mlx-c is the pinned submodule (`56b2d39`) built `MLX_C_USE_SYSTEM_MLX=ON` against the fork's installed headers, plus two small patches: a `gather_qmm` `global_scale` shim (no C ABI surface for it yet) and x86_64 f16/bf16 data accessors (uint16_t alias).
  - Builds `libjinja-linux.a` with zig c++.
- `src`: portable total/available RAM (sysctl vs `/proc/meminfo`), RSS + CPU via procfs, Linux statfs, `/proc/self/exe` exe path, llseek file sizes, ANE stub C TU, and `mlx.zig` asks `mlx_device_is_available` instead of `mlx_metal_is_available` so GPU availability is reported for the Vulkan backend too (otherwise every custom-kernel path throws "[metal_kernel] Only supports the GPU").
- `src/model.zig`: the 1-D f16→bf16 table narrowing defaults off Darwin — the Vulkan backend capability-gates bf16 ops and throws on a shape miss; `MLX_SERVE_F16_NARROW_1D=1` restores it.
- ELF rpaths: `addMlxLib`'s `@loader_path` entries don't expand on ELF; the Linux graph mirrors them in `$ORIGIN` form (without this the binary starts only with `LD_LIBRARY_PATH` set).

## Build (Linux)

```bash
git clone --recurse-submodules <fork> && cd mlx-serve
./scripts/fetch-zig.sh && export PATH="$PWD/.zig-toolchain:$PATH"
# stage the Linux mlx tree once (mlx-omarchy checkout):
#   cd <mlx-omarchy> && scripts/prepare-mlx.sh   # prints <mlx-omarchy>/.work/mlx
MLX_SOURCE=<mlx-omarchy>/.work/mlx ./scripts/build-mlx-linux.sh
zig build -Doptimize=ReleaseFast
./zig-out/bin/mlx-serve --version
```

System deps: cmake, Vulkan headers + a Vulkan 1.2 driver, `libwebp`, `libdns_sd` (Arch: avahi; Debian: `libavahi-compat-libdnssd-dev`).

## Verification

Clean-checkout build + serve verified on M1 Max (Asahi/Omarchy Linux, aarch64): `mlx-serve 26.9.5-dev`, `mlx 0.32.3` (fork), `mlx-c 56b2d39`. End-to-end OpenAI-compatible serving of `Qwen2.5-7B-Instruct-4bit` with greedy output matching `mlx_lm.server` and oMLX token-for-token on the same prompt. Measured single-stream decode (128 greedy tokens, 8 interleaved rounds): ~5.7 tok/s on 7B 4-bit — functional but slow; the standard MLX op set over Vulkan carries none of the Metal fast paths, and oMLX's custom kernels are why it hits ~33 tok/s on the same backend family. Prompt Lookup Decoding (on by default) showed no effect on this benchmark.

Upstreaming notes: this PR is intentionally minimal on the macOS side — the only behavior changes for macOS are the host-gated minos default (no effect on macOS) and nothing else; every Linux path is target-gated.


ddalcu:
🤖 **MLXServe-PRBot approved** — reviewed at head `673e4407`, no issues found.

What I checked, since the risk in a port like this is a quiet macOS regression: every new branch is `comptime builtin.os.tag.isDarwin()`, so `noGpuBackend`, `wiredLimitBytes`, `getMetalBufferLimit`, `narrow1dEnabled` and the statfs struct are byte-for-byte the old path on Darwin. The `ios` → `macos_engines` flip inverts the sense correctly and all three options tables define it (`build.zig:125` true, `:372` false, `:549` false), so no graph can silently pick the wrong module. The stub surfaces match the real ones for what actually gets addressed — `LlamaEngine.ChatTurn` (`main.zig:2261`) is nested in both, `Ds4Session.evalSpeculativeSampled` matches `arch/ds4.zig:581` argument for argument. `fileFingerprint`'s lseek-to-end is safe because the window reads are `pread`, and `readProcFile`'s single read covers both procfs files many times over with `MemAvailable:`/`VmRSS:` near the top.

ddalcu:
Nice clean PR !

[joshuaswarren](https://github.com/joshuaswarren) - is a legend ✅ 
PR is a minimal diff, no ai slop - ✅ 

Thank you ! This was a surprise !

PR #475, opened 2026-09-19
feat: Laya typed-decision models (model_type laya) + POST /v1/decisions

sbusso:
## Summary

Adds native support for [Laya](https://huggingface.co/convaiinnovations/laya) typed-decision models (`model_type: laya`), starting with the MLX FP16 export [`aac6fef/laya-multilingual-mlx`](https://huggingface.co/aac6fef/laya-multilingual-mlx). Laya is a non-autoregressive ModernBERT/mmBERT encoder with a decision head: it answers typed questions (`choice`, ordinal `score`, boolean `noul`) about a text or JSON state with calibrated probabilities in one forward pass.

- **New model type `laya`** (`src/laya.zig`): config parsing, prompt/sequence construction identical to the reference `laya_mlx` runtime, ModernBERT encoder (rotary, GeGLU, alternating global/local attention), decision head, marker scorer, action head, calibration, batching, `mlx_compile`-wrapped forward.
- **New endpoint `POST /v1/decisions`**: `{"model", "state", "questions"}` in, Laya's `predict` JSON schema out. Chat requests on a decision model get a 400 that points to the endpoint. `/v1/models` advertises a `decisions` capability.
- **Discovery**: the checkpoint has no top-level `config.json`; `model_discovery` detects `rl_agent_config.json` + `encoder/config.json`. On-demand load via `--model-dir` works.
- **Tokenizer**: HF `Metaspace` pre-tokenizer and BPE `byte_fallback` encode + `ByteFallback` decode (`src/tokenizer.zig`). Both are gated on the parsed `tokenizer.json`, so Gemma-class models keep their current path; the byte-fallback change also fixes out-of-vocab characters for any checkpoint that declares `byte_fallback: true`.
- Docs (`docs/api.md`, `docs/models.md`), index page row, unit tests, fixtures dumped from `laya_mlx`, and an end-to-end HTTP test (`tests/test_laya_decisions.sh`).

## Parity

Against `laya_mlx` fixtures, 6 states (en, fr, hi, CJK ext-B, mixed scripts, cuneiform), tolerance 0.01:

| state | max abs probability diff |
|---|---|
| en | 0.0005 |
| fr | 0.0000 |
| hi | 0.0001 |
| cjk_extb | 0.0001 |
| mixed_scripts | 0.0002 |
| cuneiform | 0.0003 |

Encoder hidden states are bit-exact with `laya_mlx`; scorer logits match to 4 decimals. Tokenizer ids and decode round-trips match HF `tokenizers` on 11 byte-fallback cases.

## Latency

1 request, 3 questions, median of 30 after warm-up, M5 Max:

| path | median |
|---|---|
| in-process forward | 4.2 ms (`laya_mlx`: 4.1 ms) |
| HTTP round trip incl. tokenization | 5.2 ms |

## Test plan

- [x] `zig build test`: 2604 passed, 176 skipped, 0 failed
- [x] `tests/test_laya_decisions.sh`: PASS=16 FAIL=0 (discovery, parity, validation, latency)
- [x] Existing Gemma tokenizer tests unchanged

## Open question for maintainers

`/v1/decisions` is a new endpoint outside the OpenAI / Anthropic / Ollama surfaces. Happy to rename or reshape it (e.g. under `/v1/classifications`, or as an extension of `/v1/embeddings`) if you prefer a different wire. Marked as draft for that reason.

## Known limits

- Unigram-model byte fallback is not implemented (no supported checkpoint uses it).
- `pyJson` state serialization matches `json.dumps` for the value shapes in the fixtures; exponent-form floats are not reproduced.
- Weight memory doubles briefly at load.


ddalcu:
oh snap... i knew this was coming 😄 

My name Jev - https://www.youtube.com/shorts/3bf_0osukto

WIl review it ! Thanks

ddalcu:
merged, thank you !  added a swift ui also, so you can play with it after downloading.

sbusso:
Thanks for merging this, and for the review.

A review of the Laya engine after the merge found a few problems. The fixes are in #511, one commit per change. Please look at it first for the memory fix: on main, each new input shape keeps its own copy of the embedding table, so memory grows fast under varied traffic.

PR #476, opened 2026-09-20
qwen4_exp: accept raw BF16 ngram_table.bin

amitpandey-io:
Adds a raw-BF16 ngram table mode (header bits 16, single weight BF16 [rows, dim], no scales/biases). PLE lookups stay bit-exact for tables that never went through affine quant.

Why: lets a pack ship the original BF16 n-gram table (95 GB, SSD-resident, per-row reads) instead of a quantized one.

Verified (M5 Max, macOS 26.6.2): zig build test passes incl. a new raw-row unit test; served Qwen3.8-Flash-Next (4-bit trunk) with a 320M-row BF16 table over HTTP at 1k/16k/62k/156k prompts, log shows n-gram table 320001536 rows x 160 (16-bit, mmapped), outputs coherent, decode within noise of the 8-bit table. No tests/ integration script: it needs a 95 GB table, CI-infeasible.


ddalcu:

The raw arm holds up where it matters: `headerRegion` already validates shape×elem against the `data_offsets` span and the absolute end against `map.len`, so `row`'s `w_off + r*dim*2` can't walk off the mapping; every `wcols`/`scols` consumer (`preadSite`, the pool branch) sits behind the early return, so the zeroed columns are never divided by or indexed with; `warmMain` walks the file by offset and is bits-agnostic, so warm still works; and the serial `gather` arm is byte-identical to the fallback the quantized path already uses. One contract wart.

- `src/qwen4_exp.zig:305` — a raw BF16 table must still carry a `group_size` in `__metadata__` even though the mode has no groups, and neither the format comment nor the errors say so.
  Failure scenario: someone writes a converter from the new doc comment (`:10` — "`weight` BF16 [R, dim], no scales/biases, `"bits":"16"`", no mention of `group_size`) and omits the key. `parse` returns `error.NgramTableHeader` at `:305`, or `error.NgramTableBits` at `:310` if they write `"0"` for the field they have no value for — both errors name something other than the missing key, on a 95 GB artifact whose only feedback is that the model won't load. The new test passes `"group_size":"32"` (`:800`), so it does not catch this. Either default `gs` to 1 when `bits == 16` before the range check, or say in the comment that the key is still required.

ddalcu (approved):

PR #477, opened 2026-09-20
Qwen-Image-2.1 image generation (quantized packs)

ddalcu:
## Summary

Adds **Qwen-Image-2.1** (`Qwen/Qwen-Image-2.1`, Apache-2.0) as the fourth image backend, a port of the pure-MLX reference in [mflux PR #736](https://github.com/mflux-community/mflux/pull/736). Quantized packs only, sized for 32 GB and 16 GB Macs.

- `src/qwen_image.zig`: 7.1B block-causal single-stream DiT (two sdpa calls per block: causal text + maskless image, never a dense mask; ONE modulation shared by all blocks, read at t=0 for text and the sampled t for image tokens), 3-axis RoPE on interleaved pairs, 64-channel /16 VAE (decoder + encoder), FlowMatchEuler with the resolution-dependent exponential shift.
- Text encoder is `mage_flow.TextEncoder`, which now reads its hidden/intermediate widths from the checkpoint (Qwen3-VL-8B here, 4B for MageFlow).
- Features: txt2img, img2img (VAE encoder loads on first use), real CFG (`guidance_scale` != 1 runs two forwards per step against `negative_prompt`). 40 steps when a request names none (`ImageEngine.defaultSteps`). No edit, no LoRA, no conditioning rebalance, matching the reference.
- Wiring: `gen.zig` union arm + capability predicates, `model_discovery.zig`, residency estimator, Swift presets + UI gating, docs, changelog, test matrix.

## Sample renders

Prompt: *A red fox sitting in fresh snow, holding a wooden sign that says "MLX", soft morning light, photograph* (seed 42, guidance 1.0)

| 8-bit pack, 1024x1024, 40 steps | 4-bit pack, 512x512, 20 steps |
|---|---|
|  |  |

## Packs

`tests/convert_qwen_image21_weights.py` keeps the checkpoint's own layout and key names, quantizes the DiT block linears and text-encoder layer linears (group 64), drops the vision tower / `lm_head` / VAE `time_conv`, and writes the root `config.json` (`model_type: qwen_image21`) discovery classifies on. It streams one tensor at a time, so it converts the 33 GB bf16 repo on a 32 GB Mac.

| Preset | DiT | Text encoder | VAE | On disk | For |
|---|---|---|---|---|---|
| `--preset 32gb` | 8-bit | 8-bit | f32 | **17.6 GB** | 32 GB Macs |
| `--preset 16gb` | 4-bit | 4-bit | f32 | **10.7 GB** | 16 GB Macs |

Kept dense: the VAE, `embed_tokens` (read with `mlx_take_axis`), norms, and the DiT's small/shared linears (img_in, txt_in, timestep embedder, the shared modulation, norm_out, proj_out: ~0.3 GB).

Both are on the Hub, which is what the app presets point at: [ddalcu/Qwen-Image-2.1-MLX-Serve-8bit](https://huggingface.co/ddalcu/Qwen-Image-2.1-MLX-Serve-8bit) (17.65 GB) and [ddalcu/Qwen-Image-2.1-MLX-Serve-4bit](https://huggingface.co/ddalcu/Qwen-Image-2.1-MLX-Serve-4bit) (10.69 GB). Their model cards say they load only on this branch until it ships; that notice should come off the cards at release.

## Performance and memory (M1 Pro, 32 GB)

| Pack | Size | Steps | Wall clock (incl. load) | Peak process footprint |
|---|---|---|---|---|
| 8-bit | 1024x1024 | 40 (default) | 985 s, ~23 s/step | **12.95 GB** |
| 8-bit | 512x512 | 8 | 69 s | - |
| 4-bit | 512x512 | 20 | 118 s | - |
| 4-bit | 1024x1024 | 3 | 87 s | **9.55 GB** |

For reference, the mflux PR reports ~1.5 s/step at 1024x1024 bf16 on an M5 Max, ~46 GB peak bf16 / ~30.7 GB with `-q 8`.

Two things keep the footprint down:

- **Staged text encoder.** The encoder is half the pack and idle for the whole denoise. Where weights + a 4 GB generation allowance pass 3/4 of the GPU working set (`gen.qwenImageStagesTextEncoder`), the engine loads it per request and frees it before the first DiT forward; `gen.qwenImagePeakBytes` bills the same plan, so the load gate and the engine cannot disagree. With it staged the 40-step denoise is flat at DiT + buffer pool (5.5 GB on the 4-bit pack at 1024x1024). The measured 4 GB over resident matches the allowance.
- **Exact banded VAE.** The first 1024x1024 decode peaked at **18.1 GB** on a 5.5 GB engine while `mlx_get_peak_memory` read 4.9 GB: MLX's 3x3 conv holds an unfolded input copy (11 GB at 288 channels) and a stage keeps ~6 full-resolution f32 intermediates alive. Large convs now run in row strips and whole stages in row bands with a cropped halo. Both are exact, since everything in a stage except its 3x3 convs is per-pixel (the channel norm is `rms_norm`). 18.1 -> 9.55 GB, identical pixels.

## Parity

`tests/dump_qwen_image21_fixtures.py` builds a TINY random-weight DiT + VAE with the reference's own classes, writes them as a pack in the real key names, and dumps one fp32 forward each. No checkpoint needed.

| Oracle | cos | rms ratio |
|---|---|---|
| RoPE table | 1.000000 | 1.000000 |
| DiT forward | 1.000000 | 0.999998 |
| VAE decode / encode | 1.000000 | 1.000000 |
| VAE decode / encode, bands forced | 1.000000 | 1.000000 |
| Strip conv vs whole conv | 1.000000 | 1.000000 |

Plus a bf16 DiT pass asserting the stream stays bf16, the sigma schedule pinned against the reference scheduler's own output, and an env-gated in-process e2e on a real pack (`QWEN_IMAGE_E2E_MODEL`).

## Testing

- `zig build test`: 0 failures with the fixtures enabled. Note `prefix_cache ... the lien weighs the share a restore DELIVERS` failed intermittently in full runs, including on a clean `main` with these changes stashed; it passes alone and passed on the final run. Unrelated, looks flaky.
- `bash app/test.sh`: 3551 tests, 0 failures, including the new preset-capability test.
- Converter `--self-test` passes; both real packs converted and generated correct images (fox in snow holding a legible "MLX" sign).
- **Not run:** `tests/test_qwen_image_gen.sh` (boots its own server). `QWEN_IMAGE_MODEL=<pack> ./tests/test_qwen_image_gen.sh`

## Not ported

Edit variant, LoRA key mapping, text-prefix KV caching across steps (the reference skips these too), and the website copy.


ddalcu:

Checked the seams this shares with the other backends rather than the whole port. `Engine.load` has a complete errdefer ladder (`qwen_image.zig:1354`→`:1383`), so a failure at any stage unwinds the earlier ones — the partial-construction class that bites elsewhere. The staging decision is genuinely one answer: `ImageEngine.load` and `estimatePeakResidentBytesIn` both feed `qwenImageStagesTextEncoder` the same `sumSafetensorsIn` total (which does descend into `text_encoder/`, so `-| te` is the real remainder), and the engine frees the staged encoder inside `encodeConditioning`'s defer — before the denoise — which is exactly what `stagedPeakBytes(rest, {te, transient})` bills. `clampKreaDim` rounds to a multiple of 16 and clamps to 256..2048, both multiples of `VAE_DOWNSAMPLE`, so `height / 16` is always exact. The CFG blend's `defer free(v)` reads `v` at scope exit, so the reassignment to `blended` is freed once and the original explicitly — no leak. I also swept the module for the free-then-write-back handle pattern: zero hits. One defect.

- `src/qwen_image.zig:329` — `computeSigmas` divides by zero for `steps == 1`: `shifted()`'s own `steps_f <= 1` arm returns 1.0 for every `i`, so `scale` is `(1.0 - 1.0) / 0.98 == 0`, and the loop's `(1.0 - shifted(...)) / scale` is `0.0 / 0.0` — NaN.
  Failure scenario: `POST /v1/images/generations` with `"steps": 1` against a Qwen-Image pack. `handleImage` applies no floor to `steps` (`gen.zig:1862` reads it straight from the body, unlike the music route's `[4,100]` check at `:2587`), so `sigmas[0]` is NaN, the single Euler step scales the velocity by `0 - NaN`, and the VAE decodes noise into a garbage image with no error anywhere. `steps == 0` is safe (`generateImage` maps it to `DEFAULT_STEPS`); only 1 degenerates, and the schedule test only covers 8 steps (`:1572`). Returning `{1.0, 0.0}` for a single step matches what the reference scheduler produces.

ethicalvats:
using it works like a charm

any plans on implementing edit and refs to img ?

where is the bottleneck?

PR #478, opened 2026-09-21
feat: support original Xing4.0 BF16 checkpoints

beamivalice:
## Summary
Load the original Xing4.0-29B-A4B BF16 checkpoint directly, without conversion or a Python serving runtime.
- Add full MLA + four-stream mHC, native SentencePiece BPE, and shared `<param_key>` tool parsing; expose the architecture in the app.
- Preserve the reference's BF16/f32 boundaries and full-score YaRN scaling; stage expert banks in memory and bill prefill working memory.
- Scope: serial BF16 inference, no MTP or quantized weight packs; full 256K context is not a validation claim.

## Verification — M5 Max, 128 GB, macOS 26.6.2
- ReleaseFast build passed; `zig build test -Doptimize=ReleaseFast`: 2,631 passed, 172 skipped, zero failures.
- `bash app/test.sh`: 3,558 tests, 26 skipped, zero failures.
- Tiny original-layout CPU-reference fixture: prefill, cached decode, and position 4096 pass (<0.006 max logit error; 1% relative-RMS bound).
- `tests/test_xing4.py` passes on the original BF16 checkpoint: real load, English/Chinese, streaming parity, thinking, tool call/history, Messages, Responses, concurrency, and unload. Reasoning framing whitespace is normalized for parity; answer bytes are exact. Server uses the memory guard, with no override.


beamivalice:
Screenshot 2569-09-21 at 10 21 49

beamivalice:
## Mixed affine weights + compressed MLA update

Pushed the extension on top of this PR. The local checkpoint is **17,961,945,544 weight bytes (~17.96 GB)**: routed experts **affine 4-bit / group 64**, other 2-D trunk matrices **affine 8-bit / group 64**. Norms/scalars remain unquantized; the unused MTP layer is omitted. Original BF16 weights are unchanged.

### Measured performance

**Apple M5 Max, 128 GB, macOS 26.6.2, AC power.** llmprobe **0.6.11**, warmup + **3 measured runs**, medians, greedy/thinking-off, 192-token decode workload. Both servers: ReleaseFast, context 32K, prefill chunk 512, one slot, BF16 KV, prefix cache off, speculation off. Runs were sequential in the same session; both sustained-load checks reported 0% drift. These are not directly comparable to the app screenshot's different settings.

Baseline: mlx-serve **71a2b44**, original BF16 weights, expanded MLA cache. New: this extension, mixed-affine weights, latent cache and inference optimizations. Gains combine the weight and implementation changes; they are not an isolated latent-cache speed claim.

| Scenario | Previous BF16 | Mixed + latent | Improvement |
|---|---:|---:|---:|
| Standard decode | 25.6 tok/s | **67.2 tok/s** | **2.63×** |
| Standard prefill, 2,037 tokens | 577.2 tok/s | **1,916.3 tok/s** | **3.32×** |
| Standard TTFT | 418 ms | **201 ms** | 52% lower |

| Actual prompt tokens | BF16 decode | New decode | Decode gain | BF16 prefill | New prefill | BF16 → new TTFT |
|---:|---:|---:|---:|---:|---:|---:|
| 503 | 24.7 tok/s | **63.4 tok/s** | **2.57×** | 570 tok/s | **1,633 tok/s** | 882 → **308 ms** |
| 8,228 | 19.3 tok/s | **47.5 tok/s** | **2.46×** | 529 tok/s | **1,097 tok/s** | 15,553 → **7,498 ms** |
| 16,377 | 15.7 tok/s | **39.3 tok/s** | **2.50×** | 455 tok/s | **786 tok/s** | 35,990 → **20,829 ms** |

### KV memory per token

Both cache representations below use **BF16**, not KV quantization.

- Expanded: `40 × 32 × (192 + 128) × 2 = 819,200 bytes/token` (**800 KiB**).
- Latent: `40 × (512 + 64) × 2 = 46,080 bytes/token` (**45 KiB**).
- **94.375% less KV storage — a 17.78× reduction.**

| Tokens | Expanded BF16 KV | Latent BF16 KV |
|---:|---:|---:|
| 8,192 | 6.25 GiB | **0.352 GiB** |
| 16,384 | 12.50 GiB | **0.703 GiB** |
| 32,768 | 25.00 GiB | **1.406 GiB** |
| 131,072 | 100.00 GiB | **5.625 GiB** |

The storage table is calculated from validated cache shapes; it excludes weights, capacity slack, duplicate prefix entries and working buffers. The 128K row is **not** a claim of a completed 128K inference test.

A separate real-server test completed **32,763-token cold, warm, divergent and warm-divergent requests**, with two prefix-cache entries and a 4 GB cache budget. Warm requests reused **32,762 tokens**. MLX active memory after replies was **19.02–20.53 GiB**; allocator peak reached **26.24 GiB**. This is correctness/capacity evidence, not an additional throughput benchmark.

### Scope and validation

- Original BF16 and mixed-affine tiny CPU-reference fixtures pass through prefill, cached decode and position 4096 at the existing error bounds. Quantized arithmetic uses an all-expert fixture; independent F32 router tests cover top-k selection.
- Mixed-pack real HTTP tests pass for English/Chinese, thinking, streaming, tools/history, Messages, Responses, concurrency and unload.
- Full Zig suite: **2,648 passed, 172 skipped**. Converter self-test: **40 checks**.
- Muse Spark 1.3 Contributor and DeepSeek V4.1 Flash performed parallel read-only audits through OpenCode Go. Their suggestions were checked against the pinned MLX source and tests rather than accepted as performance claims.
- Attention remains dense and inference remains serial; no MTP or full 128K/256K runtime claim. Further runs were stopped at the user's request.

ddalcu:
🤖 **MLXServe-PRBot approved** — reviewed at head `8d885826`, no issues found.

7.5k lines is more than one pass can cover honestly, so I reviewed the three places this could hurt models that already work, and all three hold:

- **The shared MLA path.** `mlaAttnWith` is Ling 3.0's attention too, and the diff looked like it deleted the kv_b expansion. It does not — the expansion moved verbatim into the `else` arm of `if (cfg.mla_latent_kv)`, op for op against `main`, and `mla_latent_kv` is set only inside the `xing4_0` branch (`model.zig:2954`), so no other arch can reach the latent path. The geometry helpers at `model.zig:820/830/842/862` are all ternaries on the same flag, false for everyone else.
- **The disk cache.** `modelFingerprint` now delegates with a null namespace, which keeps every existing SSD root byte-stable, and only the latent layout adds a marker (`cacheLayoutNamespace`, wired at the single attach site in `scheduler.zig:4270`). The test pins both halves — old == null-namespace, marker != old, marker stable.
- **The new protobuf parser.** `readVarint` bounds-checks every byte and cannot overflow its `u6` shift (the `i == 9` guard forces a terminating byte before the next `shift += 7`), and `readBytes` computes its bound as `data.len - pos` with `pos <= len` maintained, so a hostile length cannot wrap.

Not reviewed in depth: `xing_hc.zig`, `mla.zig` internals, and the 1.8k-line converter — the fixture tests are what stand behind those.

